PR #95, opened 2021-09-07
Adding notes for contributors

ant0nsc:


docs/source/contributing.md
```diff
@@ -1,17 +1,57 @@
...
+    """
+```
+
+## Unit testing
+- DO write unit tests for each new function or class that you add.
+- DO extend unit tests for existing functions or classes if you change their core behaviour.
+- DO try your best to write unit tests that are fast. Very often, this can be done by reducing data size to a minimum.
+Also, it is helpful to avoid long-running integration tests, but try to test at the level of the smallest involved
+function.
+- DO ensure that your tests are designed in a way that they can pass on the local box, even if they are relying on
```
JonathanTripp, line 52: Maybe local machine instead of local box?

docs/source/contributing.md
```diff
@@ -1,17 +1,57 @@
...
+specific cloud features. If required, use `unittest.mock` to enable the tests to run without cloud. 
+- DO run all unit tests on your dev box before submitting your changes. The test suite is designed to pass completely
```
JonathanTripp, line 54: And here too, although here it is dev box

docs/source/contributing.md
```diff
@@ -1,17 +1,57 @@
...
+```
+
+## Unit testing
+- DO write unit tests for each new function or class that you add.
+- DO extend unit tests for existing functions or classes if you change their core behaviour.
+- DO try your best to write unit tests that are fast. Very often, this can be done by reducing data size to a minimum.
+Also, it is helpful to avoid long-running integration tests, but try to test at the level of the smallest involved
+function.
+- DO ensure that your tests are designed in a way that they can pass on the local box, even if they are relying on
+specific cloud features. If required, use `unittest.mock` to enable the tests to run without cloud.
```
JonathanTripp, line 53: Maybe locally instead of "without cloud"?

JonathanTripp (approved):

PR #96, opened 2021-09-07
Logging SDK version

ant0nsc:


src/health/azure/himl.py
```diff
@@ -31,16 +32,15 @@
...
-logger.setLevel(logging.DEBUG)
-
 RUN_RECOVERY_FILE = "most_recent_run.txt"
 WORKSPACE_CONFIG_JSON = "config.json"
 AZUREML_COMMANDLINE_FLAG = "--azureml"
 RUN_CONTEXT = Run.get_context()
 OUTPUT_FOLDER = "outputs"
 LOGS_FOLDER = "logs"
 AML_IGNORE_FILE = ".amlignore"
+SDK_NAME = "innereye"
```
JonathanTripp, line 42: What is the SDK name and version? Isn't it hi-ml? Shouldn't the version be the same as in PyPI?

JonathanTripp (approved):

PR #112, opened 2021-09-17
Add Tensorboard ability to read from different folders

mebristo:
Update himl_tensorboard to work with tensorboard files that are not necessarily stored in the 'logs' folder


src/health/azure/himl_tensorboard.py
```diff
@@ -3,17 +3,112 @@
...
+
+
+class WrappedTensorboard(Tensorboard):
+    def __init__(self, remote_root: str, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.remote_root = remote_root
+
+    def download_tb_files_from_run(self, run: Run, local_tb_dir: str) -> None:
+        """
+        Download tensorboard files from run.
```
ant0nsc, line 37: here and below: Sphinx wants a blank line between sumary and the first :param

src/health/azure/himl_tensorboard.py
```diff
@@ -3,17 +3,112 @@
...
 OUTPUT_DIR = ROOT_DIR / "outputs"
+TENSORBOARD_DIR = ROOT_DIR / "tensorboard_logs"
+
+
+class WrappedTensorboard(Tensorboard):
+    def __init__(self, remote_root: str, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.remote_root = remote_root
+
+    def download_tb_files_from_run(self, run: Run, local_tb_dir: str) -> None:
```
ant0nsc, line 35: is that a method defined in Tensorboard? it is not used here.

src/health/azure/himl_tensorboard.py
```diff
@@ -3,17 +3,112 @@
...
+        Download tensorboard files from run.
+        :param run: A single AML run
+        :param local_tb_dir: the local directory to download TensorBoard files into
+        :return:
+        """
+
+        try:
+            run.download_files(prefix=self.remote_root, output_directory=local_tb_dir)
+        except Exception as e:
+            raise e
```
ant0nsc, line 46: that try/catch is a no-op?

ant0nsc (changes requested):

PR #113, opened 2021-09-21
Add function to download Run files without timeout

mebristo:
Run.download_files has timeout issues, but Run.download_file (singular) does not. So here we add a function to lookup all files associated with a run (optionally filtered by some prefix) and to download them individually

This PR also extends to updating the himl-download script to use this new function, and improves on the command line args used by this script.


hi-ml-azure/conftest.py
```diff
@@ -50,3 +50,4 @@ def random_folder() -> Generator:
     remove_and_create_folder(folder)
     print(f"Created temporary folder for test: {folder}")
     yield folder
+    shutil.rmtree(folder)
```
ant0nsc, line 53: I would love to leave that in place. I find it very helpful to have all results still in the results folder - they will anyway be cleaned up completely at the start of the next test run.

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -636,3 +705,114 @@ def test_get_aml_runs(tmp_path: Path) -> None:
...
+    config = ScriptRunConfig(
+        source_directory=".",
+        command=["cd ."],  # command that does nothing
+        compute_target="local"
+    )
+    run = experiment.submit(config)
+
+    file_to_upload = tmp_path / "dummy_file.txt"
+    with open(file_to_upload, "w+") as f_path:
+        f_path.write("Hello world")
```
ant0nsc, line 735:
```suggestion
    file_contents = "Hello world"
    file_to_upload.write_text(file_contents)
```

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -636,3 +705,114 @@ def test_get_aml_runs(tmp_path: Path) -> None:
...
+
+    start_time = time.perf_counter()
+    util.download_run_file(run, "dummy_file", output_file_path)
+    end_time = time.perf_counter()
+    time_dont_validate_checksum = end_time - start_time
+
+    assert output_file_path.exists()
+    with open(output_file_path, 'r') as f_path:
+        data = f_path.read()
+        assert data == "Hello world"
```
ant0nsc, line 751:
```suggestion
    assert output_file_path.read_text() == file_contents
```

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -636,3 +705,114 @@ def test_get_aml_runs(tmp_path: Path) -> None:
...
+
+    # Now delete the file and try again with _validate_checksum == True
+    os.remove(output_file_path)
```
ant0nsc, line 754:
```suggestion
    output_file_path.unlink()
```

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -636,3 +705,114 @@ def test_get_aml_runs(tmp_path: Path) -> None:
...
+    assert not output_file_path.exists()
+    start_time = time.perf_counter()
+    util.download_run_file(run, "dummy_file", output_file_path, validate_checksum=True)
+    end_time = time.perf_counter()
+    time_validate_checksum = end_time - start_time
+
+    assert output_file_path.exists()
+    with open(output_file_path, 'r') as f_path:
+        data = f_path.read()
+        assert data == "Hello world"
```
ant0nsc, line 764:
```suggestion
    assert output_file_path.read_text() == file_contents
```

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -636,3 +705,114 @@ def test_get_aml_runs(tmp_path: Path) -> None:
...
+
+    logging.info(f"Time to download file without checksum: {time_dont_validate_checksum} vs time with"
+                 f"validation {time_validate_checksum}.")
+
+
+def _get_file_names(pref: str = "") -> List[str]:
+    file_names = ["somepath.txt", "abc/someotherpath.txt", "abc/def/anotherpath.txt"]
+    if len(pref) > 0:
+        return [u for u in file_names if pref in u]
```
ant0nsc, line 773: startswith rather than "pref in u"?

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -636,3 +705,114 @@ def test_get_aml_runs(tmp_path: Path) -> None:
...
+
+
+def test_get_run_file_names() -> None:
+    with patch("azureml.core.Run") as mock_run:
+        expected_file_names = _get_file_names()
+        mock_run.get_file_names.return_value = expected_file_names
+        # check that we get the expected run paths if no filter is applied
+        run_paths = util.get_run_file_names(mock_run)  # type: ignore
+        assert len(run_paths) == len(expected_file_names)
+        assert sorted(run_paths) == sorted(expected_file_names)
```
ant0nsc, line 785: why that?

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -636,3 +705,114 @@ def test_get_aml_runs(tmp_path: Path) -> None:
...
+
+        # Now check we get the expected run paths if a filter is applied
+        prefix = "abc"
+        expected_file_names_filtered = _get_file_names(pref=prefix)
+        run_paths = util.get_run_file_names(mock_run, prefix=prefix)
+        assert len(run_paths) == len(expected_file_names_filtered)
```
ant0nsc, line 791: How about you work with assertions like `all(f.startswith(prefix) for f in util.get_run_file_names(run, prefix))`? This makes it, in my view, clearer what you are really testing for.

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -636,3 +705,114 @@ def test_get_aml_runs(tmp_path: Path) -> None:
...
+
+
+@pytest.mark.parametrize("prefix", ["", "abc"])
+def test_download_run_files(tmp_path: Path, prefix: str) -> None:
+    # Assert that 'downloaded' paths don't exist to begin with
+    dummy_paths = [x[0] for x in _get_file_names(pref=prefix)]
+    expected_paths = [tmp_path / dummy_path for dummy_path in dummy_paths]
+    assert sum([p.exists() for p in expected_paths]) == 0
```
ant0nsc, line 799: sum(...)==0 is a very numpy-y way of expressing it. `assert not any(...)` is clearer, in my view.

hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -523,3 +550,45 @@ def get_aml_runs(args: Namespace, workspace: Workspace, run_id_source: AzureRunI
...
+    :param prefix: The optional prefix to filter Run files by
+    :return: A list of paths within the Run's container
+    """
+    all_files = run.get_file_names()
+    return [f for f in all_files if f.startswith(prefix)] if prefix else all_files
+
+
+def download_run_files(run: Run, output_dir: Path, prefix: str = "") -> None:
+    """
+    Download all files for a given run, which optionally start with a given prefix
```
ant0nsc, line 569: For both download methods, we should discuss how they are parameterized. At present, most people seem to have trouble instantiating the Run object in the first place.

ant0nsc (changes requested):

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ created.
 ### Changed
 - ([#112](https://github.com/microsoft/hi-ml/pull/112)) Update himl_tensorboard to work with files not in 'logs' directory
 - ([#106](https://github.com/microsoft/hi-ml/pull/106)) Split into two packages. Most of existing package renamed to hi-ml-azure, remained remains hi-ml.
+- ([#113](https://github.com/microsoft/hi-ml/pull/113) Add helper function to download files from AML Run)
```
ant0nsc, line 21: please add to changelog that some functions were moved, and commandline args to script?

hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -405,7 +476,7 @@ def get_most_recent_run_id(run_recovery_file: Path) -> str:
     assert run_recovery_file.is_file(), "When running in cloud builds, this should pick up the ID of a previous \
                                          training run"
     run_id = run_recovery_file.read_text().strip()
-    print(f"Read this run ID from file: {run_id}")
+    logging.info(f"Read this run ID from file: {run_id}. Starting file download")
```
ant0nsc, line 479: "starting file download" does not always apply

hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -476,27 +556,148 @@ def get_latest_aml_runs_from_experiment(args: Namespace, workspace: Workspace) -
     return list(islice(experiment.get_runs(tags=tags), num_runs))
 
 
-def get_aml_runs_from_recovery_ids(args: Namespace, workspace: Workspace) -> List[Run]:
+def get_aml_runs_from_recovery_ids(args: Namespace, aml_workspace: Optional[Workspace] = None,
```
ant0nsc, line 559: there is code duplication - a funciton to get multiple runs should use "get single run" as a building block. 
Also, do we really need two arguments that differ only by the plural "s"? Wouldn't it be more pythonic to have run_recovery_id to accept single or multipe run IDs)

hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -476,27 +556,148 @@ def get_latest_aml_runs_from_experiment(args: Namespace, workspace: Workspace) -
...
+
+def get_aml_run_from_run_id(run_id: str, aml_workspace: Optional[Workspace] = None,
+                            workspace_config_path: Optional[Path] = None) -> Run:
+    """
+    Retrieve an Azure ML Run, firstly by retrieving the corresponding Workspace, and then getting the
+    run according to the specified run_id
+
+    :param run_id: the parameter corresponding to the 'id' property of the Run
+    :param aml_workspace: Optional Azure ML Workspace object
+    :param workspace_config_path: Optional path to a Workspace config file
```
ant0nsc, line 625: Can you add where the config file is found if not provided as arg?

hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -476,27 +556,148 @@ def get_latest_aml_runs_from_experiment(args: Namespace, workspace: Workspace) -
...
+                                 workspace_config_path: Optional[Path] = None) -> Run:
+    """
+    Lookup the run_id arg and then retrieve the Azure ML Run object with this id.
+
+    :param args: Command line args
+    :param aml_workspace: Optional Azure ML Workspace object
+    :param workspace_config_path: Optional path to a Workspace config file
+    :return: The Azure ML Run object with the id as specified by args.run_id
+    """
+    def _get_run_id_from_args(args: Namespace) -> str:
```
ant0nsc, line 659: is it really worth to have a function for that?

hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -476,27 +556,148 @@ def get_latest_aml_runs_from_experiment(args: Namespace, workspace: Workspace) -
...
+        Attempt to retrieve a Run from a given Workspace, specified by run_id.
+
+        :param run_id: the run_id to return
+        :param workspace: the Workspace to search in
+        :return: An Azure ML Run object
+        """
+        try:
+            run = workspace.get_run(run_id)
+        except HttpOperationError as e:
+            if 'was not found' in e.message:
```
ant0nsc, line 639: where's the 'else' part for that if?

hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -476,27 +556,148 @@ def get_latest_aml_runs_from_experiment(args: Namespace, workspace: Workspace) -
...
+    """
+    Retrieve an Azure ML Run, firstly by retrieving the corresponding Workspace, and then getting the
+    run according to the specified run_id
+
+    :param run_id: the parameter corresponding to the 'id' property of the Run
+    :param aml_workspace: Optional Azure ML Workspace object
+    :param workspace_config_path: Optional path to a Workspace config file
+    :return: The Azure ML Run object with the given run_id
+    """
+    def _get_aml_run_from_runid(run_id: str, workspace: Workspace) -> Run:  # pragma: no cover
```
ant0nsc, line 628: is it worth having a function for that?

hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -476,27 +556,148 @@ def get_latest_aml_runs_from_experiment(args: Namespace, workspace: Workspace) -
...
     Retrieve AzureML Runs for each of the Run Ids specified in args.
 
     :param args: command line args including experiment name and number of runs to return
-    :param workspace: AML Workspace
+    :param aml_workspace: Optional Azure ML Workspace object
+    :param workspace_config_path: Optional path containing AML Workspace settings
     :return: List of AML Runs
     """
-    runs = [workspace.get_run(r_id) for r_id in args.run_ids]
+    def _get_run_ids_from_args(args: Namespace) -> List[str]:  # pragma: no cover
```
ant0nsc, line 685: worth having that function?

hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -513,13 +714,104 @@ def get_aml_runs(args: Namespace, workspace: Workspace, run_id_source: AzureRunI
...
+    download_run_files(run, output_dir, prefix=prefix)
+
+
+def download_run_file(run: Run, filename: str, output_path: Path, validate_checksum: bool = False) -> None:
+    """
+    A wrapper around AML Run's download_file method, that handles timeouts
+
+    :param run: The AML Run to download associated file for
+    :param filename: The name of the file as it exists in Azure storage
+    :param output_path: Local path to which the file should be downloaded
```
ant0nsc, line 786: should output_path be a folder or a file? Can we handle both? Should we return the Path to which the file was actually downloaded?

hi-ml-azure/src/health/azure/himl_tensorboard.py
```diff
@@ -142,15 +150,16 @@ def main() -> None:  # pragma: no cover
...
         required=False,
         help="Optional run recovery ids of the runs to plot"
     )
     parser.add_argument(
         "--run_ids",
         default=[],
         nargs="+",
+        required=False,
         help="Optional run ids of the runs to plot"
     )
```
ant0nsc, line 164: Should we merge those args, and accept any run ID, run_recovery ID, or combination thereof? Could also shorten the arg name to "--runs"

hi-ml-azure/tests/testazure/scripts/script_that_downloads_run_files.py
```diff
@@ -0,0 +1,30 @@
...
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--output_path", type=str, required=True)
+
+    args = parser.parse_args()
+    output_path = Path(args.output_path)
+    output_path.mkdir(exist_ok=True)
+    print(f"Created output directory: {output_path}")
+
+    run_ctx = Run.get_context()
```
ant0nsc, line 20: I don't get the purpose of this script - I see it is used in a test, with compute target local. But what is being tested here, because we don't use any of our functions to download, just the API?

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -452,15 +513,31 @@ def test_determine_run_id_source(tmp_path: Path) -> None:
     mock_args = parser.parse_args(["--experiment", "fake_experiment"])
     assert util.determine_run_id_source(mock_args) == util.AzureRunIdSource.EXPERIMENT_LATEST
 
-    # If run recovery id is provided, expect source to be that
-    mock_args = parser.parse_args(["--run_recovery_ids", "experiment:run1234"])
+    parser = ArgumentParser()
+    parser.add_argument("--run_recovery_id", type=str)
+    parser.add_argument("--run_id", type=str)
```
ant0nsc, line 518: This is not an ideal pattern for writing tests. You are defining a parser here, and check if determine_run_id_source still does the right thing. But your actual code uses a parser defined in the download script - if you accidentally made a change to that parser, your test here would still pass, and hence not provide much value. It would be better if you refactored the parser creation in the download script such that the parser creation lives in a function, that you can invoke here in the test to create a parser.

hi-ml-azure/tests/testazure/test_azure_util.py
```diff
@@ -539,100 +613,382 @@ def _get_experiment_runs() -> List[MockRun]:
...
+
+    assert output_file_path.exists()
+    assert output_file_path.read_text() == file_contents
+
+    logging.info(f"Time to download file without checksum: {time_dont_validate_checksum} vs time with"
+                 f"validation {time_validate_checksum}.")
+
+
+def test_download_run_file_during_run(tmp_path: Path) -> None:
+    # This test will create a Run in your workspace (using only local compute)
```
ant0nsc, line 950: As mentioned above, I don't get this test. The script uses the normal API to download files? 
If we wanted to test downloading during an AML run, I would suggest creating a local run first (as you are doing in the test above), and then submit a script to real AML that uses our custom download code, and NOT supply a workspace name. That script should download a file from that previous run, and check its contents, all in AML.

ant0nsc (changes requested):
I have a couple of questions, and maybe suggestions for simplifying the code.

hi-ml-azure/src/health/azure/azure_util.py
mebristo: That is what I originally intended to implement but then there were a few scenarios in which I wanted to call get_aml_run_from_id multiple times whilst iterating over run_ids, hence it made sense to me to create this wrapper function which performs that behavior (although granted this particular function does not call the singular version of the function so I can clean that up).

I also felt like it was clearer to have a separate argument for run_ids (plural) as it explicitly tells the user they can supply more than one id. If we just use the arg run_id we would still have to have a check at the start of each function whether it's a list of ids or a single one, which doesn't seem any neater to me. Happy to discuss further offline!

hi-ml-azure/src/health/azure/azure_util.py
mebristo: I think so, because the default for this arg is an empty list, so this function would return an empty list without complaining

hi-ml-azure/src/health/azure/azure_util.py
mebristo: This function is for downloading a single file so I think output_path should be a  single file. Yes I'll edit it 
 return the path

hi-ml-azure/src/health/azure/himl_tensorboard.py
mebristo: Personally I prefer the idea of keeping them explicit but we can discuss

hi-ml-azure/tests/testazure/scripts/script_that_downloads_run_files.py
mebristo: Yes good catch - I should be calling our own function for download_file in this script. (The point of it it to test that download_file works inside an AML run)

hi-ml-azure/tests/testazure/test_azure_util.py
mebristo: This parser doesn't replicate any parser from any one script - the idea was to add all possible run sources to it -hence why it has both run_id and run_ids which no single script would have

hi-ml-azure/tests/testazure/test_azure_util.py
mebristo: To your first point - that was a mistake. It was supposed to call our own download_files 

To your second point - checking it's contents in AML, meaning that the AML run should fail if the contents are incorrect? What then would this test check for - whether that run succeeds?

hi-ml-azure/src/health/azure/azure_util.py
mebristo: Discussed offline - we will continue exploring this and will create a new PR if we do decide to refactor

hi-ml-azure/tests/testazure/test_azure_util.py
mebristo: Replaced with Jonathan's test script pattern

ant0nsc (approved):

PR #128, opened 2021-10-06
Adding a barrier for downloading in distributed jobs

ant0nsc:


hi-ml-azure/src/health/azure/azure_util.py
```diff
@@ -17,11 +17,13 @@
 
 import conda_merge
 import ruamel.yaml
+from astroid import Module
```
mebristo, line 20: Is this an intentional import or an error?

PR #132, opened 2021-10-11
Profile methods for loading png image files

JonathanTripp:


.vscode/settings.json
```diff
@@ -27,7 +27,7 @@
     "python.linting.mypyArgs": [
         "--config-file=${workspaceFolder}/mypy.ini"
     ],
-    "python.linting.pycodestyleEnabled": true,
+    "python.linting.pycodestyleEnabled": false,
```
ant0nsc, line 30: what's that setting for?

Makefile
```diff
@@ -31,6 +31,7 @@ pip: pip_build pip_test call_pip_local
 # update current conda environment
 conda_update:
 	conda env update -n $(CONDA_DEFAULT_ENV) --file environment.yml
+	conda env update -n $(CONDA_DEFAULT_ENV) --file hi-ml/testhiml/testhiml/utils/image_loading/environment.yml
 	conda env update -n $(CONDA_DEFAULT_ENV) --file hi-ml/testhiml/testhiml/utils/slide_image_loading/environment.yml
```
ant0nsc, line 35: can we keep those two experimental things out of the conda env please? whoever touches these evaluation scripts needs to do due diligence to set up their own conda env.

docs/source/loading_images.md
```diff
@@ -0,0 +1,267 @@
...
+    :param input_filename: Source image file path.
+    :return: torch.Tensor of shape (C, H, W).
+    """
+    torch_tensor = torch.load(input_filename)
+    return torch_tensor
+```
+
+## Results
+
+All the above methods were ran against 122 RGB images, and 122 greyscale images, repeated 10 times. So in total there were 2440 calls to each of the functions. The total times were:
```
ant0nsc, line 252: Are the performance numbers on RGB and greyscale the same? Any change we can have tables for the two image types separately?

ant0nsc (approved):

Makefile
JonathanTripp: They are there to make pyright happy. I can change the makefile so that pyright uses a different conda env

docs/source/loading_images.md
JonathanTripp: Ok, I'll profile separately

ant0nsc (approved):

PR #149, opened 2021-11-02
Add k-fold cross validation wrapper around HyperDrive

mebristo:
Add k-fold cross validation wrapper around HyperDrive, with tests, docs and an example of how to run in AML (see docs/examples/10)


docs/source/examples.md
```diff
@@ -274,3 +274,39 @@ By default, the `hi-ml` package assumes that you supply a commandline argument `
 the commandline) to trigger a submission of the present script to AzureML. If you wish to control it via a different
 flag, coming out of your own argument parser, use the `submit_to_azureml` argument of the function
 `health.azure.himl.submit_to_azure_if_needed`. 
+
+## Training with k-fold cross validation in Azure ML
+The example [sample 10](examples/10/sample.rst) demonstrates how to enable cross validation, by specifying the number
+of splits required.  If this number is > 1, a parent run is instantiated, with a number of child runs equalling the
+specified splits. For example,
```
ant0nsc, line 281: Maybe make that a bit more specific? There is a parent run with 5 child if doing 5-fold CV. First run gets additional arguments `--count=5 --index=0`, ...

docs/source/examples/10/environment.yml
```diff
@@ -0,0 +1,7 @@
+name: iris-env
+dependencies:
+  - pip=20.1.1
+  - python=3.7.3
+  - scikit-learn
+  - pip:
+    - hi-ml-azure>=0.1.0
```
ant0nsc, line 7: >=0.1.10

docs/source/examples/10/inputs.py
```diff
@@ -0,0 +1,48 @@
+#  ------------------------------------------------------------------------------------------
+#  Copyright (c) Microsoft Corporation. All rights reserved.
+#  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+#  ------------------------------------------------------------------------------------------
+#  ------------------------------------------------------------------------------------------
+#  Copyright (c) Microsoft Corporation. All rights reserved.
+#  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+#  ------------------------------------------------------------------------------------------
```
ant0nsc, line 8: duplicate license

docs/source/examples/10/inputs.py
```diff
@@ -0,0 +1,48 @@
...
+
+def main() -> None:
+    # import some data to play with
+    iris = datasets.load_iris()
+    X = iris.data[:, :2]  # we only take the first two features.
+    y = iris.target
+    inputs = Path("inputs")
+    inputs.mkdir(exist_ok=False)
+    X_csv = inputs / "X.csv"
+    np.savetxt(X_csv, X, delimiter=',')
```
ant0nsc, line 26: maybe kill this whole script? Writing a dataset to file is already covered in another example

docs/source/examples/10/sample.py
```diff
@@ -0,0 +1,105 @@
...
+    # if we aren't running inside a child run, create the dataset if it doesn't already exist
+    if args.cross_validation_split_index is None and not Path(train_splits_file).is_file():
+        print("Creating splits")
+        k_folds = KFold(n_splits=int(args.num_cross_validation_splits), shuffle=True, random_state=0)
+        splits = np.array(list(k_folds.split(X)))
+        indices_train_splits, indices_test_splits = [], []
+        for split in splits:
+            indices_train_splits.append(split[0])
+            indices_test_splits.append(split[1])
+        np.savetxt(train_splits_file, np.vstack(indices_train_splits), delimiter=",")
```
ant0nsc, line 77: I think the whole script could be simplified if we don't write any files

docs/source/examples/10/sample.rst
```diff
@@ -0,0 +1,2 @@
+.. literalinclude:: sample.py
+   :emphasize-lines: 28, 62
```
ant0nsc, line 2: are these line numbers correct?

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -164,6 +166,35 @@ def create_run_configuration(workspace: Workspace,
...
+    """
+    # TODO: do we need to ensure that run_config isn't MPIConfig?
+    logging.info(f"Creating a HyperDriveConfig. Please be aware that this expects to find the metric {metric_name}"
+                 f" logged to the Run during your training script.")
+    return HyperDriveConfig(
+        run_config=run_config,
+        hyperparameter_sampling=GridParameterSampling(
+            {
+                "cross_validation_split_index": choice(list(range(num_cross_validation_splits)))
+            }),
```
ant0nsc, line 191: I think cross_validation_split_index should be an argument

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -295,7 +326,10 @@ def submit_to_azure_if_needed(  # type: ignore
         submit_to_azureml: Optional[bool] = None,
         tags: Optional[Dict[str, str]] = None,
         after_submission: Optional[Callable[[Run], None]] = None,
-        hyperdrive_config: Optional[HyperDriveConfig] = None) -> AzureRunInfo:  # pragma: no cover
+        hyperdrive_config: Optional[HyperDriveConfig] = None,
+        num_cross_validation_splits: int = 1,
+        cross_validation_metric_name: str = "val/loss"
```
ant0nsc, line 331: I'm not sure we should bloat the interface of submit_if_needed even more. in addition, we need one more argument that is the name of the commandline arg. It would be just as good if we suggest to users to pass in the result of your helper function in hyperdrive_config.

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1252,3 +1255,51 @@ def torch_barrier() -> None:
...
+    :param run_id: The run id (type: str) of the parent run. The type of this run must be an AML HyperDriveRun
+    :param workspace: An AzureML Workspace object which contains the run with this id
+    :return: A Pandas DataFrame containing the aggregated metrics from each child run
+    """
+    run = get_aml_run_from_run_id(run_id, aml_workspace=workspace)
+    assert isinstance(run, HyperDriveRun)
+    metrics: DefaultDict = defaultdict()
+    for child_run in run.get_children():
+        child_run_metrics = child_run.get_metrics()
+        child_run_crossval_split = json.loads(child_run.tags.get('hyperparameters')).get('cross_validation_split_index')
```
ant0nsc, line 1298: cross_validation_split_index should not be hardcoded

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1252,3 +1255,51 @@ def torch_barrier() -> None:
...
+    For a given HyperDrive run id, retrieves the metrics from each of its children and then aggregates it.
+    Returns a DataFrame where each column is one child run, and each row is a metric logged by that child run.
+    For example, for a HyperDrive run with 2 children, where each logs epoch, accuracy and loss, the result
+    would look like:
+
+    |              | 0               | 1                  |
+    |--------------|-----------------|--------------------|
+    | epoch        | [1, 2, 3]       | [1, 2, 3]          |
+    | accuracy     | [0.7, 0.8, 0.9] | [0.71, 0.82, 0.91] |
+    | loss         | [0.5, 0.4, 0.3] | [0.45, 0.37, 0.29] |
```
ant0nsc, line 1271: your examples assume that all metrics have the same length. What happens if that is not the case?

ant0nsc: Maybe it would be better to return a dictionary mapping from metric name to dataframe?

hi-ml-azure/src/health_azure/utils.py
ant0nsc: Actually - to make that even more generic, can we read out ALL hyperparameters, and attach them as columns to the dataframe?

ant0nsc (changes requested):
Looking good, some small comments, and one bigger ask: Should we rather take all crossval-related out of the argument list to submit_if_needed, and expect people to pass in the result of calling your wrapper function?

docs/source/examples/10/sample.py
mebristo: I'm not sure where else to make the splits? I don't want the child runs to make splits becuase then they will each be using different versions of splits.. so the purpose of this was to make sure we split the data once first and then put it somewhere where all child runs could access it

docs/source/examples/10/sample.rst
mebristo: Nope they were out of date. Updated now

hi-ml-azure/src/health_azure/utils.py
mebristo: (discussed offline)

hi-ml-azure/src/health_azure/utils.py
mebristo: (discussed offline)

docs/source/examples/10/sample.py
mebristo: Solved - have re-created the inputs.py file to handle this

docs/source/examples.md
```diff
@@ -274,3 +274,87 @@ By default, the `hi-ml` package assumes that you supply a commandline argument `
...
+    
+     hyperdrive_config = create_crossval_hyperdrive_config(num_cross_validation_splits,
+                                                           cross_val_split_name=cross_val_split_name,
+                                                           metric_name=metric_name)
+    ```
+    where:
+    - `num_cross_validation_splits` is the number of k-fold cross validation splits you require
+    - `cross_val_split_name` is the name of the argument given to each child run, whose value denotes which split that child represents (this parameter
+    defaults to 'cross_validation_split_index', in which case, supposing you specified 2 cross validation splits, one would 
+    receive the arguments ['cross_validation_split_index' '0'] and the other would receive ['cross_validation_split_index' '1']].
```
ant0nsc, line 296: Is a "--" added automagically?

docs/source/examples.md
```diff
@@ -274,3 +274,87 @@ By default, the `hi-ml` package assumes that you supply a commandline argument `
...
+    ```python
+    from health_azure import create_crossval_hyperdrive_config
+    
+     hyperdrive_config = create_crossval_hyperdrive_config(num_cross_validation_splits,
+                                                           cross_val_split_name=cross_val_split_name,
+                                                           metric_name=metric_name)
+    ```
+    where:
+    - `num_cross_validation_splits` is the number of k-fold cross validation splits you require
+    - `cross_val_split_name` is the name of the argument given to each child run, whose value denotes which split that child represents (this parameter
```
ant0nsc, line 294: nitpicking - should we shorten the argument name? cross_val_count? The other one should more clealry say it's a commandline arg, maybe cross_val_index_arg_name?

docs/source/examples.md
```diff
@@ -274,3 +274,87 @@ By default, the `hi-ml` package assumes that you supply a commandline argument `
 the commandline) to trigger a submission of the present script to AzureML. If you wish to control it via a different
 flag, coming out of your own argument parser, use the `submit_to_azureml` argument of the function
 `health.azure.himl.submit_to_azure_if_needed`. 
+
+## Training with k-fold cross validation in Azure ML
+To train with k-fold cross validation, you must do two things.
```
ant0nsc, line 279: Maybe introduce the concept of a parent run and child runs here first, pointing to AML docu?

docs/source/examples.md
```diff
@@ -274,3 +274,87 @@ By default, the `hi-ml` package assumes that you supply a commandline argument `
...
+    )
+    ```
+    This will create a parent (HyperDrive) Run, with `num_cross_validation_split` children - each one associated with a different data split.
+
+## Retrieving the aggregated results of a cross validation/ HyperDrive run
+
+You can retrieve a Pandas DataFrame of the aggregated results from your cross validation run as follows: 
+
+```python
+from health_azure import utils
```
ant0nsc, line 329: add the aggregate function to utils/__init__, and change the import here?

docs/source/examples.md
```diff
@@ -274,3 +274,87 @@ By default, the `hi-ml` package assumes that you supply a commandline argument `
...
+
+df = utils.aggregate_hyperdrive_metrics(run.id, workspace)
```
ant0nsc, line 331: can we have a more friendly overload that accepts a string runID, and auto fills in the workspace?

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -164,6 +166,35 @@ def create_run_configuration(workspace: Workspace,
...
 
 
+def create_crossval_hyperdrive_config(num_cross_validation_splits: int,
+                                      cross_val_split_name: str = "cross_validation_split_index",
+                                      metric_name: str = "val/loss") -> HyperDriveConfig:
+    """
+    Creates an Azure ML HyperDriveConfig object for running cross validation. Note: this config expects a metric
+    named <metric_name> to be logged in your training script([see here](
+    https://docs.microsoft.com/en-us/azure/machine-learning/how-to-tune-hyperparameters#log-metrics-for-hyperparameter-tuning))
+    :param num_cross_validation_splits: The number of splits for k-fold cross validation
```
ant0nsc, line 176: blank line missing before :param

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -164,6 +166,35 @@ def create_run_configuration(workspace: Workspace,
...
+    :param cross_val_split_name: The name of the argument received by each of the child runs that indicates which
+        split that child represents.
+    :param metric_name: The name of the metric that the HyperDriveConfig will compare runs by. Please note that it is
+        your responsibility to make sure a metric with this name is logged to the Run in your training script
+    :return: an Azure ML HyperDriveConfig object
+    """
+    # TODO: do we need to ensure that run_config isn't MPIConfig?
```
ant0nsc, line 183: TODO?

hi-ml-azure/src/health_azure/utils.py
ant0nsc: There is something to be done where still. Your code here will only work if people go with the default argument name. Also, it will fail for any more generic type of hyperdrive run.

ant0nsc (changes requested):

docs/source/examples.md
mebristo: Sure - the function name already specified cross validation so maybe the arg name should be 'num_splits'?

docs/source/examples.md
mebristo: Yes, updated to reflect this

ant0nsc (approved):

PR #174, opened 2021-12-20
Fixing checkpoint downloader

vale-salvatelli:
This PR contains changes to the class CheckpointDownloader to fix a bug with the hierarchies of the downloaded folders.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ the section headers (Added/Changed/...) and incrementing the package version.
 - ([#161](https://github.com/microsoft/hi-ml/pull/161)) Empty string as target folder for a dataset creates an invalid mounting path for the dataset in AzureML (fixes #160)
 - ([#167](https://github.com/microsoft/hi-ml/pull/167)) Fix bugs in logging hyperparameters: logging as name/value
   table, rather than one column per hyperparameter. Use string logging for all hyperparameters
+- ([#174](https://github.com/microsoft/hi-ml/pull/174)) Fix bugs in downloading checkpoint from AML run
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ the section headers (Added/Changed/...) and incrementing the package version.
 - ([#161](https://github.com/microsoft/hi-ml/pull/161)) Empty string as target folder for a dataset creates an invalid mounting path for the dataset in AzureML (fixes #160)
 - ([#167](https://github.com/microsoft/hi-ml/pull/167)) Fix bugs in logging hyperparameters: logging as name/value
   table, rather than one column per hyperparameter. Use string logging for all hyperparameters
-- ([#174](https://github.com/microsoft/hi-ml/pull/174)) Fix bugs in downloading checkpoint from AML run
+- ([#174](https://github.com/microsoft/hi-ml/pull/174)) Fix bugs in returned local_checkpoint_path when downloading checkpoints from AML run
 
 ### Removed
 
--- a/hi-ml-azure/src/health_azure/utils.py
+++ b/hi-ml-azure/src/health_azure/utils.py
@@ -619,7 +619,7 @@ def get_workspace(aml_workspace: Optional[Workspace] = None, workspace_config_pa
         auth = get_authentication()
         return Workspace.from_config(path=str(workspace_config_path), auth=auth)
     else:
-        raise ValueError("Workspace config file does not valid.")
+        raise ValueError("Workspace config file is not valid.")
 
 
 def create_run_recovery_id(run: Run) -> str:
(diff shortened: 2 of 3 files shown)


hi-ml-azure/src/health_azure/utils.py
```diff
@@ -467,7 +467,7 @@ def from_string(self, x: str) -> List[str]:
 class CheckpointDownloader:
     def __init__(self, run_id: str, checkpoint_filename: str, azure_config_json_path: Path = None,
                  aml_workspace: Workspace = None, download_dir: PathOrString = "checkpoints",
-                 remote_checkpoint_folder: PathOrString = "checkpoints") -> None:
+                 remote_checkpoint_dir: PathOrString = "checkpoints") -> None:
```
vale-salvatelli, line 470: Renamed for consistency with the other variables folder = dir

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -480,25 +480,29 @@ def __init__(self, run_id: str, checkpoint_filename: str, azure_config_json_path
...
+        return self.download_dir / run_id 
 
     @property
     def remote_checkpoint_path(self) -> Path:
-        return self.remote_checkpoint_folder / self.checkpoint_filename
+        return self.remote_checkpoint_dir / self.checkpoint_filename
+    
+    @property
+    def local_checkpoint_path(self) -> Path:
+        return self.local_checkpoint_dir / self.remote_checkpoint_path
```
vale-salvatelli, line 505: this is the core of the change - the underlying function is downloading the full hierarchy of folders from the run

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -608,10 +612,11 @@ def get_workspace(aml_workspace: Optional[Workspace] = None, workspace_config_pa
         else:
             raise ValueError("No workspace config file given, nor can we find one.")
 
-    if workspace_config_path.is_file():
+    if workspace_config_path:
```
Shruthi42, line 615: Why are we removing the check for is_file() here? Before this we already handle the case where it is None, so it makes sense to check that the file exists at this stage.

CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ the section headers (Added/Changed/...) and incrementing the package version.
 - ([#161](https://github.com/microsoft/hi-ml/pull/161)) Empty string as target folder for a dataset creates an invalid mounting path for the dataset in AzureML (fixes #160)
 - ([#167](https://github.com/microsoft/hi-ml/pull/167)) Fix bugs in logging hyperparameters: logging as name/value
   table, rather than one column per hyperparameter. Use string logging for all hyperparameters
+- ([#174](https://github.com/microsoft/hi-ml/pull/174)) Fix bugs in downloading checkpoint from AML run
```
Shruthi42, line 45: Could we change this to pinpoint the exact issue? Perhaps something like "Fix bug in returned local_checkpoint_path when downloading checkpoints from AML run".

hi-ml-azure/src/health_azure/utils.py
vale-salvatelli: I agree with you. But for a reason I don't understand the is_file() raises an error "AttributeError: 'Workspace' object has no attribute 'is_file'". If I breakpoint and check the type returned workspace_config_path is indeed a Path - and Workspace.from_config(path=str(workspace_config_path), auth=auth) works - and it should not if workspace_config_path is not a path. This is the only solution I found to move forward

CHANGELOG.md
vale-salvatelli: Done!

hi-ml-azure/src/health_azure/utils.py
javier-alvarez: Do you have repro steps for this? It seems like a different issue than the description of the PR

vale-salvatelli: Agreed is a different issue (and I am not fixing it just working around it). I can reproduce the issue - tried a few times - but run out of ideas on how to address. How you suggest to move forward? 

I am primarily trying to work on the DeepMIL PANDA pipeline and this was one of the many things broken - happy to leave this as unmerged branch for now if you think it's not a good enough solution

vale-salvatelli: I added this condition `if isinstance(workspace_config_path, Path) and not workspace_config_path.is_file():` to cover for the case a path is passed but the path does not exist

vale-salvatelli: Discussed at voice and reverted this change plus added coverage for wrong type in input

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1046,6 +1053,7 @@ def get_run_file_names(run: Run, prefix: str = "") -> List[str]:
     :return: A list of paths within the Run's container
     """
     all_files = run.get_file_names()
+    print(f"Selecting files with prefix {prefix}")
```
Shruthi42, line 1056: Did you leave this print statement in by accident?

Shruthi42 (approved):

hi-ml-azure/src/health_azure/utils.py
vale-salvatelli: No, I made it a bit nicer and left it on purpose - I found it's useful to know which prefix is used to select among run files

PR #176, opened 2022-01-06
Updated augmentations

maxilse:
1. Fixed HEDJitter augmentation, the previous version was also jittering the D channel
2. Removed dependency of StainNormalization on skimage, which makes it also faster
3. Added GaussianBlur that is faster than the torchvision version
4. Added rotation by multiples of 90 degrees

Adjusted/added the test for all augmentations.
All the above augmentations are frequently used for histopathology.

Added opencv (cv2) and torchvision to the environment since they are necessary for the augmentations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,13 @@ the section headers (Added/Changed/...) and incrementing the package version.
 ## [Unreleased]
 
 ### Added
+- ([#176](https://github.com/microsoft/hi-ml/pull/176)) Add GaussianBlur and RotationByMultiplesOf90 augmentations. Added torchvision and opencv to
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,22 @@ For each Pull Request, the affected code parts should be briefly described and a
 release. In the first PR after a release has been made, a section for the upcoming release should be added, by copying
 the section headers (Added/Changed/...) and incrementing the package version.
 
+## [Unreleased]
```
ant0nsc, line 13: Change `unreleased` to 0.1.14 please

ant0nsc (approved):

PR #193, opened 2022-02-25
Adding transformation adaptor

vale-salvatelli:
PR to add a transformation adaptor to allow transformations working only on Tensors/Images to work with a dict in input

### DIFF
--- a/hi-ml-histopathology/src/histopathology/models/transforms.py
+++ b/hi-ml-histopathology/src/histopathology/models/transforms.py
@@ -4,7 +4,7 @@
 #  ------------------------------------------------------------------------------------------
 
 from pathlib import Path
-from typing import Mapping, Sequence, Union
+from typing import Mapping, Sequence, Union, Callable, Dict
 
 import torch
 import numpy as np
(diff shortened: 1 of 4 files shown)


maxilse (approved):

hi-ml-histopathology/src/histopathology/models/transforms.py
```diff
@@ -43,6 +43,36 @@ def load_image_stack_as_tensor(image_paths: Sequence[PathOrString],
...
 
+def transform_dict_adaptor(function: Callable, k_input: str = None, k_output: str = None) -> Callable:
+    """Adapt transformations to work with an input dictionary (rather than a tensor).
+       We can't reuse monai.transforms.adaptors because it is only compatible with transformations that accept
+       a dict as input.
+
+    :param function: a transformation function
+    :param k_input: key of the input dictionary that contains the object
+        to which function should be applied
+    :param k_output: key of the input dictionary where to place the function output. If None
```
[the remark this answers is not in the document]
vale-salvatelli: good catch. fixed now

hi-ml-histopathology/src/histopathology/models/transforms.py
```diff
@@ -43,6 +43,37 @@ def load_image_stack_as_tensor(image_paths: Sequence[PathOrString],
...
+    def _inner(ditems: dict) -> Dict:
+        if k_input is None:
+            dinputs = ditems
+        else:
+            dinputs = ditems[k_input]
+        ret = function(dinputs)
+        if k_output is None:
+            ditems = ret
+        else:
+            if isinstance(ret, type(ditems[k_output])):
```
dccastro, line 68: For my information, I'm not sure I understand this. Is the dictionary expected to already contain `k_output`? And why do the types need to match? This seems overly restrictive—consider e.g. loading an image tensor from a path.

Another comment is that the MONAI docs suggest that transforms shouldn't modify input dictionaries in-place, but rather call `ditems.copy()` before assigning the outputs.

dccastro (approved):

hi-ml-histopathology/src/histopathology/models/transforms.py
vale-salvatelli: I think they're both good comments:
- correct we could extend this adaptor to be compatible with a broader set of transformations and also give the option to add a key rather than just overwriting. I wanted to keep this as simple as possible to solve our current case.
- do you know why MONAI suggest to copy? here I am a bit worried about memory consumption if every transformation is creating a copy but happy to follow up on this

harshita-s (approved):

hi-ml-histopathology/src/histopathology/models/transforms.py
dccastro: On the second point, it's just a shallow copy of the dictionary, in case the original needs to be used elsewhere (e.g. if passed to multiple parallel transformations).

PR #220, opened 2022-03-07
FIX: Config loader fails when using relative directories

ant0nsc:
Config loader fails depending on whether runner is called via path that involves `..` or not.

Bug fix is the change in `runner.py`, all other changes are just places where we might get wrong results because of `Path.absolute`

Other changes improve search for conda environment files in corner cases

Please follow the guidelines for PRs contained [here](docs/source/contributing.md). Checklist:

- [ ] Ensure that your PR is small, and implements one change.
- [ ] Add unit tests for all functions that you introduced or modified.
- [ ] Run PyCharm's code cleanup tools on your Python files.
- [ ] Ensure that documentation renders correctly in Sphinx (run Sphinx via `make html` in the `docs folder)
- [ ] Link the correct GitHub issue for tracking.
- [ ] When merging your PR, **replace the default merge message** with a brief description of your PR,
and if needed a motivation why that change was required.

### DIFF
--- a/hi-ml-azure/src/health_azure/paths.py
+++ b/hi-ml-azure/src/health_azure/paths.py
@@ -18,7 +18,7 @@ def is_himl_used_from_git_repo() -> bool:
 
     :return: False if HI-ML is installed as a package, True if used via source from git.
     """
-    health_ml_root = Path(__file__).parent.parent
+    health_ml_root = Path(__file__).resolve().parent.parent
     logging.debug(f"health_ml root: {health_ml_root}")
     if health_ml_root.parent.stem == "site-packages":
         return False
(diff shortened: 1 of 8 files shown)


hi-ml-azure/src/health_azure/paths.py
```diff
@@ -44,4 +44,5 @@ def git_repo_root_folder() -> Path:
     """
     if not is_himl_used_from_git_repo():
         raise ValueError("This function can only be used if the HI-ML package is used directly from the git repo.")
-    return Path(__file__).parent.parent.parent.parent
+    current_file = Path(__file__).resolve()
```
mebristo, line 47: Think the docstring needs to be updated here - if hi-ml has been installed as a pip package this will raise an error, rather than return None

mebristo (approved):

Issue #231, opened 2022-03-11
Many warnings are shown when build the docs


fepegar:
```
cp ../CHANGELOG.md source/
Running Sphinx v4.1.2
loading pickled environment... done
myst v0.15.2: MdParserConfig(renderer='sphinx', commonmark_only=False, enable_extensions=['dollarmath'], dmath_allow_labels=True, dmath_allow_space=True, dmath_allow_digits=True, dmath_double_inline=False, update_mathjax=True, mathjax_classes='tex2jax_process|mathjax_process|math|output_area', disable_syntax=[], url_schemes=['http', 'https', 'mailto', 'ftp'], heading_anchors=None, heading_slug_func=None, html_meta=[], footnote_transition=True, substitutions=[], sub_delimiters=['{', '}'], words_per_minute=200)
[autosummary] generating autosummary for: CHANGELOG.md, api/api.rst, api/health_azure.AzureRunInfo.rst, api/health_azure.DatasetConfig.rst, api/health_azure.aggregate_hyperdrive_metrics.rst, api/health_azure.create_aml_run_object.rst, api/health_azure.create_crossval_hyperdrive_config.rst, api/health_azure.create_run_configuration.rst, api/health_azure.create_script_run.rst, api/health_azure.download_checkpoints_from_run_id.rst, ..., first_steps.md, histopathology.md, hyperdrive.md, index.rst, loading_images.md, logging.md, lowpriority.md, runner.md, self_supervised_models.md, whole_slide_images.md
[automodsumm] api/api.rst: found 23 automodsumm entries to generate
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 1 source files that are out of date
updating environment: 0 added, 15 changed, 0 removed
reading sources... [100%] index
/home/fernando/git/hi-ml/hi-ml-azure/src/health_azure/utils.py:docstring of health_azure.utils.aggregate_hyperdrive_metrics:7: WARNING: Line block ends without a blank line.
/home/fernando/git/hi-ml/hi-ml-azure/src/health_azure/utils.py:docstring of health_azure.utils.aggregate_hyperdrive_metrics:19: WARNING: Line block ends without a blank line.
/home/fernando/git/hi-ml/hi-ml-azure/src/health_azure/utils.py:docstring of health_azure.utils.aggregate_hyperdrive_metrics:19: WARNING: Inline substitution_reference start-string without end-string.
[...]

PR #232, opened 2022-03-11
DOC: Fix some Sphinx warnings

fepegar:
Many warnings are printed out when building the Sphinx docs, as shown in
- #231

This PR fixes two of the three issues.

Before:

Capture2

After:

Capture


docs/Makefile
```diff
@@ -18,4 +18,5 @@ help:
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
 	cp ../CHANGELOG.md source/
+	cp CONTRIBUTING.md source/
```
mebristo, line 21: Does this file currently exist? Should this be `../CONTRIBUTING.md` if we are moving it to the root? Also I don't believe we currently have this file at the root level, so would you alsoneed to move that as part of this PR?

fepegar: > Does this file currently exist?

Yes, it's in `hi-ml/docs/`.

> Should this be ../CONTRIBUTING.md if we are moving it to the root?

If we do, yes. But I wanted to ask your opinion before moving it. Shall I?

mebristo: I see no reason not to. @ant0nsc?

ant0nsc (approved):

docs/Makefile
ant0nsc: fine to move, sure.

PR #227, opened 2022-03-09
ENH: Add transformer pooling

maxilse:
The main thing this PR does is to add TransformerPooling. However, I moved the constructer of the pooling layers outside of DeepMIL to allow for more flexibility in the constructor arguments. This required changing some tests.

The transformer pooling is based on the pytorch Transformer class, however, in the pytorch implementation you have to pass a flag (need_weights) to the forward (??? like why not the init???) of the MultiheadAttention class. See line 186 of attention_layer.py

One open question is: Does the rescaling of the attention weights in forward() of TransformerPooling sense? See line 244 of attention_layer.py.

PR #233, opened 2022-03-14
ENH: Multiclass AUROC, metrics test, and max pooling in DeepMIL

harshita-s:
This PR adds to DeepMIL:
1) AUROC as a multiclass metric
2) Tests for multiclass metrics
3) Max pooling layer


hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
```diff
@@ -208,41 +221,12 @@ def test_metrics() -> None:
 
     for key, metric_obj in module_metrics_dict.items():
         value = metric_obj.compute()
-        expected_value = independent_metrics_dict[key](predicted_probs, true_labels)
+        if n_classes > 1:
+            expected_value = independent_metrics_dict[key](predicted_probs, true_labels.squeeze())
```
dccastro, line 225: I feel like a `squeeze()` shouldn't be necessary in this case. If it is, it suggests the shapes could be handled more cleanly upstream. I'd be in favour of making the shapes consistent everywhere, and special-casing on `n_classes` only where strictly necessary. This should make our lives easier by limiting code complexity and reducing chance of bugs. What do you think?

dccastro: This is also related to the other comment in `deepmil.py`.

hi-ml/src/health_ml/networks/layers/attention_layers.py
```diff
@@ -24,6 +24,21 @@ def forward(self, features: Tensor) -> Tuple[Tensor, Tensor]:
         return (attention_weights, pooled_features)
 
 
+class MaxPoolingLayer(nn.Module):
+    """Max pooling returns uniform weights and the maximum feature vector over the first axis"""
+
+    # args/kwargs added here for compatibility with parametrised pooling modules
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
dccastro, line 31: This empty constructor is no longer necessary after Max's PR https://github.com/microsoft/hi-ml/pull/227.

hi-ml/src/health_ml/networks/layers/attention_layers.py
```diff
@@ -24,6 +24,21 @@ def forward(self, features: Tensor) -> Tuple[Tensor, Tensor]:
...
+        super().__init__()
+
+    def forward(self, features: Tensor) -> Tuple[Tensor, Tensor]:
+        num_instances = features.shape[0]
+        attention_weights = torch.full((1, num_instances), 1. / num_instances)
```
dccastro, line 36: [Speculative] I understand there's no obvious way to assign instance attentions when the `max` is taken dimension-wise. Do you think it could it make sense to extract a normalised 'histogram' of selected indices? I.e. the attention weight of each instance being proportional to the number of dimensions for which its value was selected (e.g. using `torch.bincount(indices, minlength=num_dimensions)`). Might this be useful as a proxy?

hi-ml/src/health_ml/networks/layers/attention_layers.py
```diff
@@ -24,6 +24,21 @@ def forward(self, features: Tensor) -> Tuple[Tensor, Tensor]:
         return (attention_weights, pooled_features)
 
 
+class MaxPoolingLayer(nn.Module):
```
dccastro, line 27: Could you add a test for this class in `test_attentionlayers.py`, similar to `test_mean_pooling()`?

hi-ml/src/health_ml/networks/layers/attention_layers.py
dccastro: In any case, it would be helpful to add a comment here to clarify that the weights currently returned are arbitrarily uniform because max-pooling works per dimension.

hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
```diff
@@ -142,7 +151,7 @@ def validate_metric_inputs(scores: torch.Tensor, labels: torch.Tensor) -> None:
     def is_integral(x: torch.Tensor) -> bool:
         return (x == x.long()).all()  # type: ignore
 
-    assert scores.shape == labels.shape
+    assert scores.shape[0] == labels.shape[0]
```
dccastro, line 154: Related to my other comments, what would you think of writing instead
```python
assert labels.shape == (scores.shape[0],)
```
to enforce that `labels` is always a 1D tensor?

dccastro:
Good work adding multi-class support! My main feedback is that the `DeepMILModule` and related tests could be simpler if we ensured our labels and predictions tensors had consistent shapes across binary and multi-class cases (limiting the number of `if n_classes > 1: ...`). My suggestion is labels should always be `(N,)` and predictions (logits or probabilities) should be `(N,)` for binary and `(N, C)` for multi-class, which is consistent with [Torchmetrics' expected input shapes](https://torchmetrics.readthedocs.io/en/latest/references/modules.html#input-types).

hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -248,7 +249,10 @@ def _shared_step(self, batch: Dict, batch_idx: int, stage: str) -> Dict[ResultsK
         results = dict()
         for metric_object in self.get_metrics_dict(stage).values():
             if self.n_classes > 1:
-                metric_object.update(predicted_probs, bag_labels.squeeze())
+                if bag_labels.shape == torch.Size([1, 1]):
+                    metric_object.update(predicted_probs, bag_labels.squeeze(0))
+                else:
+                    metric_object.update(predicted_probs, bag_labels.squeeze())
```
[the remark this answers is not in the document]
harshita-s: Thanks for the suggestion. I have now made this change. There is an edge case with n_classes=1 and batch_size=1 which I have considered as well.

hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
```diff
@@ -106,15 +106,24 @@ def _test_lightningmodule(
...
         if metric_name == MetricsKey.CONF_MATRIX:
             continue
-        score = metric_object(preds.view(-1, 1), bag_labels.view(-1, 1))
+        if n_classes > 1:
+            if bag_labels.shape == torch.Size([1, 1]):
+                score = metric_object(probs, bag_labels.squeeze(0))
+            else:
+                score = metric_object(probs, bag_labels.squeeze())
+        else:
+            score = metric_object(probs.view(-1, 1), bag_labels)
```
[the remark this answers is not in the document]
harshita-s: Thanks , I have made this change

hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
harshita-s: I agree, I have tried to limit the special-casing on `n_classes`

hi-ml/src/health_ml/networks/layers/attention_layers.py
harshita-s: I have removed this now, thanks.

hi-ml/src/health_ml/networks/layers/attention_layers.py
harshita-s: It is a good idea to add weights based on how many times the maximum value was selected. I have changed it in the code, please have a look

hi-ml/src/health_ml/networks/layers/attention_layers.py
harshita-s: Done, thanks

PR #253, opened 2022-03-18
BUG: Pin package versions for histopathology

ant0nsc:
Pin package versions in hi-ml-histopathology/environment.yml to avoid environment discrepancies.

### DIFF
--- a/hi-ml-histopathology/environment.yml
+++ b/hi-ml-histopathology/environment.yml
@@ -11,7 +11,23 @@ dependencies:
   - python-blosc==1.7.0
   - torchvision=0.11.1
   - pip:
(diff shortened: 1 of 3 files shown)


javier-alvarez (approved):

javier-alvarez (commented):
Change description checklist

hi-ml-histopathology/environment.yml
```diff
@@ -39,8 +52,6 @@ dependencies:
       - six==1.15.0
       - stopit==1.1.2
       - tabulate==0.8.7
-      - torch>=1.10.0
```
vale-salvatelli, line 42: this dependency seems dropped - was this intentional?

PR #254, opened 2022-03-21
ENH: Linearly scale the learning rate of ssl models

maxilse:
As seen in https://arxiv.org/abs/1706.02677, we automatically rescale the learning rate by the number of GPUs to train.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Each release contains a link for "Full Changelog"
 ## 0.1.14
 
 ### Added
-
+- ([#254](https://github.com/microsoft/hi-ml/pull/254)) Automatically rescale the learning rate by the number of GPUs available.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,7 +10,7 @@ Each release contains a link for "Full Changelog"
 ## 0.1.14
 
 ### Added
-
+- ([#254](https://github.com/microsoft/hi-ml/pull/254)) Automatically rescale the learning rate by the number of GPUs available.
```
ant0nsc, line 13: Just FYI, we don't need to update changelog anymore

ant0nsc (dismissed):

CHANGELOG.md
maxilse: shall I remove this then? Or can it stay?

ant0nsc (approved):

PR #256, opened 2022-03-21
BUG: Perform validation and test on full slide

harshita-s:
In this PR, `max_bag_size_inf` is added to the data module parameters of `TilesDataModule` and passed in `BaseMIL`. It is the upper bound on number of tiles in each slide during validation and test stages. If 0 (default), it will return all tiles in the slide. It can be configured from the individual experimental configs based on the slide datasets and available memory.

### DIFF
--- a/hi-ml-histopathology/src/histopathology/datamodules/base_module.py
+++ b/hi-ml-histopathology/src/histopathology/datamodules/base_module.py
@@ -48,7 +48,7 @@ class TilesDataModule(LightningDataModule):
         will return all samples in each bag. If > 0 , bags larger than `max_bag_size` will yield
         random subsets of instances.
         :param batch_size: Number of slides to load per batch.
-        :param max_bag_size_inf: Upper bound on number of tiles in each loaded bag during validation and test stages. 
+        :param max_bag_size_inf: Upper bound on number of tiles in each loaded bag during validation and test stages.
         If 0 (default), will return all samples in each bag. If > 0 , bags larger than `max_bag_size_inf` will yield
         random subsets of instances.
         :param seed: pseudorandom number generator seed to use for shuffling instances and bags. Note that randomness in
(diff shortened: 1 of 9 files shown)


hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -121,9 +126,15 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
                 return torch.load(f, map_location=memory_location)
 
         generator = _create_generator(self.seed)
+
+        if stage in ['val', 'test']:
+            eff_max_bag_size = self.max_bag_size_inf
```
vale-salvatelli, line 131: For the future it might be convenient to decouple also validation and test phase. Validate on the full slide each time might be time consuming, this is why it might be advantageous to have the option to do full slide only for the testing phase.

vale-salvatelli:
The change looks good. Can we extend some test to include this parameter? At least checking that if we change it in the config, the number of tiles that are loaded are indeed different?

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
harshita-s: Thanks, we could decouple this in future.

hi-ml-histopathology/testhisto/testhisto/datamodules/test_datamodule_caching.py
```diff
@@ -177,3 +180,34 @@ def test_tile_id_coverage(mock_data_dir: Path, cache_mode: CacheMode, precache_l
...
+
+
+@pytest.mark.parametrize('cache_mode, precache_location, cache_dir_provided',
+                         [(CacheMode.DISK, CacheLocation.SAME, True),
+                          (CacheMode.DISK, CacheLocation.CPU, True),
+                          (CacheMode.MEMORY, CacheLocation.SAME, True),
+                          (CacheMode.MEMORY, CacheLocation.CPU, True),
+                          (CacheMode.MEMORY, CacheLocation.NONE, False),
+                          (CacheMode.NONE, CacheLocation.NONE, False)
+                          ])
```
vale-salvatelli, line 198: The cache consistency is tested above. For this specific test why we need to parametrize over all the options? Shouldn't this be independent of the cache mode?

vale-salvatelli (approved):

PR #257, opened 2022-03-21
ENH: Better handling of duplicates when merging conda envs

mebristo:
If duplicates are encountered when merging conda environment files, if exactly one such package is pinned ('==' for pip packages and '=' for conda), use that. Otherwise, raise an exception.

### DIFF
--- a/hi-ml-azure/src/health_azure/utils.py
+++ b/hi-ml-azure/src/health_azure/utils.py
@@ -10,6 +10,7 @@ import json
 import logging
 import os
 import re
+import shutil
 import sys
 import tempfile
 from argparse import (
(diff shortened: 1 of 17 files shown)


hi-ml-azure/src/health_azure/utils.py
```diff
@@ -895,54 +896,125 @@ def _log_conda_dependencies_stats(conda: CondaDependencies, message_prefix: str)
...
+    def __init__(self, dependency_str: str) -> None:
+        self.package_name = ""
+        self.operator = ""
+        self.version = ""
+        self._split_dependency_str(dependency_str)
+
+    def _split_dependency_str(self, dependency_str: str) -> None:
+        parts = _split_dependency(dependency_str)
+        self.package_name = parts[0]
+        if len(parts) > 0:
```
ant0nsc, line 924: I think this should be len(parts) > 1?

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -895,54 +896,125 @@ def _log_conda_dependencies_stats(conda: CondaDependencies, message_prefix: str)
...
+        if dependency_name in all_deps:
+            all_deps[dependency_name].append(dependency)
         else:
-            unique_deps[dep_name] = (dep_version, dep_join)
+            all_deps[dependency_name] = [dependency]
 
-    unique_deps_list = [f"{pkg}{joiner}{vrsn}" for pkg, (vrsn, joiner) in unique_deps.items()]
+    unique_deps: List[PackageDependency] = _resolve_dependencies(all_deps, pinned_operator)
+
+    unique_deps_list = [f"{dep.package_name}{dep.operator}{dep.version}" for dep in unique_deps]
```
ant0nsc, line 1008: You could move that f"..." string into a __str__ method of the PackageDependency class

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -895,54 +896,125 @@ def _log_conda_dependencies_stats(conda: CondaDependencies, message_prefix: str)
...
+    """
+    parts: List[str] = re.split('(<=|==|=|>=|<|>)', dep_str)
+    if len(parts) == 1:
+        return (parts[0].strip(), "", "")
+    if len(parts) == 3:
+        return tuple(p.strip() for p in parts)
+    raise ValueError(f"Unable to split this package string: {dep_str}")
+
+
+class PackageDependency:
```
ant0nsc, line 914: add docu for this class?

hi-ml-azure/src/health_azure/utils.py
ant0nsc: and for the sake of paranoia, also guard line 926 similarly

ant0nsc: Wait, no. all of these checks are obsolete because split... returns a 3-tuple in all cases

ant0nsc (dismissed):
Looking good as is, but maybe some small-fry cleanup that would be nice

hi-ml-azure/src/health_azure/utils.py
mebristo: Good catch - I've removed this check

ant0nsc (approved):

PR #268, opened 2022-03-28
ENH: Add dataloader kwargs in tiles data modules, including num_workers

dccastro:
Enables passing keyword arguments to the dataloaders returned by a `TilesDataModule`. With this, we can now customise settings like `num_workers`, which massively accelerates experiments without caching (e.g. fine-tuning runs).

Note that `num_workers > 0` does not work if the encoder forward pass happens in the dataloader workers (i.e. how we currently do feature caching), as this is only allowed in the main process (or DDP processes).

### DIFF
--- a/hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILECrck.py
+++ b/hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILECrck.py
@@ -108,12 +108,15 @@ class DeepSMILECrck(BaseMIL):
 
     def get_data_module(self) -> TilesDataModule:
         image_key = TcgaCrck_TilesDataset.IMAGE_COLUMN
-        transform = Compose(
-            [
+        if self.is_finetune:
+            transform = LoadTilesBatchd(image_key, progress=True)
+            dataloader_kwargs = dict(num_workers=os.cpu_count(), pin_memory=True)
+        else:
+            transform = Compose([
                 LoadTilesBatchd(image_key, progress=True),
-                EncodeTilesBatchd(keys=image_key, encoder=self.encoder, chunk_size=self.encoding_chunk_size)
-            ]
-        )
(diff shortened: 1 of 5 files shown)


hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILECrck.py
```diff
@@ -108,12 +108,16 @@ def setup(self) -> None:
 
     def get_data_module(self) -> TilesDataModule:
         image_key = TcgaCrck_TilesDataset.IMAGE_COLUMN
-        transform = Compose(
-            [
+        if self.is_finetune:
+            transform = LoadTilesBatchd(image_key, progress=True)
+            dataloader_kwargs = dict(num_workers=os.cpu_count(), pin_memory=True)
```
ant0nsc, line 113: here and below: `num_workers==os.cpu_count()` is quite likely too much. To the best of my knowledge, this will turn out to be number of workers per GPU. So if you have 8 GPUs, you're overloading the GPU by a factor of 8. This will increase the chance of process starvation. `floor(cpu_count / gpu_count)`?

ant0nsc (changes requested):

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILECrck.py
dccastro: Happy to try this, but I was following the guidance I found online. For example, from the Lightning docs on DDP: https://pytorch-lightning.readthedocs.io/en/latest/guides/speed.html#num-workers.

dccastro: Updated now following your suggestion and some further advice I found online.

ant0nsc (approved):

PR #270, opened 2022-03-28
BUG: Fix spurious test errors

ant0nsc:
Test pipeline fails occasionally when testing download from Azure because files already exist.

### DIFF
--- a/hi-ml-azure/testazure/testazure/test_azure_util.py
+++ b/hi-ml-azure/testazure/testazure/test_azure_util.py
@@ -24,7 +24,6 @@ import param
 import pytest
 from _pytest.capture import CaptureFixture
 from _pytest.logging import LogCaptureFixture
-from azureml._vendor.azure_storage.blob import Blob
 from azureml.core import Experiment, Run, ScriptRunConfig, Workspace
 from azureml.core.authentication import ServicePrincipalAuthentication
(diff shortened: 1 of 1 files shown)


hi-ml-azure/testazure/testazure/test_azure_util.py
```diff
@@ -1399,9 +1398,21 @@ def test_get_run_source(dummy_recovery_id: str,
...
+    """Deletes all existing files in blob storage at the location that the test uses.
+
+    param datastore: The datastore from which the files should be deleted.
+    param prefix: The prefix string for the files that should be deleted.
+    """
+    container = datastore.container_name
+    existing_blobs = list(datastore.blob_service.list_blobs(prefix=prefix,
+                                                            container_name=container))
+    for existing_blob in existing_blobs:
+        datastore.blob_service.delete_blob(container_name=container, blob_name=existing_blob.name)
```
mebristo, line 1411: Should we add a check here to make sure that deletion was successful?

ant0nsc: Wouldn't an unsuccessful deletion simply raise an error? Not sure it's worth adding extra complexity here. Could be that the deletion is not immediately visible, for example.

mebristo (approved):

PR #275, opened 2022-03-29
ENH: Using default workspace datastore when datastore name is missing 

ant0nsc:
If no datastore is given, use the workspace default datastore.

### DIFF
--- a/hi-ml-azure/src/health_azure/datasets.py
+++ b/hi-ml-azure/src/health_azure/datasets.py
@@ -309,7 +309,7 @@ def find_workspace_for_local_datasets(aml_workspace: Optional[Workspace],
             workspace = get_workspace(aml_workspace, workspace_config_path)
             logging.info(f"Found workspace for datasets: {workspace.name}")
         except Exception as ex:
-            logging.info(f"Could not find workspace for datasets: {ex.message}. Exception: {ex}")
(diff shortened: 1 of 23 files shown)


vale-salvatelli (approved):

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -29,8 +29,9 @@ def get_datastore(workspace: Workspace, datastore_name: str) -> Datastore:
     if not datastore_name:
         if len(existing_stores) == 1:
```
mebristo, line 30: I think it's worth adding a comment here that it is possible to have default datastore set, hence why we perform this check first

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -123,19 +124,18 @@ def to_input_dataset_local(self, workspace: Optional[Workspace]) -> Tuple[Option
         returned.
 
         :param workspace: The AzureML workspace to read from.
-        :return: Pair of optional path to dataset and optional mountcontext.
+        :return: Pair of path to dataset and optional mountcontext.
```
mebristo, line 127: In the type signature the path to dataset seems to still be optional? Also nit: should be 'paths' plural

mebristo (approved):

vale-salvatelli (approved):

PR #276, opened 2022-03-29
ENH: DRAFT WSI dataloader and tiling on the fly

kenza-bouzid:
Added support for WSI dataloading by creating a new SlidesDataModule that uses SlidesDataset and tiles on the fly. 
Main changes include:
- New HistoDataModule mother class from which SlidesDataModule and TilesDataModule inherit. 
- Use of TileOnGridd transform from MONAI for on fly tiling.
- Unit test for SlidesDataModule.
- Create mock WSI data by stitching tiles from pathmnist dataset.
- Configs for launching deepmil pipeline on a subset of PANDA dataset.


hi-ml-histopathology/src/histopathology/configs/classification/custom_subset/dump_dataset_csv.py
```diff
@@ -0,0 +1,43 @@
...
+) -> None:
+
+    # get target wsi ids
+    with open(datalist_json, "r") as f:
+        slides = json.load(f)
+    wsi_ids = [slide["image"].split(".")[0] for slide in slides[subset]]
+
+    # select corresponding tiles from dataset
+    df = pd.read_csv(dataset_csv)
+    df.to_csv(dataset_csv, index=False)
```
vale-salvatelli, line 26: not sure I understand the need of this operation

hi-ml-histopathology/src/histopathology/configs/classification/custom_subset/panda/sub_train_slides.csv
```diff
@@ -0,0 +1,90 @@
...
+da1bf67eeb0d10c8021d69a0d6128f15,radboud,3,4+3
+da1de69a3097c5cecfc5a5c20f125340,radboud,0,negative
+da2319b5a8f33c13bcf2380c25b78e37,radboud,3,4+3
+da2502ee900c4fd5653d4433d05c8cde,radboud,2,3+4
+da4b2e029abb2741bcfef60fa7ab10b0,radboud,5,5+4
+da517beeeaf103a75e2a716d46643b49,radboud,5,5+4
+da5bbea85af39c137a264ee0bd56ef50,radboud,5,4+5
+da5e1279af2669d8b8abf40d450bd896,karolinska,0,0+0
+da5fc79f09ea997183453a16d8f3c393,karolinska,0,0+0
+da610fb22e9ba742ebb7a9117b2f24cc,radboud,2,3+4
```
vale-salvatelli, line 90: We selected 89 slides for training and 14 for validation. Just checking these numbers were what you wanted? I thought we were thinking much smaller numbers for benchmarking at the beginning, but I might just have lost part of the conversation

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -14,25 +15,32 @@
...
+    DISK = "disk"
 
 
 class CacheLocation(Enum):
-    NONE = 'none'
-    CPU = 'cpu'
-    SAME = 'same'
+    NONE = "none"
+    CPU = "cpu"
+    SAME = "same"
```
vale-salvatelli, line 39: Checking if your formatting setup match those of the repo? It shouldn't happen when you commit that single quote transform into double quote of the opposite. But here it might be an old file that has been untouched for a while

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -14,25 +15,32 @@
 
 from health_ml.utils.bag_utils import BagDataset, multibag_collate
 from health_ml.utils.common_utils import _create_generator
+from health_ml.utils.wsi_utils import list_data_collate
 
-from histopathology.datasets.base_dataset import TilesDataset
+from histopathology.datasets.base_dataset import SlidesDataset, TilesDataset
+from histopathology.datasets.panda_dataset import PandaDataset
```
vale-salvatelli, line 21: Why this is needed here? this is a base_module, I would not expect any dataset specific to happen here

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +201,139 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+        self.random_offset = random_offset
+        self.pad_full = pad_full
+        self.background_val = background_val
+        self.filter_mode = filter_mode
+
+    def prepare_data(self) -> None:
+        if self.precache_location != CacheLocation.NONE:
+            self._load_dataset(self.train_dataset, stage="train")
+            self._load_dataset(self.val_dataset, stage="val")
+            self._load_dataset(self.test_dataset, stage="test")
```
vale-salvatelli, line 274: the only difference I see with TilesDataModule is that here we don't shuffle.  We could make the shuffling a parameter and put this method in the base HistoDataModule (or we could shuffle also when we work with slides)

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +201,139 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+        if dataset_pickle_path and dataset_pickle_path.is_file():
+            if self.precache_location == CacheLocation.CPU:
+                memory_location = torch.device("cpu")
+                print(f"Loading dataset from {dataset_pickle_path} into {memory_location}")
+            else:
+                # by default torch.load will reload on the same device it was saved from
+                memory_location = None  # type: ignore
+
+            with dataset_pickle_path.open("rb") as f:
+                return torch.load(f, map_location=memory_location)
```
vale-salvatelli, line 288: Do we need these caching options for the slides pipeline?

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +201,139 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+                LoadImaged(
+                    keys=slides_dataset.IMAGE_COLUMN,
+                    reader=WSIReader,
+                    backend="cucim",
+                    dtype=np.uint8,
+                    level=self.level,
+                    image_only=True,
+                ),
+                TileOnGridd(
+                    keys=PandaDataset.IMAGE_COLUMN,
```
vale-salvatelli, line 301: this should be a generic class, we should remove any hard-coded dependency from specific datasets, we can pass this variable to the constructor if it's not available

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -14,25 +15,32 @@
...
+    SAME = "same"
 
 
-class TilesDataModule(LightningDataModule):
-    """Base class to load the tiles of a dataset as train, val, test sets"""
+class HistoDataModule(LightningDataModule):
+    """Base class to load a histopathology dataset as train, val, test sets"""
 
     def __init__(self, root_path: Path, max_bag_size: int = 0, batch_size: int = 1,
                  max_bag_size_inf: int = 0,
```
vale-salvatelli, line 46: `transforms` is in the docstring but not here. Also some of the docstring might need some generalization because now they refer to tiles

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
```diff
@@ -32,3 +34,63 @@ def get_splits(self) -> Tuple[PandaTilesDataset, PandaTilesDataset, PandaTilesDa
         return (PandaTilesDataset(self.root_path, dataset_df=splits.train),
                 PandaTilesDataset(self.root_path, dataset_df=splits.val),
                 PandaTilesDataset(self.root_path, dataset_df=splits.test))
+
+
+class SubPandaTilesDataModule(TilesDataModule):
+    """ subPandaTilesDataModule is the child class of PandaDataModule specific to PANDA dataset
```
vale-salvatelli, line 40:
```suggestion
    """ subPandaTilesDataModule is a child class of TilesDataModule specific to PANDA dataset. The difference with `PandaTilesDataModule` is that

File: hi-ml-histopathology/src/histopathology/datamodules/panda_module.py [hunk 10]
```diff
@@ -32,3 +34,63 @@ def get_splits(self) -> Tuple[PandaTilesDataset, PandaTilesDataset, PandaTilesDa
         return (PandaTilesDataset(self.root_path, dataset_df=splits.train),
                 PandaTilesDataset(self.root_path, dataset_df=splits.val),
                 PandaTilesDataset(self.root_path, dataset_df=splits.test))
+
+
+class SubPandaTilesDataModule(TilesDataModule):
```
vale-salvatelli, line 39: We might want to call this class differently because its peculiarity is more in how the splits are built (it could be a subset or the whole dataset) and the fact it's not compatible with cross-validation.

On the longer-term cross-validation is something we want to build. Maybe worth to reason a second if it's easier to build that compatibility after or we should do it from the start.

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
```diff
@@ -32,3 +34,63 @@ def get_splits(self) -> Tuple[PandaTilesDataset, PandaTilesDataset, PandaTilesDa
...
+    def __init__(self, train_csv: Union[str, Path], val_csv: Union[str, Path], **kwargs: Any) -> None:
+        self.train_csv = train_csv
+        self.val_csv = val_csv
+        super().__init__(**kwargs)
+
+    def get_splits(self) -> Tuple[PandaTilesDataset, PandaTilesDataset, PandaTilesDataset]:
+        return (
+            PandaTilesDataset(self.root_path, self.train_csv),
+            PandaTilesDataset(self.root_path, self.val_csv),
+            PandaTilesDataset(self.root_path, self.val_csv),
```
vale-salvatelli, line 54: we're returning two times the validation set, this is prone to errors down the line, should we have a test file? or return an empty dataset?

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
```diff
@@ -32,3 +34,63 @@ def get_splits(self) -> Tuple[PandaTilesDataset, PandaTilesDataset, PandaTilesDa
...
+                                                subject_column=dataset.TILE_ID_COLUMN,
+                                                group_column=dataset.SLIDE_ID_COLUMN)
+
+        if self.crossval_count > 1:
+            # Function get_k_fold_cross_validation_splits() will concatenate train and val splits
+            splits = splits.get_k_fold_cross_validation_splits(self.crossval_count)[self.crossval_index]
+
+        return (PandaDataset(self.root_path, dataset_df=splits.train),
+                PandaDataset(self.root_path, dataset_df=splits.val),
+                PandaDataset(self.root_path, dataset_df=splits.test))
```
vale-salvatelli, line 78: Same comment as above

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
```diff
@@ -32,3 +34,63 @@ def get_splits(self) -> Tuple[PandaTilesDataset, PandaTilesDataset, PandaTilesDa
...
+    def __init__(self, train_csv: Union[str, Path], val_csv: Union[str, Path], **kwargs: Any) -> None:
+        self.train_csv = train_csv
+        self.val_csv = val_csv
+        super().__init__(**kwargs)
+
+    def get_splits(self) -> Tuple[PandaDataset, PandaDataset, PandaDataset]:
+        return (
+            PandaDataset(self.root_path, self.train_csv),
+            PandaDataset(self.root_path, self.val_csv),
+            PandaDataset(self.root_path, self.val_csv),
```
vale-salvatelli, line 95: still same comment

hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -226,7 +226,7 @@ def _shared_step(self, batch: Dict, batch_idx: int, stage: str) -> BatchResultsT
 
         results = dict()
         for metric_object in self.get_metrics_dict(stage).values():
-            metric_object.update(predicted_probs, bag_labels.view(batch_size,))
+            metric_object.update(predicted_probs, bag_labels.view(batch_size,).int())
```
vale-salvatelli, line 229: why this change was needed?

hi-ml-histopathology/testhisto/test_data/whole_slide_images/dump_mock_dataframe.py
```diff
@@ -0,0 +1,36 @@
...
+def create_mock_dataframe(n_samples: int = 4) -> None:
+    data: dict = {"image_id": [], "data_provider": [], "isup_grade": [], "gleason_score": []}
+    for i in range(n_samples):
+        rand_id = np.random.randint(0, 11)
+        data["image_id"].append(f"_{i}")
+        data["data_provider"].append(np.random.choice(DATA_PROVIDERS, 1)[0])
+        data["isup_grade"].append(ISUP_GRADES[rand_id])
+        data["gleason_score"].append(GLEASON_SCORES[rand_id])
+    df = pd.DataFrame(data=data)
+    df.to_csv("pathmnist/dataset.csv", index=False)
```
vale-salvatelli, line 27: do we need to save this permanently in the disk? if not better to create the mock dataframe on the fly and fix a seed to be sure the test dataframe is always the same for each test

hi-ml-histopathology/testhisto/test_data/whole_slide_images/dump_mock_wsi.py
```diff
@@ -0,0 +1,146 @@
...
+import torchvision.transforms as transforms
+
+from torch import Tensor
+from medmnist import INFO
+from tifffile import TiffWriter
+from typing import Iterable, List
+from health_azure.utils import is_local_rank_zero
+from health_ml.utils.common_utils import logging_to_stdout
+
+
```
vale-salvatelli, line 21: docstring are always useful :)

hi-ml-histopathology/testhisto/test_data/whole_slide_images/dump_mock_wsi.py
```diff
@@ -0,0 +1,146 @@
...
+    n_channels: int = 3,
+    n_samples: int = 4,
+    n_series: int = 3,
+    different_tiles: bool = False,
+) -> None:
+
+    data_loader = get_pathmnist_data_loader(n_repeat)
+    for sample_counter in range(n_samples):
+        if different_tiles:
+            os.makedirs(f"pathmnist/_{sample_counter}", exist_ok=True)
```
vale-salvatelli, line 71: when this folder is used? in general to help the reader you might consider giving telling variable names to recurring folders, paths rather than creating the full path each time

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -147,24 +147,27 @@ def create_model(self) -> DeepMILModule:
...
+                                       class_weights=self.data_module.class_weights,
+                                       l_rate=self.l_rate,
+                                       weight_decay=self.weight_decay,
+                                       adam_betas=self.adam_betas,
+                                       is_finetune=self.is_finetune,
+                                       class_names=self.class_names,
+                                       outputs_handler=outputs_handler
+                                       )
+        # WARNING: We set slides_dataset out of the constructor so that it's not saved as a hyperparameter as
+        # slides_datasets are not dumpable by yaml.
```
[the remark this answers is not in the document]
kenza-bouzid: This has been moved to a separate PR.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -147,24 +147,27 @@ def create_model(self) -> DeepMILModule:
...
+        deepmil_module.outputs_handler.slides_dataset = self.get_slides_dataset()
```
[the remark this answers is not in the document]
kenza-bouzid: This has been moved to a separate PR.

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -181,3 +184,36 @@ def __init__(self, **kwargs: Any) -> None:
...
+            train_csv=self.train_csv,
+            val_csv=self.val_csv,
+            root_path=self.local_datasets[0],
+            max_bag_size=self.max_bag_size,
+            batch_size=self.batch_size,
+            transform=transform,
+            cache_mode=self.cache_mode,
+            precache_location=self.precache_location,
+            cache_dir=self.cache_dir,
+        )
```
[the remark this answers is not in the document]
kenza-bouzid: This was mainly to create the baseline for the PANDA subset but apparently this might be useful for testing as well. It might be better to move to a dedicated module config. As for the finetune option, I don't really know if we need it for the benchmark, I kept the exact same logic as in the base config and only change the subset datamodule.

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -181,3 +182,36 @@ def __init__(self, **kwargs: Any) -> None:
...
         super().__init__(encoder_type=HistoSSLEncoder.__name__, **kwargs)
+
+
+class SubPandaImageNetMIL(PandaImageNetMIL):
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.crossval_count = 1
+        self.train_csv = os.path.join(
+            fixed_paths.repository_root_directory(), "custom_subset/panda/sub_train_tiles.csv")
+        self.val_csv = os.path.join(fixed_paths.repository_root_directory(), "custom_subset/panda/sub_val_tiles.csv")
```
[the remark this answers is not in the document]
ant0nsc: I'm not sure this path really works? the file seems to be at histo/src/histo/configs/classication/custom_subset?

hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
```diff
@@ -0,0 +1,121 @@
...
+    for sample in dataloader:
+        tiles = sample["image"]
+        assert tiles.shape[1] >= min_expected_tile_count
+
+
+@pytest.mark.skipif(no_gpu, reason="Test requires GPU")
+@pytest.mark.gpu
+@pytest.mark.parametrize("level", [0, 1, 2])
+def test_multi_resolution_tiling(level: int) -> None:
+    batch_size, tile_count, level, channels = 1, 16, 0, 3
```
vale-salvatelli, line 85: level should not be overwritten here, it's passed as parameter

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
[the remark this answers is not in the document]
ant0nsc: Agree that moving to a separate module, maybe in the test folder, could help.

hi-ml/src/health_ml/utils/wsi_utils.py
```diff
@@ -0,0 +1,19 @@
...
+def list_data_collate(batch: List) -> Any:
+    """
+        Combine instances from a list of dicts into a single dict, by stacking them along first dim
+        [{'image' : 3xHxW}, {'image' : 3xHxW}, {'image' : 3xHxW}...] - > {'image' : Nx3xHxW}
+        followed by the default collate which will form a batch BxNx3xHxW
+    """
+
+    for i, item in enumerate(batch):
+        data = item[0]
+        data["image"] = torch.tensor(np.array([ix["image"] for ix in item]))
```
vale-salvatelli, line 17: `image` should be a parameter to make this a utils function, we might have datasets where the string is different

hi-ml-histopathology/src/histopathology/configs/classification/custom_subset/dump_dataset_csv.py
```diff
@@ -0,0 +1,43 @@
+#  ------------------------------------------------------------------------------------------
+#  Copyright (c) Microsoft Corporation. All rights reserved.
+#  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+#  ------------------------------------------------------------------------------------------
+import argparse
```
ant0nsc, line 5: This file and the custom subset files should also move to a different module, maybe in the test folder.

hi-ml-histopathology/src/histopathology/configs/classification/custom_subset/dump_dataset_csv.py
```diff
@@ -0,0 +1,43 @@
...
+        slides = json.load(f)
+    wsi_ids = [slide["image"].split(".")[0] for slide in slides[subset]]
+
+    # select corresponding tiles from dataset
+    df = pd.read_csv(dataset_csv)
+    df.to_csv(dataset_csv, index=False)
+    df.loc[df[image_id].isin(wsi_ids)].to_csv(dest_dataset_csv, index=False)
+
+
+if __name__ == "__main__":
```
ant0nsc, line 30: Can this subsetting script actually be run on-the-fly? At present, you need a slides list, run this script to create a tiles list. But the operations are so cheap that we could do that subsetting in the dataloader.

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +201,139 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
-        return DataLoader(transformed_bag_dataset, batch_size=self.batch_size,
-                          collate_fn=multibag_collate, shuffle=shuffle, generator=generator,
-                          **dataloader_kwargs)
+        return DataLoader(
+            transformed_bag_dataset,
+            batch_size=self.batch_size,
+            collate_fn=multibag_collate,
+            shuffle=shuffle,
+            generator=generator,
+            pin_memory=False,  # disable pinning as loaded data may already be on GPU
```
ant0nsc, line 221: Just read through this thread here https://stackoverflow.com/questions/55563376/pytorch-how-does-pin-memory-work-in-dataloader, after which I'd say to always pin memory, and the comment is also not needed

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +201,139 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+                    tile_size=self.tile_size,
+                    step=self.step,
+                    random_offset=self.random_offset,
+                    pad_full=self.pad_full,
+                    background_val=self.background_val,
+                    filter_mode=self.filter_mode,
+                    return_list_of_dicts=True,
+                ),
+            ]
+        )
```
vale-salvatelli, line 312: Just noting that I think we will need some design changes to integrate with deepmil (but we can tackle in the next PR):

- we will need to put together in a tensor the tiles and the labels like here https://github.com/Project-MONAI/tutorials/blob/622d460e0845cebff5f32022b2ef35448b61a1ee/pathology/multiple_instance_learning/panda_mil_train_evaluate_pytorch_gpu.py#L369 
- we will need the ability to pass different transformations for training and testing purposes

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +201,139 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+            Defaults to ``None``.
+        tile_size: size of the square tile
+            Defaults to ``224``.
+        step: step size to create overlapping tiles
+            Defaults to ``None`` (same as tile_size)
+        random_offset: Randomize position of the grid, instead of starting from the top-left corner
+            Defaults to ``False``.
+        pad_full: pad image to the size evenly divisible by tile_size
+            Defaults to ``False``.
+        background_val: the background constant (e.g. 255 for white background)
```
ant0nsc, line 241: Maybe say what the background setting does?

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +201,139 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+    ) -> DataLoader:
+        transformed_slides_dataset = self._load_dataset(dataset, stage)
+        generator = _create_generator(self.seed)
+        return DataLoader(
+            transformed_slides_dataset,
+            batch_size=self.batch_size,
+            collate_fn=list_data_collate,
+            shuffle=shuffle,
+            generator=generator,
+            pin_memory=False,  # disable pinning as loaded data may already be on GPU
```
ant0nsc, line 337: See my comment above about pin_memory

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
```diff
@@ -32,3 +34,63 @@ def get_splits(self) -> Tuple[PandaTilesDataset, PandaTilesDataset, PandaTilesDa
...
                 PandaTilesDataset(self.root_path, dataset_df=splits.test))
+
+
+class SubPandaTilesDataModule(TilesDataModule):
+    """ subPandaTilesDataModule is the child class of PandaDataModule specific to PANDA dataset
+    Method get_splits() returns the train, val, test splits from a subset of the PANDA dataset specified by
+    train/validation dataframes.
+    """
+
+    def __init__(self, train_csv: Union[str, Path], val_csv: Union[str, Path], **kwargs: Any) -> None:
```
ant0nsc, line 45: We have a type annotation PathOrString for that. But generally, I would try to do without str for paths, and use Path throughout.

hi-ml-histopathology/src/histopathology/configs/classification/custom_subset/dump_dataset_csv.py
kenza-bouzid: Here I extact wsi ids from the json file (as in MONAI pipeline) then look for the corresponding tiles in our dataset.csv

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
vale-salvatelli: I gave more thought on this. Personally, I would be in favor of removing this complexity from the `SlidesDataModule`. Unless the last transformation is an encoding stage that uses a fixed encoder (that we don't want to fine-tune further) we will not be able to cache the dataset, too large. 
In principle we could still give the user the possibility to use a MONAI CachedDataset. In this case what will be cached are the non-random transformations - however if we want the tiling to be random at each step - this means we would simply cache in memory loaded slides and the cache will be filled up before the end of a training loop, so I don't think in practice this operation would bring any advantage to training speed.
Any thoughts on this aspect @dccastro?

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +201,139 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+    def __init__(
+        self,
+        level: Optional[int] = 0,
+        tile_count: Optional[int] = None,
+        tile_size: Optional[int] = 224,
+        step: Optional[int] = None,
+        random_offset: Optional[bool] = False,
+        pad_full: Optional[bool] = False,
+        background_val: Optional[int] = 255,
+        filter_mode: Optional[str] = "min",
```
vale-salvatelli, line 257: I would default this to `random`, this way we will be sampling a different set of tiles at each step, otherwise we would need to sample many more tiles per slide to get a representative sample

hi-ml-histopathology/src/histopathology/configs/classification/custom_subset/panda/sub_train_slides.csv
kenza-bouzid: Yes, indeed! I just run into some issues and had to add more examples to check my assumptions. I will reduce that to the numbers we agreed upon.

hi-ml/src/health_ml/utils/wsi_utils.py
```diff
@@ -0,0 +1,19 @@
+from async_timeout import Any
+import torch
+import numpy as np
+from torch.utils.data.dataloader import default_collate
+from typing import List
+
+
+def list_data_collate(batch: List) -> Any:
+    """
```
ant0nsc, line 9: The name of the function is too generic. it is only doing something for images, so could be called image_collate?

hi-ml/src/health_ml/utils/wsi_utils.py
```diff
@@ -0,0 +1,19 @@
...
+        Combine instances from a list of dicts into a single dict, by stacking them along first dim
+        [{'image' : 3xHxW}, {'image' : 3xHxW}, {'image' : 3xHxW}...] - > {'image' : Nx3xHxW}
+        followed by the default collate which will form a batch BxNx3xHxW
+    """
+
+    for i, item in enumerate(batch):
+        data = item[0]
```
ant0nsc, line 16: I'm a little confused, what are the datatypes here? the list comprehension creates a list of numpy arrays or tensors? can that be simplified?

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
kenza-bouzid: I see your point for cross validation. I was just trying to match the subsets used in the MONAI pipeline. That's why I hard fixed the train/validation sets based on the csv files I create from the MONAI json file.

I should go through the MONAI code once again but I'm not sure they're doing cross-validation.

It definitely needs to be renamed, any suggestions?

ant0nsc (changes requested):
Overall, this is looking great! Left comments.
One important thing: Fixed stored files have historically always been troublemakers. Can you refactor to create a dataset of files on the fly, into a folder that is passed as an argument? This way, each test can be configured to use exactly the type of fake dataset that we need, and we can easily clean up afterwards.

PR #279, opened 2022-03-29
BUG: Params pl_limit_train(/val)_batches should accept float

mebristo:
Updates pl_limit_train_batches and pl_limit_val_batches to accept float types

### DIFF
--- a/hi-ml/testhiml/testhiml/test_model_trainer.py
+++ b/hi-ml/testhiml/testhiml/test_model_trainer.py
@@ -87,7 +87,8 @@ def test_create_lightning_trainer_with_callbacks() -> None:
     assert isinstance(storing_logger, StoringLogger)
 
 
-def test_create_lightning_trainer_limit_batches():
+@pytest.mark.fast
+def test_create_lightning_trainer_limit_batches() -> None:
     model_name = "HelloWorld"
     model_config_loader = ModelConfigLoader()
     container = model_config_loader.create_model_config_from_name(model_name)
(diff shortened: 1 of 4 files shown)


ant0nsc (dismissed):

hi-ml/testhiml/testhiml/test_model_trainer.py
```diff
@@ -87,6 +88,33 @@ def _get_trainer_arguments() -> Dict[str, Any]:
     assert isinstance(storing_logger, StoringLogger)
 
 
+@pytest.mark.fast
+def test_create_lightning_trainer_limit_batches() -> None:
```
dccastro, line 92: Would there be a way to test that the trainer has recognised the parameter we passed in, e.g. check the number of train/val batches we get? Or is this not necessary because we trust PL?

ant0nsc: I would not test at that level. As a default, I feel we should trust that PL does what it says on the box.

ant0nsc (dismissed):

ant0nsc (approved):

hi-ml/testhiml/testhiml/test_model_trainer.py
mebristo: I have added some very basic checks that make sure that the PL Trainer objects have the expected (limited) number of batches

dccastro (approved):

ant0nsc (approved):

PR #282, opened 2022-03-30
BUG: Workaround to avoid saving a dataframe as a hyperparameter

kenza-bouzid:
This is a bug fix for the error raised when trying to save slides_dataset as a hyperparameter by deepmil lightning module. The slides_dataset should be set out of the constructor.

### DIFF
--- a/hi-ml-histopathology/src/histopathology/utils/output_utils.py
+++ b/hi-ml-histopathology/src/histopathology/utils/output_utils.py
@@ -351,7 +351,7 @@ class DeepMILOutputsHandler:
     def test_outputs_dir(self) -> Path:
         return self.outputs_root / "test"
 
-    def set_slides_dataset(self, slides_dataset):
+    def set_slides_dataset(self, slides_dataset: SlidesDataset):
         self.slides_dataset = slides_dataset
 
     def _save_outputs(self, epoch_results: EpochResultsType, outputs_dir: Path) -> None:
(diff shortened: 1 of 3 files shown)


vale-salvatelli (dismissed):

hi-ml-histopathology/src/histopathology/utils/output_utils.py
```diff
@@ -315,14 +315,14 @@ class DeepMILOutputsHandler:
...
-                 primary_val_metric: MetricsKey, maximise: bool) -> None:
+                 class_names: Optional[Sequence[str]], primary_val_metric: MetricsKey,
+                 maximise: bool, slides_dataset: Optional[SlidesDataset] = None) -> None:
         """
         :param outputs_root: Root directory where to save all produced outputs.
         :param n_classes: Number of MIL classes (set `n_classes=1` for binary).
         :param tile_size: The size of each tile.
         :param level: The downsampling level (e.g. 0, 1, 2) of the tiles if available (default=1).
-        :param slides_dataset: Optional slides dataset from which to plot thumbnails and heatmaps.
+        :param slides_dataset: Optional slides dataset from which to plot thumbnails and heatmaps (default=None).
```
ant0nsc, line 325: I would actually remove that completely from the constructor arguments, it's such an easy trap to fall into.
You could also provide a "set_slides_dataset" method on this class that sets the datasets, rather than relying on people knowing that there is a magic field on that class.

ant0nsc (changes requested):
Code is OK as is, but you could improve slightly to remove that trap that the constructor argument still is, see comments.

ant0nsc (approved):

vale-salvatelli (approved):

PR #283, opened 2022-03-30
ENH: Enable overriding ExperimentConfig parameters from container

dccastro:
In a similar spirit to https://github.com/microsoft/InnerEye-DeepLearning/pull/589, this PR enables overriding `ExperimentConfig` parameters (e.g. `mount_in_azureml`, `cluster`, `num_nodes`, etc.) from a `LightningContainer`. The `update_experiment_config()` method already existed, but was currently not being called by the hi-ml runner.

### DIFF
--- a/hi-ml/src/health_ml/runner.py
+++ b/hi-ml/src/health_ml/runner.py
@@ -147,6 +147,9 @@ class Runner:
         apply_overrides(container, parser2_result.overrides)  # type: ignore
         container.validate()
 
+        # Allow overriding ExperimentConfig params from within the container
+        container.update_experiment_config(experiment_config)
+
         self.lightning_container = container
 
         return parser2_result
--- a/hi-ml/testhiml/testhiml/test_runner.py
+++ b/hi-ml/testhiml/testhiml/test_runner.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import List, Optional
 from unittest.mock import patch, MagicMock
 
+import param
 import pytest
 from _pytest.capture import SysCapture
 from azureml.train.hyperdrive import HyperDriveConfig
@@ -14,6 +15,7 @@ from azureml.train.hyperdrive import HyperDriveConfig
 from health_azure import AzureRunInfo, DatasetConfig
 from health_ml.configs.hello_world import HelloWorld  # type: ignore
 from health_ml.deep_learning_config import WorkflowParams
+from health_ml.experiment_config import ExperimentConfig
 from health_ml.lightning_container import LightningContainer
 from health_ml.runner import Runner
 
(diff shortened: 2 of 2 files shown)


hi-ml/testhiml/testhiml/test_runner.py
```diff
@@ -66,6 +68,70 @@ def test_parse_and_load_model(mock_runner: Runner, model_name: Optional[str], cl
...
+    experiment_config = mock_runner.experiment_config
+    assert experiment_config is not None
+    container = mock_runner.lightning_container
+    assert container is not None
+    assert isinstance(container, DummyContainerWithExperimentConfigOverrides)
+
+    # Current ExperimentConfig parameter priority is as follows:
+    # 1. Container
+    # 2. CLI
+    # 3. ExperimentConfig defaults
```
ant0nsc, line 113: I think it would make sense to have CLI as the highest priority level. Anytime someone goes through the pain of typing out something on the commandline, they probably really mean it ;-) You could achieve that by calling the `update_experiment_config` in runner line 139.

ant0nsc (changes requested):

hi-ml/testhiml/testhiml/test_runner.py
dccastro: The trouble with that is that the container may use some of its parameters to override `ExperimentConfig`. For example, what should be the expected behaviour in this case?
```python
# python runner.py --model MyContainer --batch_size 42

class MyContainer(LightningContainer):
  def update_experiment_config(experiment_config):
    experiment_config.tag = f"cool experiment with batch_size={self.batch_size}"
```

dccastro: If we choose to call `update_experiment_config()` before applying the CLI overrides to the container params, I'd at least make it a `@staticmethod` to clarify that you should only use literals or read-only values. However, I think this option is very limiting (cf. example above).

In reality, we only need the model name first to instantiate the container. How about the following:
1. Parse the runner arguments (`parser1`).
2. Instantiate container based on `--model`.
3. Override container params with remaining CLI args (`parser2`).
4. Call a `container.get_experiment_config()` method that instantiates an `ExperimentConfig` with whatever user-defined defaults.
5. Override experiment config with runner arguments (`parser1`).

hi-ml/src/health_ml/runner.py
```diff
@@ -147,6 +147,9 @@ def parse_and_load_model(self) -> ParserResult:
         apply_overrides(container, parser2_result.overrides)  # type: ignore
         container.validate()
 
+        # Allow overriding ExperimentConfig params from within the container
+        container.update_experiment_config(experiment_config)
```
vale-salvatelli, line 151: I think command line should overwrite what is the config file, therefore this should happen before `apply_overrides`

hi-ml/testhiml/testhiml/test_runner.py
vale-salvatelli: @dccastro not sure I follow the solution above, happy to follow up at voice if it helps. Overall I agree with @ant0nsc. CLI must have priority. And I think the logic with which we update arguments must be straightforward and easy to read otherwise it will become a nightmare. How often do you think we will be in the situation you describe? Is it something we already require in some of our configs?

dccastro:
Closing as low-priority and out-of-date.

PR #285, opened 2022-04-01
ENH: Add slide dataloader

kenza-bouzid:
Added support for WSI dataloading by creating a new SlidesDataModule that uses SlidesDataset and tiles on the fly.
Main changes include:

- New HistoDataModule mother class from which SlidesDataModule and TilesDataModule inherit.
- Use of TileOnGridd transform from MONAI for on fly tiling.
- Unit test for SlidesDataModule.
- Create mock WSI data by stitching tiles from pathmnist dataset.
- Generate mock data on the fly.


hi-ml-histopathology/src/histopathology/utils/wsi_utils.py
```diff
@@ -0,0 +1,20 @@
+import torch
+import numpy as np
+
+from typing import List
+from histopathology.utils.naming import SlideKey
+from torch.utils.data.dataloader import default_collate
+
+
+def list_data_collate(batch: List[dict]) -> List[dict]:
```
ant0nsc, line 9: give that a more telling name, like "image_collate"?

hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
```diff
@@ -0,0 +1,107 @@
...
+from testhisto.utils.utils_slides_datamodule import MockSlidesDataModule, MockWSIGenerator, MockWSIType
+
+
+no_gpu = not is_gpu_available()
+N_REPEAT_DIAG = 4
+
+
+@pytest.fixture(scope="session")
+def mock_wsi_root_dir(tmpdir_factory: pytest.TempdirFactory) -> py.path.local:
+    tmp_root_dir = tmpdir_factory.mktemp("mock_wsi")
```
ant0nsc, line 19: pytest has a fixture called "tmp_path" that you can use here, I think that's more easy than the tmpdirfactory

hi-ml-histopathology/testhisto/testhisto/utils/utils_slides_datamodule.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+class MockWSIGenerator:
+    """Generator class to create mock WSI on the fly. A mock WSI resembles to:
+                                [**      ]
+                                [  **    ]
+                                [    **  ]
+                                [      **]
+        where * reprents 2 tiles stitched along the Y axis.
+    :param METADATA_POSSIBLE_VALUES: Possible values to be assigned to the dataset metadata.
```
ant0nsc, line 74: For documentation to render correctly in sphinx, there needs to be a line between the description and the start of the :param section

ant0nsc (changes requested):
Looking great, good work! Just some minor comments

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -103,43 +101,80 @@ def get_splits(self) -> Tuple[TilesDataset, TilesDataset, TilesDataset]:
...
+
+    def _get_transformed_dataset(
+        self, dataset: Dataset, transform: Union[Sequence[Callable], Callable]
+    ) -> Dataset:
+        if self.cache_mode is CacheMode.MEMORY:
+            dataset = CacheDataset(dataset, transform, num_workers=1)  # type: ignore
+        elif self.cache_mode is CacheMode.DISK:
+            dataset = PersistentDataset(dataset, transform, cache_dir=self.cache_dir)  # type: ignore
+            if self.precache_location != CacheLocation.NONE:
+                import tqdm  # TODO: Make optional
```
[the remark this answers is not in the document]
kenza-bouzid: It was taken from existing code!

hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
kenza-bouzid: I used the tmpdirfactory because it is a session-scoped fixture so that we generate the mock data once for the entire test suite.
Here's the [doc](https://docs.pytest.org/en/6.2.x/tmpdir.html). Or do you prefer the tmp_path fixture?

ant0nsc: ah, yes, that makes sense. Leave as is.

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -103,43 +101,80 @@ def get_splits(self) -> Tuple[TilesDataset, TilesDataset, TilesDataset]:
...
-            self._load_dataset(self.train_dataset, stage='train', shuffle=True)
-            self._load_dataset(self.val_dataset, stage='val', shuffle=True)
-            self._load_dataset(self.test_dataset, stage='test', shuffle=True)
-
     def _dataset_pickle_path(self, stage: str) -> Optional[Path]:
         if self.cache_dir is None or self.cache_mode == CacheMode.NONE:
             return None
         return self.cache_dir / f"{stage}_dataset.pt"
 
+    def prepare_data(self) -> None:
```
vale-salvatelli, line 109: Better to raise an error saying this method needs to be implemented in the child class. It can help also to explain what th method should do given it returns None

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -45,13 +51,7 @@ def __init__(self, root_path: Path, max_bag_size: int = 0, batch_size: int = 1,
...
-        :param max_bag_size: Upper bound on number of tiles in each loaded bag during training stage. If 0 (default),
-        will return all samples in each bag. If > 0 , bags larger than `max_bag_size` will yield
-        random subsets of instances.
         :param batch_size: Number of slides to load per batch.
-        :param max_bag_size_inf: Upper bound on number of tiles in each loaded bag during validation and test stages.
-        If 0 (default), will return all samples in each bag. If > 0 , bags larger than `max_bag_size_inf` will yield
-        random subsets of instances.
         :param seed: pseudorandom number generator seed to use for shuffling instances and bags. Note that randomness in
         train/val/test splits is handled independently in `get_splits()`. (default: `None`)
         :param transform: A transform to apply to the source tiles dataset, or a composition of
```
vale-salvatelli, line 57: outdated comment, `LoadTilesBatchd` is the default only in TilesDataModule

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
kenza-bouzid: I think this is only needed in the TilesDataModule for caching, the SlidesDataModule does not implement it, I will remove it from the mother class

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +185,126 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+                    step=self.step,
+                    random_offset=self.random_offset,
+                    pad_full=self.pad_full,
+                    background_val=self.background_val,
+                    filter_mode=self.filter_mode,
+                    return_list_of_dicts=True,
+                ),
+            ]
+        )
+        transformed_slides_dataset = Dataset(slides_dataset, self.transform or base_transform)
```
vale-salvatelli, line 289: I was wondering if in this case is more useful to say `Compose(base_transform, self.transform) if self.transform else base_transform`)? I think most of the time we will want to add transformations after the tiling and in the current format this would imply redefining somewhere all the parameters for loading and tiling.

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +185,126 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+                    tile_count=self.tile_count,
+                    tile_size=self.tile_size,
+                    step=self.step,
+                    random_offset=self.random_offset,
+                    pad_full=self.pad_full,
+                    background_val=self.background_val,
+                    filter_mode=self.filter_mode,
+                    return_list_of_dicts=True,
+                ),
+            ]
```
vale-salvatelli, line 298: By default shouldn't we also end with a `ToTensorD` as in https://github.com/Project-MONAI/tutorials/blob/a810488d96b878f830473fb5b6498b035f010617/pathology/multiple_instance_learning/panda_mil_train_evaluate_pytorch_gpu.py#L369? What is the current shape of the dataloder output? this is also something we can try to address in the next PR when we integrate with deepmil, just a warning that this output will need to match what we have from the TilesDataModule

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -103,43 +101,80 @@ def get_splits(self) -> Tuple[TilesDataset, TilesDataset, TilesDataset]:
...
-
     def _dataset_pickle_path(self, stage: str) -> Optional[Path]:
         if self.cache_dir is None or self.cache_mode == CacheMode.NONE:
             return None
         return self.cache_dir / f"{stage}_dataset.pt"
 
+    def prepare_data(self) -> None:
+        pass
+
+    def _get_transformed_dataset(
```
vale-salvatelli: this method, that is the only one making use of the caching parameters, is currently called only in the TilesDataModule. I see two options:
* we support caching only for the TilesDataModule and we move this method inside that child class, with also all the respective parameters.
* we decide to keep the caching option also in the SlidesDataModule, we keep this method and caching parameters in the base_module and call _get_transformed_dataset inside the `load_dataset` implementation of `SlidesDataModule`.

For the sake of keeping this first slide implementation as simple as possible I would vote for the second

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
```diff
@@ -32,3 +32,26 @@ def get_splits(self) -> Tuple[PandaTilesDataset, PandaTilesDataset, PandaTilesDa
         return (PandaTilesDataset(self.root_path, dataset_df=splits.train),
                 PandaTilesDataset(self.root_path, dataset_df=splits.val),
                 PandaTilesDataset(self.root_path, dataset_df=splits.test))
+
+
+class PandaSlidesDataModule(SlidesDataModule):
+    """ PandaSlidesDataModule is the child class of SlidesDataModule specific to PANDA dataset
+    Method get_splits() returns the train, val, test splits from the PANDA dataset
```
vale-salvatelli, line 39: Just to highlight that this split could bring differences with the MONAI pipeline - PANDA has some presplit in train and validation. Here we're ignoring those splits and, I understand, splitting the training dataset only in train/val/test

hi-ml-histopathology/src/histopathology/utils/wsi_utils.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+        Combine instances from a list of dicts into a single dict, by stacking them along first dim
+        [{'image' : 3xHxW}, {'image' : 3xHxW}, {'image' : 3xHxW}...] - > {'image' : Nx3xHxW}
+        followed by the default collate which will form a batch BxNx3xHxW
+    """
+
+    for i, item in enumerate(batch):
+        data = item[0]
+        data[SlideKey.IMAGE] = torch.tensor(np.array([ix[SlideKey.IMAGE] for ix in item]))
```
vale-salvatelli, line 18: I think I had left this comment also in the previous draft PR, maybe among those you haven't seen: if this needs to be a generic function you shouldn't hardcode SlideKey.IMAGE here, you can pass the key as parameter of the function

kenza-bouzid: Yes, I left a comment on the previous PR as well, this function is used for collate_fn of torch Dataloader. If we add extra parameters, we should use partial(SlideKey.IMAGE) to specify the key, would that be ok?

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +185,126 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+            collate_fn=list_data_collate,
+            shuffle=shuffle,
+            generator=generator,
+            pin_memory=True,
+            **dataloader_kwargs,
+        )
 
     def train_dataloader(self) -> DataLoader:
-        return self._get_dataloader(self.train_dataset, 'train', shuffle=True, **self.dataloader_kwargs)
+        return self._get_dataloader(self.train_dataset, shuffle=True, **self.dataloader_kwargs)
```
vale-salvatelli, line 317: I think long term we need to pass here the information about the stage, it's needed to determined which set of transforms to apply

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +185,126 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+                ),
+            ]
+        )
+        transformed_slides_dataset = Dataset(slides_dataset, self.transform or base_transform)
+        return transformed_slides_dataset
+
+    def _get_dataloader(
+        self, dataset: SlidesDataset, shuffle: bool, **dataloader_kwargs: Any
+    ) -> DataLoader:
+        transformed_slides_dataset = self._load_dataset(dataset)
```
vale-salvatelli, line 304: following up on the discussion we had during the stand-up: we need to be able to pass different transforms to train and val/test (we don't want augmentation at testing time), here could be a good place where to pass the info about the stage to _load_dataset

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
kenza-bouzid: Yes, I noticed that as well and that's why I have created this SubPandaDataModule in the previous PR. Do you think we should go for a fix presplit? In which case we won't be able to cross validate...

hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
```diff
@@ -0,0 +1,107 @@
...
+@pytest.mark.gpu
+def test_tiling_on_the_fly(mock_wsi_root_dir: py.path.local) -> None:
+    batch_size, tile_count, tile_size, level, channels = 1, 16, 28, 0, 3
+    datamodule = MockSlidesDataModule(
+        root_path=mock_wsi_root_dir, batch_size=batch_size, tile_count=tile_count, tile_size=tile_size, level=level
+    )
+    dataloader = datamodule.train_dataloader()
+    for sample in dataloader:
+        # sanity check for expected shape
+        tiles, wsi_id = sample["image"], sample["slide_id"][0]
```
vale-salvatelli, line 50: I always suggest avoiding hard-coded string - they make maintenance a nightmare,  you can use `SildesDataset.IMAGE_COLUMN` and ``SildesDataset.SLIDE_ID_COLUMN``

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
kenza-bouzid: Oh, I actually just moved all the caching options to TilesDatModule only. As you said last time, caching would be detrimental if we want to have random tiling.

hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
vale-salvatelli: It applies also the hard-coded strings in the rest of this module

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
kenza-bouzid: The current output shape is (batch_size, tile_count, channels, img_size, img_size) Not sure if it matches the TilesDataModule but we can maybe address that in the next PR.

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
kenza-bouzid: definitely,  good point!

hi-ml-histopathology/testhisto/testhisto/utils/utils_slides_datamodule.py
```diff
@@ -0,0 +1,248 @@
...
+class MockSlidesDataset(SlidesDataset):
+    """Mock and child class of SlidesDataset, to be used for testing purposes.
+    It overrides the following, according to the PANDA cohort settings:
+
+    :param SLIDE_ID_COLUMN: CSV column name for slide ID set to "image_id".
+    :param LABEL_COLUMN: CSV column name for tile label set to "isup_grade".
+    :param METADATA_COLUMNS: Column names for all the metadata available on the CSV dataset file.
+    """
+
+    SLIDE_ID_COLUMN = "image_id"
```
vale-salvatelli, line 34: not entirely sure why we want to overwrite `SlidesDataset.SLIDE_ID_COLUMN`

hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
```diff
@@ -0,0 +1,107 @@
...
+        assert tiles.shape == (batch_size, tile_count, channels, tile_size, tile_size)
+
+        # check tiling on the fly at different resolutions
+        original_tile = np.load(str(mock_wsi_root_dir.join(f"{wsi_id}_tile.npy")))[0]
+        for i in range(tile_count):
+            assert (original_tile[:, :: 2 ** level, :: 2 ** level] == tiles[0, i].numpy()).all()
```
vale-salvatelli, line 96: A comment here referring to the fact that multi resolution versions of the mock image has been created via 2 factor downsampling might help the reader

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
vale-salvatelli: Sounds good to me. This method and cache parameters should then not be in the base class

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
vale-salvatelli: I think it's fine for now. Just soemthing to keep in mind when we will want to compare results

vale-salvatelli:
Great work @kenza-bouzid! still a few suggestions but I think ready to merge after those

hi-ml-histopathology/src/histopathology/utils/wsi_utils.py
vale-salvatelli: sorry I hadn't seen your reply there! It's OK we can leave it as custom function instead of generic one. Maye change the docstring to make that clear `Combine images from a list of dicts into a single dict`?

hi-ml-histopathology/testhisto/testhisto/utils/utils_slides_datamodule.py
kenza-bouzid: I overwrote it as in PandaDataset but it might be unnecessary!

vale-salvatelli (approved):

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -150,40 +193,130 @@ def _load_dataset(self, tiles_dataset: TilesDataset, stage: str, shuffle: bool)
...
+                random_offset=self.random_offset,
+                pad_full=self.pad_full,
+                background_val=self.background_val,
+                filter_mode=self.filter_mode,
+                return_list_of_dicts=True,
+            ),
+        ]
+
+        transforms = [*base_transform, self.transform] if self.transform else base_transform
+        transformed_slides_dataset = Dataset(slides_dataset, Compose(transforms))
```
vale-salvatelli, line 299: I would add a test to see if this works

ant0nsc (approved):

ant0nsc (approved):

vale-salvatelli (approved):

PR #286, opened 2022-04-01
ENH: Add validation results to crossval report

dccastro:
* Adds utilities to check whether cross-validation runs are pre- or post-addition of validation outputs.
* Adds ROC and PR curves for validation data, if available.
* Enables selecting which score column to use for generating ROC/PR curves.
* Updates default argument name for `crossval_count/index`.

### DIFF
--- a/hi-ml-histopathology/src/histopathology/utils/report_utils.py
+++ b/hi-ml-histopathology/src/histopathology/utils/report_utils.py
@@ -13,7 +13,7 @@ from azureml.core import Experiment, Run, Workspace
 
 from health_azure.utils import (aggregate_hyperdrive_metrics, download_file_if_necessary, get_aml_run_from_run_id,
                                 get_tags_from_hyperdrive_run)
-from histopathology.utils.output_utils import AML_LEGACY_TEST_OUTPUTS_CSV, AML_OUTPUTS_DIR, AML_TEST_OUTPUTS_CSV, AML_VAL_OUTPUTS_CSV
+from histopathology.utils.output_utils import AML_LEGACY_TEST_OUTPUTS_CSV, AML_TEST_OUTPUTS_CSV, AML_VAL_OUTPUTS_CSV
 
 
 def run_has_val_and_test_outputs(run: Run) -> bool:
@@ -81,8 +81,7 @@ def collect_crossval_outputs(parent_run_id: str, download_dir: Path, aml_workspa
             raise ValueError(f"Child run expected to have the tag '{crossval_arg_name}'")
         child_dir = download_dir / str(child_run_index)
         try:
-            remote_filename = AML_OUTPUTS_DIR + "/" + output_filename
-            child_csv = download_file_if_necessary(child_run, remote_filename, child_dir / output_filename,
+            child_csv = download_file_if_necessary(child_run, output_filename, child_dir / output_filename,
                                                    overwrite=overwrite)
             all_outputs_dfs[child_run_index] = pd.read_csv(child_csv)
         except Exception as e:


fepegar:
Before I review this, can you please take a look at the failing tests?

hi-ml-histopathology/src/histopathology/utils/report_utils.py
```diff
@@ -15,8 +15,51 @@
...
+    :return: `True` if the run has validation and test outputs, `False` if it is a legacy run with
+        only test outputs.
+    """
+    outputs_filename = "test_output.csv"
+    val_outputs_filename = "val/" + outputs_filename
+    test_outputs_filename = "test/" + outputs_filename
+
+    outputs_folder = "outputs/"
+    all_available_files: List[str] = run.get_file_names()
+    output_files = [filename[len(outputs_folder):] for filename in all_available_files
```
mebristo, line 32: Is it simpler to set `val_outputs_filename = "outputs/val/" + outputs_filename` and then let this condition be `if val_ouputs_filename in all_available_files`? (and similarly for test_outputs_filename)

hi-ml-histopathology/src/histopathology/utils/report_utils.py
```diff
@@ -15,8 +15,51 @@
...
+                    if filename.startswith(outputs_folder)]
+    if val_outputs_filename in output_files and test_outputs_filename in output_files:
+        return True
+    elif outputs_filename in output_files:
+        return False
+    else:
+        raise ValueError(f"Run {run.display_name} ({run.id}) does not have the expected files: {output_files}")
```
mebristo, line 39: I think this error message may be a bit confusing since it doesn't tell the user the expected file it is looking for (`outputs_filename` )

hi-ml-histopathology/src/histopathology/utils/report_utils.py
```diff
@@ -15,8 +15,51 @@
...
+    :return: `True` if all children have validation and test outputs, `False` if all children are
+        legacy runs with only test outputs.
+    """
+    have_val_and_test_outputs = [run_has_val_and_test_outputs(child_run) for child_run in parent_run.get_children()]
+    if all(have_val_and_test_outputs):
+        return True
+    elif not any(have_val_and_test_outputs):
+        return False
+    else:
+        raise ValueError(f"Parent run {parent_run.display_name} ({parent_run.id}) has children with and without "
```
mebristo, line 57: Would the user know what is meant by 'with and without val and test outputs, or should we explicitly say which 2 files we are looking for here?

hi-ml-histopathology/testhisto/testhisto/utils/test_report_utils.py
```diff
@@ -10,7 +10,76 @@
...
+
+def test_run_has_val_and_test_outputs() -> None:
+    outputs_folder = "outputs/"
+    outputs_filename = "test_output.csv"
+    legacy_outputs_filename = outputs_folder + outputs_filename
+    val_outputs_filename = outputs_folder + "val/" + outputs_filename
+    test_outputs_filename = outputs_folder + "test/" + outputs_filename
+    arbitrary_filename = outputs_filename + "some_other_file"
+
+    run = MagicMock(display_name="mock run", id="abc123")
```
mebristo, line 26: For extra certainty, you could try creating a real Run objects with `create_aml_run_object` and uploading (or not) the expected files, then calling `run_has_val_and_test_outputs` on those

hi-ml-histopathology/src/histopathology/utils/analysis_plot_utils.py
```diff
@@ -171,8 +172,8 @@ def _plot_crossval_roc_and_pr_curves(crossval_dfs: Dict[int, pd.DataFrame], roc_
     """
     for k, tiles_df in crossval_dfs.items():
         slides_groupby = tiles_df.groupby(ResultsKey.SLIDE_ID)
-        labels = slides_groupby[ResultsKey.TRUE_LABEL].agg(pd.Series.mode)
-        scores = slides_groupby[ResultsKey.PROB].agg(pd.Series.mode)
+        labels = slides_groupby[ResultsKey.TRUE_LABEL].first()
```
mebristo, line 175: Do we expect all the tiles in the group to have the same label & score? If not, should we explain in the docstring the decision to take the first in the group, rather than the most frequent?

hi-ml-histopathology/src/histopathology/scripts/generate_crossval_report.py
```diff
@@ -61,14 +61,37 @@ def generate_html_report(parent_run_id: str, output_dir: Path, workspace_config_
...
-        roc_pr_curves_fig_path = report_dir / "roc_pr_curves.png"
-        fig.savefig(roc_pr_curves_fig_path, bbox_inches='tight')
-        report.add_images([roc_pr_curves_fig_path], base64_encode=True)
+        report.add_heading("Validation ROC and PR curves", level=3)
+        fig = plot_crossval_roc_and_pr_curves(val_outputs_dfs, scores_column='prob_class1')
+        val_roc_pr_curves_fig_path = report_dir / "val_roc_pr_curves.png"
+        fig.savefig(val_roc_pr_curves_fig_path, bbox_inches='tight')
+        report.add_images([val_roc_pr_curves_fig_path], base64_encode=True)
+
+    if include_test:
```
mebristo, line 81: We already check `if include_test` on line 58 - can this block move inside there?

dccastro:
> Before I review this, can you please take a look at the failing tests?

@fepegar Seems to have been a spurious HTTP error in an unrelated test. All green on re-run. 🤷

hi-ml-histopathology/src/histopathology/utils/report_utils.py
```diff
@@ -15,8 +15,51 @@
...
+    :raises ValueError: If the run does not have the expected output file(s).
+    :return: `True` if the run has validation and test outputs, `False` if it is a legacy run with
+        only test outputs.
+    """
+    outputs_filename = "test_output.csv"
+    val_outputs_filename = "val/" + outputs_filename
+    test_outputs_filename = "test/" + outputs_filename
+
+    outputs_folder = "outputs/"
+    all_available_files: List[str] = run.get_file_names()
```
fepegar, line 31: It would be nice to use `pathlib.Path` in this function, instead of strings. Then, once you've defined the output, test and val folder, you can just `return test_outputs_path.is_file() and val_outputs_path.is_file()`.

hi-ml-histopathology/src/histopathology/utils/report_utils.py
```diff
@@ -15,8 +15,51 @@
                                 get_tags_from_hyperdrive_run)
 
 
+def run_has_val_and_test_outputs(run: Run) -> bool:
```
fepegar, line 18: I think maybe the purpose of this function should be limited to checking whether the run has val and test outputs. Checking whether a test folder is randomly missing feels like an extra responsibility that could be just omitted.

hi-ml-histopathology/src/histopathology/utils/report_utils.py
fepegar: If this error is going to be printed out, we should probably also say which child runs are incomplete.

hi-ml-histopathology/testhisto/testhisto/utils/test_report_utils.py
```diff
@@ -10,7 +10,76 @@
...
+    with patch.object(run, 'get_file_names', return_value=[test_outputs_filename]):
+        with pytest.raises(ValueError, match="does not have the expected files"):
+            run_has_val_and_test_outputs(run)
+
+    with patch.object(run, 'get_file_names', return_value=[arbitrary_filename]):
+        with pytest.raises(ValueError, match="does not have the expected files"):
+            run_has_val_and_test_outputs(run)
+
+
+def test_crossval_runs_have_val_and_test_outputs() -> None:
```
fepegar, line 47: [Same comments about hard-coded variables and loops for tests]

fepegar (changes requested):

hi-ml-histopathology/src/histopathology/utils/analysis_plot_utils.py
dccastro: Yes, the CSV is constructed such that all tiles from the same slide have the same label and score. Mode was an arbitrary way to get this unique value, but I think it was somehow was failing in my latest environment.

hi-ml-histopathology/src/histopathology/utils/report_utils.py
dccastro: Normally I would indeed use `Path`s, but these are files in an AML run; they don't exist on the local machine.

hi-ml-histopathology/src/histopathology/utils/analysis_plot_utils.py
fepegar: Just to make sure and make the code slightly more explicit, shall we add something something like:

```python
label_series = slides_groupby[ResultsKey.TRUE_LABEL]
assert len(label_series.unique()) == 1
labels = label_series.first()
```

hi-ml-histopathology/src/histopathology/utils/report_utils.py
fepegar: Of course 🤦‍♂️.

hi-ml-histopathology/src/histopathology/utils/report_utils.py
dccastro: Clarified the message a bit more. I was also trying to be a bit cautious about over-committing on the filenames, as they may still be subject to change.

hi-ml-histopathology/src/histopathology/scripts/generate_crossval_report.py
dccastro: As @fepegar said, the report's layout is currently:
* General run information
* AML metrics
  * Training curves
  * Validation metrics table
  * [optional] Test metrics table
* ROC/PR plots
  * Validation curves
  * [optional] Test curves

(Report page is fairly tall, so screenshot(s) would be awkward 😄)

hi-ml-histopathology/src/histopathology/scripts/generate_crossval_report.py
```diff
@@ -61,14 +61,37 @@ def generate_html_report(parent_run_id: str, output_dir: Path, workspace_config_
...
+        test_outputs_filename = "test_output.csv"
+        if has_val_and_test_outputs:
+            test_outputs_filename = "test/" + test_outputs_filename
+        test_outputs_dfs = collect_crossval_outputs(parent_run_id, report_dir, aml_workspace,
+                                                    output_filename=test_outputs_filename,
+                                                    overwrite=overwrite)
+
+        report.add_heading("Test ROC and PR curves", level=3)
+        fig = plot_crossval_roc_and_pr_curves(test_outputs_dfs, scores_column='prob_class1')
+        test_roc_pr_curves_fig_path = report_dir / "test_roc_pr_curves.png"
```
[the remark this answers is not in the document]
dccastro: For `test_outputs.csv` I agree, but these image files are only ever used right here to be immediately rendered onto the HTML report.

hi-ml-histopathology/src/histopathology/scripts/generate_crossval_report.py
fepegar: Not asking you use this, but I got curious and thought I'd share (maybe this should go to Pro tips? it's quite nice!):

Web capture_1-4-2022_184731_techdows com

hi-ml-histopathology/src/histopathology/utils/report_utils.py
dccastro: Not sure I fully understand what you mean. Currently (with the `DeepMILOutputsHandler`), we expect new experiments to produce `val/test_output.csv` and `test/test_output.csv` (among other output files). Legacy experiments (pre-outputs handler) have only `test_output.csv` (note there's no intermediate folder in this case).

I wrote this function to identify if we're in the first or second case, both of which I'd like to support for now. An exception is thrown if a run doesn't match either pattern.

hi-ml-histopathology/src/histopathology/utils/report_utils.py
dccastro: @fepegar This failure mode of mixed child runs should never happen, as Hyperdrive is managed by Azure ML and children execute with exactly the same code snapshot. If the expected files are indeed missing from any of the child runs, a `ValueError` will be raised instead from `run_has_val_test_outputs()` with the specific details.

hi-ml-histopathology/src/histopathology/scripts/generate_crossval_report.py
dccastro: @fepegar It works! 👀
image

hi-ml-histopathology/src/histopathology/utils/report_utils.py
fepegar: I just don't think this function should be expected to check whether the format is not valid and raise the exception.

hi-ml-histopathology/testhisto/testhisto/utils/test_report_utils.py
dccastro: I'm not sure this function is important enough to warrant that... There are other parts with little coverage where this approach makes more sense. We'll get around to those in time :)

hi-ml-histopathology/testhisto/testhisto/utils/test_report_utils.py
```diff
@@ -10,7 +10,76 @@
...
+        with pytest.raises(ValueError, match="does not have the expected files"):
+            run_has_val_and_test_outputs(run)
+
+    with patch.object(run, 'get_file_names', return_value=[test_outputs_filename]):
+        with pytest.raises(ValueError, match="does not have the expected files"):
+            run_has_val_and_test_outputs(run)
+
+    with patch.object(run, 'get_file_names', return_value=[arbitrary_filename]):
+        with pytest.raises(ValueError, match="does not have the expected files"):
+            run_has_val_and_test_outputs(run)
```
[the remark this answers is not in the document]
dccastro: I personally find the test more readable this way. Also, if one of the cases fails we can see immediately in the code which one it was, rather than working back from the filenames printed in the error message.

hi-ml-histopathology/testhisto/testhisto/utils/test_report_utils.py
dccastro: Here each test case is a bit different, so I don't think a loop would be much cleaner...

hi-ml-histopathology/src/histopathology/utils/report_utils.py
dccastro: I think it makes it explicit that either all children are of type A or all of type B, otherwise something has gone horribly wrong :)

hi-ml-histopathology/src/histopathology/scripts/generate_crossval_report.py
```diff
@@ -61,14 +61,37 @@ def generate_html_report(parent_run_id: str, output_dir: Path, workspace_config_
...
+    if include_test:
+        # Add test ROC and PR curves
+        test_outputs_filename = "test_output.csv"
+        if has_val_and_test_outputs:
+            test_outputs_filename = "test/" + test_outputs_filename
+        test_outputs_dfs = collect_crossval_outputs(parent_run_id, report_dir, aml_workspace,
+                                                    output_filename=test_outputs_filename,
+                                                    overwrite=overwrite)
+
+        report.add_heading("Test ROC and PR curves", level=3)
```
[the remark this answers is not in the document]
dccastro: I've now refactored into separate functions to reduce duplication as much as I could.

hi-ml-histopathology/src/histopathology/utils/report_utils.py
dccastro: I find the scope is pretty clear from the docs, and I honestly don't think this function will be heavily used anywhere else.

What would you propose instead? One alternative I see is to switch to just `get_validation_outputs()` and `get_test_outputs()`, which directly try to find the respective files in legacy and new locations, and return `None` if not found.

fepegar: My suggestion was to remove lines 33-36, but if you think it's not critical and it's safer to check if the file tree is not correct, I guess it's fine.

mebristo (approved):

PR #289, opened 2022-04-02
ENH: Add regression test framework to runner

ant0nsc:
Add functionality to the runner to compare results against a set of stored files.

### DIFF
--- a/hi-ml/src/health_ml/deep_learning_config.py
+++ b/hi-ml/src/health_ml/deep_learning_config.py
@@ -146,6 +146,18 @@ class WorkflowParams(param.Parameterized):
     hyperdrive: bool = param.Boolean(False, doc="If True, use the Hyperdrive configuration specified in the "
                                                 "LightningContainer to run hyperparameter tuning. If False, just "
                                                 "run a plain single training job.")
+    regression_test_folder: Optional[Path] = \
+        param.ClassSelector(class_=Path, default=None, allow_None=True,
+                            doc="A path to a folder that contains a set of files. At the end of training and "
+                                "model evaluation, all files given in that folder must be present in the job's output "
+                                "folder, and their contents must match exactly. When running in AzureML, you need to "
+                                "ensure that this folder is part of the snapshot that gets uploaded. The path should "
+                                "be relative to the repository root directory.")
+    regression_test_csv_tolerance: float = \
+        param.Number(default=0.0, allow_None=False,
+                     doc="When comparing CSV files during regression tests, use this value as the maximum allowed "
+                         "relative difference of actual and expected results. Default: 0.0 (must match exactly)")
+
     CROSSVAL_INDEX_ARG_NAME = "crossval_index"
     CROSSVAL_COUNT_ARG_NAME = "crossval_count"
 
--- a/hi-ml/src/health_ml/run_ml.py
+++ b/hi-ml/src/health_ml/run_ml.py
@@ -23,6 +23,7 @@ from health_ml.utils.common_utils import (
     EFFECTIVE_RANDOM_SEED_KEY_NAME, change_working_directory, logging_section,
     RUN_RECOVERY_ID_KEY, RUN_RECOVERY_FROM_ID_KEY_NAME)
 from health_ml.utils.lightning_loggers import StoringLogger
+from health_ml.utils.regression_test_utils import compare_folders_and_run_outputs
 from health_ml.utils.type_annotations import PathOrString
 
 
@@ -153,6 +154,26 @@ class MLRunner:
 
         with logging_section("Model inference"):
             self.run_inference(checkpoint_paths_for_testing)
+
+        if self.container.regression_test_folder:
+            # Comparison with stored results for cross-validation runs only operates on child run 0. This run
+            # has usually already downloaded the results for the other runs, and uploaded files to the parent
+            # run context.
+            logging.info("Comparing the current results against stored results")
+            if self.is_normal_run_or_crossval_child_0():
+                compare_folders_and_run_outputs(expected=self.container.regression_test_folder,
+                                                actual=self.container.outputs_folder,
+                                                csv_relative_tolerance=self.container.regression_test_csv_tolerance)
+            else:
+                logging.info("Skipping because this is not cross-validation child run 0.")
+
+    def is_normal_run_or_crossval_child_0(self) -> bool:
+        """
+        Returns True if the present run is a non-crossvalidation run, or child run 0 of a crossvalidation run.
+        """
+        if self.container.is_crossvalidation_enabled:
+            return self.container.crossval_index == 0
+        return True
 
     def run_inference(self, checkpoint_paths: List[Path]) -> None:
         """
--- a/hi-ml/src/health_ml/utils/regression_test_utils.py
+++ b/hi-ml/src/health_ml/utils/regression_test_utils.py
@@ -1,0 +1,170 @@
+#  -------------------------------------------------------------------------------------------
+#  Copyright (c) Microsoft Corporation. All rights reserved.
+#  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
+#  -------------------------------------------------------------------------------------------
+import logging
+import os
+from pathlib import Path
+import shutil
+import tempfile
+from typing import List, Optional
+
+from azureml.core import Run
+import pandas as pd
+
+from health_azure.utils import PARENT_RUN_CONTEXT, RUN_CONTEXT, is_running_in_azure_ml
+
+
+REGRESSION_TEST_OUTPUT_FOLDER = "OUTPUT"
+REGRESSION_TEST_AZUREML_FOLDER = "AZUREML_OUTPUT"
+REGRESSION_TEST_AZUREML_PARENT_FOLDER = "AZUREML_PARENT_OUTPUT"
+CONTENTS_MISMATCH = "Contents mismatch"
+FILE_FORMAT_ERROR = "File format error"
+MISSING_FILE = "Missing"
+CSV_SUFFIX = ".csv"
+TEXT_FILE_SUFFIXES = [".txt", ".json", ".html", ".md"]
+
+
+def compare_files(expected: Path, actual: Path, csv_relative_tolerance: float = 0.0) -> str:
+    """
+    Compares two individual files for regression testing. It returns an empty string if the two files appear identical.
+    If the files are not identical, an error message with details is return. This handles known text file formats,
+    where it ignores differences in line breaks. All other files are treated as binary, and compared on a byte-by-byte
+    basis.
+
+    :param expected: A file that contains the expected contents. The type of comparison (text or binary) is chosen
+    based on the extension of this file.
+    :param actual: A file that contains the actual contents.
+    :param csv_relative_tolerance: When comparing CSV files, use this as the maximum allowed relative discrepancy.
+    If 0.0, do not allow any discrepancy.
+    :return: An empty string if the files appear identical, or otherwise an error message with details.
+    """
+
+    def print_lines(prefix: str, lines: List[str]) -> None:
+        num_lines = len(lines)
+        count = min(5, num_lines)
+        logging.info(f"{prefix} {num_lines} lines, first {count} of those:")
+        logging.info(os.linesep.join(lines[:count]))
+
+    def try_read_csv(prefix: str, file: Path) -> Optional[pd.DataFrame]:
+        try:
+            return pd.read_csv(file)
+        except Exception as ex:
+            logging.info(f"{prefix} file can't be read as CSV: {str(ex)}")
+            return None
+
+    if expected.suffix == CSV_SUFFIX:
+        expected_df = try_read_csv("Expected", expected)
+        actual_df = try_read_csv("Actual", actual)
+        if expected_df is None or actual_df is None:
+            return FILE_FORMAT_ERROR
+        try:
+            pd.testing.assert_frame_equal(actual_df, expected_df, rtol=csv_relative_tolerance)
+        except Exception as ex:
+            logging.info(str(ex))
+            return CONTENTS_MISMATCH
+    elif expected.suffix in TEXT_FILE_SUFFIXES:
+        # Compare line-by-line to avoid issues with line separators
+        expected_lines = expected.read_text().splitlines()
+        actual_lines = actual.read_text().splitlines()
+        if expected_lines != actual_lines:
+            print_lines("Expected", expected_lines)
+            print_lines("Actual", actual_lines)
+            return CONTENTS_MISMATCH
+    else:
+        expected_binary = expected.read_bytes()
+        actual_binary = actual.read_bytes()
+        if expected_binary != actual_binary:
+            logging.info(f"Expected {len(expected_binary)} bytes, actual {len(actual_binary)} bytes")
+            return CONTENTS_MISMATCH
+    return ""
+
+
+def compare_folder_contents(expected_folder: Path,
+                            csv_relative_tolerance: float,
+                            actual_folder: Optional[Path] = None,
+                            run: Optional[Run] = None) -> List[str]:
+    """
+    Compares a set of files in a folder, against files in either the other folder or files stored in the given
+    AzureML run. Each file that is present in the "expected" folder must be also present in the "actual" folder
+    (or the AzureML run), with exactly the same contents, in the same folder structure.
+    For example, if there is a file "<expected>/foo/bar/contents.txt", then there must also be a file
+    "<actual>/foo/bar/contents.txt"
+
+    :param expected_folder: A folder with files that are expected to be present.
+    :param actual_folder: The output folder with the actually produced files.
+    :param run: An AzureML run
+    :param csv_relative_tolerance: When comparing CSV files, use this as the maximum allowed relative discrepancy.
+    If 0.0, do not allow any discrepancy.
+    :return: A list of human readable error messages, with message and file path. If no errors are found, the list is
+    empty.
+    """
+    messages = []
+    if run and not is_running_in_azure_ml(run):
+        logging.warning("Skipping file comparison because the given run context is an AzureML offline run.")
+        return []
+    files_in_run: List[str] = run.get_file_names() if run else []
+    temp_folder = Path(tempfile.mkdtemp()) if run else None
+    for file in expected_folder.rglob("*"):
+        # rglob also returns folders, skip those
+        if file.is_dir():
+            continue
+        # All files stored in AzureML runs use Linux-style path
+        file_relative = file.relative_to(expected_folder).as_posix()
+        if actual_folder:
+            actual_file = actual_folder / file_relative
+        elif temp_folder is not None and run is not None:
+            actual_file = temp_folder / file_relative
+            if file_relative in files_in_run:
+                run.download_file(name=str(file_relative), output_file_path=str(actual_file))
+        else:
+            raise ValueError("One of the two arguments run, actual_folder must be provided.")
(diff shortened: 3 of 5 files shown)


hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,170 @@
...
+REGRESSION_TEST_AZUREML_FOLDER = "AZUREML_OUTPUT"
+REGRESSION_TEST_AZUREML_PARENT_FOLDER = "AZUREML_PARENT_OUTPUT"
+CONTENTS_MISMATCH = "Contents mismatch"
+FILE_FORMAT_ERROR = "File format error"
+MISSING_FILE = "Missing"
+CSV_SUFFIX = ".csv"
+TEXT_FILE_SUFFIXES = [".txt", ".json", ".html", ".md"]
+
+
+def compare_files(expected: Path, actual: Path, csv_relative_tolerance: float = 0.0) -> str:
```
fepegar, line 28: Same comment as above regarding tolerance.

hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    def print_lines(prefix: str, lines: List[str]) -> None:
+        num_lines = len(lines)
+        count = min(5, num_lines)
+        logging.info(f"{prefix} {num_lines} lines, first {count} of those:")
+        logging.info(os.linesep.join(lines[:count]))
+
+    def try_read_csv(prefix: str, file: Path) -> Optional[pd.DataFrame]:
+        try:
+            return pd.read_csv(file)
+        except Exception as ex:
```
fepegar, line 51: Can we catch a more specific exception here?

hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,169 @@
...
+            return None
+
+    if expected.suffix == CSV_SUFFIX:
+        expected_df = try_read_csv("Expected", expected)
+        actual_df = try_read_csv("Actual", actual)
+        if expected_df is None or actual_df is None:
+            return FILE_FORMAT_ERROR
+        try:
+            pd.testing.assert_frame_equal(actual_df, expected_df, rtol=csv_relative_tolerance)
+        except Exception as ex:
```
fepegar, line 62: Can we catch a more specific exception here too?

hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    Otherwise, an error message with details is returned. This handles known text file formats,
+    and differences in line breaks are ignored. All other files are treated as binary, and
+    compared on a byte-by-byte basis.
+
+    :param expected: A file that contains the expected contents. The type of comparison (text or binary) is chosen
+    based on the extension of this file.
+    :param actual: A file to be checked against the expected file.
+    :param csv_relative_tolerance: When comparing CSV files, maximum allowed relative discrepancy.
+    If 0.0, do not allow any discrepancy.
+    :return: An empty string if the files are identical, an error message with details otherwise.
```
fepegar, line 40: Which details does the message contain? It looks like they currently don't specify what the mismatch is.

hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    Compares a set of files in a folder, against files in either the other folder or files stored in the given
+    AzureML run. Each file that is present in the "expected" folder must be also present in the "actual" folder
+    (or the AzureML run), with exactly the same contents, in the same folder structure.
+    For example, if there is a file "<expected>/foo/bar/contents.txt", then there must also be a file
+    "<actual>/foo/bar/contents.txt"
+
+    :param expected_folder: A folder with files that are expected to be present.
+    :param actual_folder: The output folder with the actually produced files.
+    :param run: An AzureML run
+    :param csv_relative_tolerance: When comparing CSV files, use this as the maximum allowed relative discrepancy.
```
fepegar, line 96: Parameters in docstring are not sorted.

hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    Compares a set of files in a folder, against files in either the other folder or files stored in the given
+    AzureML run. Each file that is present in the "expected" folder must be also present in the "actual" folder
+    (or the AzureML run), with exactly the same contents, in the same folder structure.
+    For example, if there is a file "<expected>/foo/bar/contents.txt", then there must also be a file
+    "<actual>/foo/bar/contents.txt"
+
+    :param expected_folder: A folder with files that are expected to be present.
+    :param actual_folder: The output folder with the actually produced files.
+    :param run: An AzureML run
```
fepegar, line 95: Can we elaborate on this? Should only `run` or `actual_folder` be passed? What happens if both or neither are passed?

hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,169 @@
...
+        # All files stored in AzureML runs use Linux-style path
+        file_relative = file.relative_to(expected_folder).as_posix()
+        if actual_folder:
+            actual_file = actual_folder / file_relative
+        elif temp_folder is not None and run is not None:
+            actual_file = temp_folder / file_relative
+            if file_relative in files_in_run:
+                run.download_file(name=str(file_relative), output_file_path=str(actual_file))
+        else:
+            raise ValueError("One of the two arguments run, actual_folder must be provided.")
```
fepegar, line 120: Something odd with this message.

hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    :param actual_folder: The output folder with the actually produced files.
+    :param run: An AzureML run
+    :param csv_relative_tolerance: When comparing CSV files, use this as the maximum allowed relative discrepancy.
+    If 0.0, do not allow any discrepancy.
+    :return: A list of human readable error messages, with message and file path. If no errors are found, the list is
+    empty.
+    """
+    messages = []
+    if run and not is_running_in_azure_ml(run):
+        logging.warning("Skipping file comparison because the given run context is an AzureML offline run.")
```
fepegar, line 103: I think we should be consistent with periods at the end of printed out messages. I guess they should be removed (same for `ValueError` below.

hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,169 @@
...
+    :param expected: A folder with files that are expected to be present.
+    :param actual: The output folder with the actually produced files.
+    :param csv_relative_tolerance: When comparing CSV files, use this as the maximum allowed relative discrepancy.
+    If 0.0, do not allow any discrepancy.
+    """
+    if not expected.is_dir():
+        raise ValueError(f"Folder with expected files does not exist: {expected}")
+    logging.debug(f"Current working directory: {Path.cwd()}")
+    messages = []
+    for (subfolder, message_prefix, actual_folder, run_to_compare) in \
```
fepegar, line 148: Can we create the iterable first and use `zip` in the for loop to improve readability?

hi-ml/testhiml/conftest.py
```diff
@@ -4,8 +4,12 @@
 
 
 root = Path(__file__).parent.parent.parent
-for folder in ["hi-ml-azure", "hi-ml"]:
-    full_folder = str(root / folder / "src")
+for folder in [
+    Path("hi-ml-azure") / "src",
+    Path("hi-ml-azure") / "testazure",
+    Path("hi-ml") / "src",
+]:
```
fepegar, line 11: As above, instantiating the iterable before the loop will improve readability.

hi-ml/testhiml/testhiml/test_regression_tests.py
```diff
@@ -0,0 +1,239 @@
...
+from testazure.utils_testazure import DEFAULT_WORKSPACE
+
+
+def create_folder_and_write_text(file: Path, text: str) -> None:
+    """
+    Writes the given text to a file. The folders in which the file lives are created too, unless they exist already.
+    Writing the text keeps the line separators as-is (no translation)
+    """
+    file.parent.mkdir(exist_ok=True, parents=True)
+    with file.open(mode="wt", newline="") as f:
```
fepegar, line 36: `t` is the default mode so it can be omitted.

hi-ml/testhiml/testhiml/test_regression_tests.py
```diff
@@ -0,0 +1,239 @@
...
+    compare_folder_contents,
+    compare_folders_and_run_outputs,
+)
+from testazure.utils_testazure import DEFAULT_WORKSPACE
+
+
+def create_folder_and_write_text(file: Path, text: str) -> None:
+    """
+    Writes the given text to a file. The folders in which the file lives are created too, unless they exist already.
+    Writing the text keeps the line separators as-is (no translation)
```
fepegar, line 33: Period missing.

hi-ml/testhiml/testhiml/test_regression_tests.py
```diff
@@ -0,0 +1,239 @@
...
+    expected.write_text(valid_csv)
+    actual.write_text(valid_csv)
+    assert compare_files(expected=expected, actual=actual) == ""
+
+
+@pytest.mark.parametrize("file_extension", [".png", ".whatever"])
+def test_compare_files_binary(tmp_path: Path, file_extension: str) -> None:
+    """
+    Checks the comparison of files that are not recognized as text files, for example images.
+    :param test_output_dirs:
```
fepegar, line 109: Should this be here?

hi-ml/testhiml/testhiml/test_regression_tests.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+def upload_to_run_and_compare(regression_test_subfolder: str, run_to_mock: str, tmp_path: Path) -> None:
+    """Creates a set of files in an AzureML run, and checks if the comparison tools accept/fail
+    in the right way.
+
+    :param regression_test_subfolder: The subfolder of the regression test results where the files
+    should be created (either REGRESSION_TEST_AZUREML_FOLDER or REGRESSION_TEST_AZUREML_PARENT_FOLDER)
+    :param run_to_mock: either RUN_CONTEXT or PARENT_RUN_CONTEXT
+    """
```
fepegar, line 210: `tmp_path` is missing from docstring.

fepegar (changes requested):

hi-ml/src/health_ml/run_ml.py
```diff
@@ -154,6 +155,26 @@ def run(self) -> None:
...
+            # run context.
+            logging.info("Comparing current results against stored results")
+            if self.is_normal_run_or_crossval_child_0():
+                compare_folders_and_run_outputs(expected=self.container.regression_test_folder,
+                                                actual=self.container.outputs_folder,
+                                                csv_relative_tolerance=self.container.regression_test_csv_tolerance)
+            else:
+                logging.info("Skipping as this is not cross-validation child run 0")
+
+    def is_normal_run_or_crossval_child_0(self) -> bool:
```
mebristo, line 170: 'Normal run' is a bit unclear - could we rename this `is_crossval_disabled_or_child_zero` perhaps?

hi-ml/testhiml/testhiml/test_regression_tests.py
```diff
@@ -0,0 +1,239 @@
...
+    ignored = "ignored.txt"
+    # Comparison should cover at least .csv and .txt files
+    mismatch = "mismatch.csv"
+    extra = "extra.txt"
+    subfolder = Path("folder")
+    # This file exists in both expected and actual, should not raise any alerts because it contents matches
+    # apart from linebreaks
+    create_folder_and_write_text(expected / subfolder / matching, "Line1\r\nLine2")
+    create_folder_and_write_text(actual / subfolder / matching, "Line1\nLine2")
+    # This file only exists in the expected results, and should create an error
```
mebristo, line 141: This comment is a bit confusing because it appears to apply to line 143 (containing 'extra') as well, but in actual fact we do not expect 'extra' to appear in messages

hi-ml/testhiml/testhiml/test_regression_tests.py
```diff
@@ -0,0 +1,239 @@
...
+        file1 = folder / "output.txt"
+        create_folder_and_write_text(file1, "Something")
+    # First comparison should pass
+    compare_folders_and_run_outputs(expected=expected, actual=actual, csv_relative_tolerance=0.0)
+    # Now add a file to the set of expected files that does not exist in the run: comparison should now fail
+    no_such_file = "no_such_file.txt"
+    file2 = expected / no_such_file
+    create_folder_and_write_text(file2, "foo")
+    with pytest.raises(ValueError) as ex:
+        compare_folders_and_run_outputs(expected=tmp_path, actual=Path.cwd(), csv_relative_tolerance=0.0)
```
mebristo, line 176: Is there a reason we compare tmp_path and cwd here rather than expected and actual? Isn't it possible that tmp_path and cwd already did not match?

hi-ml/src/health_ml/utils/regression_test_utils.py
ant0nsc: better now?

fepegar: I don't think any changes since I reviewed. But now I've re-read and understood what you mean. I would suggest something like `Either "run" or "actual_folder" must be provided`.

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -146,6 +146,18 @@ class WorkflowParams(param.Parameterized):
...
+        param.ClassSelector(class_=Path, default=None, allow_None=True,
+                            doc="A path to a folder that contains a set of files. At the end of training and "
+                                "model evaluation, all files given in that folder must be present in the job's output "
+                                "folder, and their contents must match exactly. When running in AzureML, you need to "
+                                "ensure that this folder is part of the snapshot that gets uploaded. The path should "
+                                "be relative to the repository root directory.")
+    regression_test_csv_tolerance: float = \
+        param.Number(default=0.0, allow_None=False,
+                     doc="When comparing CSV files during regression tests, use this value as the maximum allowed "
+                         "relative difference of actual and expected results. Default: 0.0 (must match exactly)")
```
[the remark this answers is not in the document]
ant0nsc: I agree that in general "match exactly" is bad for floating points, and I would not advise anyone to use . However, here, the dataframes in question are always read from CSV files, where they already may have been slightly truncated. All sources of randomness should be completely controlled for, so I would indeed expect an exact match.

hi-ml/src/health_ml/utils/regression_test_utils.py
ant0nsc: I tried a zip solution, but I don't really think it improves readability. Have now at least taken the list out of the `for` statement

hi-ml/src/health_ml/utils/regression_test_utils.py
ant0nsc: All removed

hi-ml/src/health_ml/utils/regression_test_utils.py
ant0nsc: I'm deliberately generous here, because I want all errors to be collected _before_ the program terminates because it sees discrepancies. I've changed the logging here though to print exception details as a warning.

hi-ml/src/health_ml/utils/regression_test_utils.py
ant0nsc: done

hi-ml/src/health_ml/utils/regression_test_utils.py
ant0nsc: Clarifid.

hi-ml/testhiml/testhiml/test_regression_tests.py
ant0nsc: I've tried to clarify.

hi-ml/testhiml/testhiml/test_regression_tests.py
ant0nsc: good catch. fixed.

hi-ml/src/health_ml/utils/regression_test_utils.py
```diff
@@ -0,0 +1,180 @@
...
+) -> List[str]:
+    """
+    Compares a set of files in a folder, against files in either the other folder or files stored in the given
+    AzureML run. Each file that is present in the "expected" folder must be also present in the "actual" folder
+    (or the AzureML run), with exactly the same contents, in the same folder structure.
+    For example, if there is a file "<expected>/foo/bar/contents.txt", then there must also be a file
+    "<actual>/foo/bar/contents.txt". If `actual_folder` is provided, then this is used to compare files against the
+    set file files in `expected_folder`, irrespective of the value in `run`. If `run` is provided, the files
+    uploaded to the AzureML run are compared against files in `expected_folder`. If neither `run` nor `actual_folder`
+    are provided, a ValueError is raised.
```
fepegar, line 98: We should improve the docstring syntax for Sphinx. This is written in Markdown style.

**Current**

```
    Compares a set of files in a folder, against files in either the other folder or files stored in the given
    AzureML run. Each file that is present in the "expected" folder must be also present in the "actual" folder
    (or the AzureML run), with exactly the same contents, in the same folder structure.
    For example, if there is a file "<expected>/foo/bar/contents.txt", then there must also be a file
    "<actual>/foo/bar/contents.txt". If `actual_folder` is provided, then this is used to compare files against the
    set file files in `expected_folder`, irrespective of the value in `run`. If `run` is provided, the files
    uploaded to the AzureML run are compared against files in `expected_folder`. If neither `run` nor `actual_folder`
    are provided, a ValueError is raised.
```

Web capture_5-4-2022_104253_wsl%24

**Slightly better**

```
    Compares a set of files in a folder, against files in either the other folder or files stored in the given
    AzureML run. Each file that is present in the ``expected_folder`` must be also present in the ``actual_folder``
    (or the AzureML run), with exactly the same contents, in the same folder structure.
    For example, if there is a file ``<expected>/foo/bar/contents.txt``, then there must also be a file
    ``<actual>/foo/bar/contents.txt``. If ``actual_folder`` is provided, then this is used to compare files against the
    set file files in ``expected_folder``, irrespective of the value in ``run``. If ``run`` is provided, the files
    uploaded to the AzureML run are compared against files in ``expected_folder``. If neither ``run`` nor `actual_folder`
    are provided, a :exc:`ValueError` is raised.

    :param expected_folder: A folder with files that are expected to be present.
    :param actual_folder: The output folder with the actually produced files.
    :param run: An AzureML run
    :param csv_relative_tolerance: When comparing CSV files, use this as the maximum allowed relative discrepancy.
    If 0.0, do not allow any discrepancy.
    :return: A list of human readable error messages, with message and file path. If no errors are found, the list is
    empty.
```

Web capture_5-4-2022_105149_wsl%24

fepegar: Even more important when other classes or methods in the library are referenced in the docstring.

fepegar: See [a TorchIO class](https://torchio.readthedocs.io/transforms/augmentation.html#randomelasticdeformation) for heavy usage of rst syntax for rich rendering.

fepegar (changes requested):

fepegar (approved):

mebristo (approved):

Issue #292, opened 2022-04-04
Mypy is potentially misconfigured


dccastro:
I've found that `mypy` has been giving inconsistent results when running from CLI, in VS Code, and in our PR builds.

Digging a bit, I encountered the [`mypy_path` config option](https://mypy.readthedocs.io/en/stable/config_file.html#confval-mypy_path), which enables configuring the search paths. In particular, setting
```ini
mypy_path = hi-ml/src,hi-ml-azure/src,hi-ml-histopathology
; These paths can also be made relative to `$MYPY_CONFIG_FILE_DIR/`
```
in `mypy.ini` would enable it to correctly follow imports between our first-party packages, which helps `mypy` surface more typing issues. It can then be invoked as intended, e.g. from `hi-ml/` (project root):
```sh
$ mypy --config=mypy.ini --install-types --non-interactive -p histopathology
```

PR #294, opened 2022-04-06
STYLE: Fix mypy misconfiguration for histopathology

dccastro:
This PR fixes the `mypy` configurations in `mypy.ini`, `settings.json`, and `Makefiles` so that it runs correctly on the whole codebase. In particular, the `hi-ml-histopathology` folder was not being discovered correctly, so now I was able to identify and fix all the longstanding, previously hidden typing errors.

Other changes include moving the slide loading benchmarks into `hi-ml-histopathology/other` and minor fixes relating to out-of-date API usage.


.vscode/settings.json
```diff
@@ -25,6 +25,7 @@
         "--config=${workspaceFolder}/.flake8",
     ],
     "python.linting.mypyEnabled": true,
+    "python.linting.mypyPath": "/anaconda/envs/himl/bin/mypy",
```
dccastro, line 28: I added this here because `mypy` had been refusing to work with the `HimlHisto` environment. Is it reasonable to expect anyone wanting to use `mypy` linting to have the `himl` environment installed?

ant0nsc (approved):

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
```diff
@@ -45,8 +45,7 @@ def get_splits(self) -> Tuple[PandaDataset, PandaDataset, PandaDataset]:
                                                 proportion_train=.8,
                                                 proportion_test=.1,
                                                 proportion_val=.1,
-                                                subject_column=dataset.TILE_ID_COLUMN,
-                                                group_column=dataset.SLIDE_ID_COLUMN)
```
vale-salvatelli, line 49: why are we deleting slide_id? this seems unelated to mypy

vale-salvatelli:
Given some files have been moved from hi-ml to hi-ml-histopathology and there are a few changes in the code, could we have a more explanatory PR description?

hi-ml-histopathology/src/histopathology/datamodules/panda_module.py
dccastro: This is the PANDA slides dataset, so here we're splitting the dataset by slide ID. `mypy` was correctly complaining that `PandaDataset` does not have a `TILE_ID_COLUMN`.

dccastro:
Closes #292

PR #300, opened 2022-04-11
ENH: Unit test for current deepmil pipeline with mock data 

kenza-bouzid:
Add a unit test for deepmil container using mock data created with pathmnist dataset.

Main contributions: 
* Refactoring of MockWSIGenerator 
* MockDeepMIL class to run a small test
* MockTilesDataModule to generate mock tiles on the fly through pytest fixture


hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
```diff
@@ -24,17 +24,17 @@ def mock_wsi_root_dir(tmpdir_factory: pytest.TempdirFactory) -> py.path.local:
     tmp_root_dir = tmpdir_factory.mktemp("mock_wsi")
     wsi_generator = MockWSIGenerator(
         tmp_path=tmp_root_dir,
-        mock_type=MockWSIType.PATHMNIST,
-        batch_size=1,
-        n_samples=4,
+        mock_type=MockHistoDataType.PATHMNIST,
+        n_tiles=1,
```
vale-salvatelli, line 28: not entirely sure `batch_size` became `n_tiles`. Shouldn't these be separate parameters?

hi-ml-histopathology/testhisto/testhisto/mocks/base_datamodule.py
```diff
@@ -0,0 +1,103 @@
...
+        n_channels: int = 3,
+        tile_size: int = 28,
+    ) -> None:
+        """
+        :param tmp_path: A temporary directory to store all generated data.
+        :param mock_type: The wsi generator mock type. Supported mock types are:
+            WSIMockType.PATHMNIST: for creating mock WSI by stitching tiles from pathmnist.
+            WSIMockType.FAKE: for creating mock WSI by stitching fake tiles.
+        :param seed: pseudorandom number generator seed to use for mocking random metadata, defaults to 42.
+        :param n_tiles: how many tiles per batch to load from pathmnist dataloader, defaults to 1.
```
vale-salvatelli, line 62:
```suggestion
        :param n_tiles: how many tiles per slide to load from pathmnist dataloader, defaults to 1.
```

hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
kenza-bouzid: I felt like batch_size is a bit misleading as it not a batch_size for the MockGenerator class itself but rather the batch_size of the dataloader. It's used a bit differently in MockWSI and MockTiles.

It is supposed to define the number of tiles we would like to use to create a Mock WSI. If set to 1 the same tile is repeated n_repeat_tile  times, however if we want a WSI with different tiles we just need to increase it.

As for MockTiles, it determines the number of tiles we would like to have per slide. Speaking of which, we can maybe pick a random number of tiles per slide in the range of n_tiles instead of a fixed number. something like np.random.randint(n_tiles - 5, n_tiles +5), what do you think? 

This is mainly why I renamed the batch_size to n_tiles as it's just the number of tiles we want to sample from the dataloader depending on the type of mock data we want to create.

hi-ml-histopathology/testhisto/testhisto/mocks/deepmil.py
```diff
@@ -0,0 +1,139 @@
...
+            batch_size=self.batch_size,
+            max_bag_size_inf=self.max_bag_size_inf,
+            transform=transform,
+            cache_mode=self.cache_mode,
+            precache_location=self.precache_location,
+            cache_dir=self.cache_dir,
+            crossval_count=self.crossval_count,
+            crossval_index=self.crossval_index,
+            dataloader_kwargs=dataloader_kwargs,
+        )
```
vale-salvatelli, line 107: I think we're rewriting too much of the logic in this test. We risk the test to go out of sync with BaseMIL and at that point it would no longer be useful. My suggestion is 

- where possible to make use of `@mock.patch` t patch the methods/function we want to overwrite - here for example can we patch TilesDataModule with MockTilesDataModule?
- refactor bits of basemil, here for example you could have a separate class method that returns `transform, dataloader_kwargs` that incorporates the logic you're copying

kenza-bouzid: Good point, I'll look into that!

hi-ml-histopathology/testhisto/testhisto/mocks/deepmil.py
```diff
@@ -0,0 +1,139 @@
...
+            self.best_checkpoint_filename_with_suffix,
+        )
+        if absolute_checkpoint_path_parent.is_file():
+            return absolute_checkpoint_path_parent
+
+        checkpoint_path = get_best_checkpoint_path(Path(self.checkpoint_folder_path))
+        if checkpoint_path.is_file():
+            return checkpoint_path
+
+        raise ValueError("Path to best checkpoint not found")
```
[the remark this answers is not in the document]
kenza-bouzid: I wasn't sure if this is needed. I tried to reproduce the DeepSMILEPanda, Oh and I actually moved these two methods to BaseMIL in my refactoring branch as it was the same for DeepSMILEPanda and DeepSMILECrck. It's probably only needed at inference time right? I'll discard it from the mock

hi-ml-histopathology/testhisto/testhisto/mocks/tiles_datamodule.py
```diff
@@ -0,0 +1,165 @@
...
+import torch
+from torchvision.utils import save_image
+
+from health_ml.utils.split_dataset import DatasetSplits
+from histopathology.datamodules.base_module import TilesDataModule
+from histopathology.datasets.base_dataset import TilesDataset
+from testhisto.mocks.base_datamodule import MockHistoDataGenerator
+
+
+class MockTilesDataset(TilesDataset):
```
vale-salvatelli, line 19: should this be `MockPandaTilesDataset` given it has some settings specific to PANDA?

hi-ml-histopathology/testhisto/testhisto/mocks/tiles_datamodule.py
```diff
@@ -0,0 +1,165 @@
...
+            proportion_test=0.1,
+            proportion_val=0.1,
+            subject_column=dataset.TILE_ID_COLUMN,
+            group_column=dataset.SLIDE_ID_COLUMN,
+        )
+        return (
+            MockTilesDataset(self.root_path, dataset_df=splits.train),
+            MockTilesDataset(self.root_path, dataset_df=splits.val),
+            MockTilesDataset(self.root_path, dataset_df=splits.test),
+        )
```
vale-salvatelli, line 50: This is another place where I would use patching to call `MockTilesDataset` instead of `TilesDataset` - tests should not replicate the logic of the code we want to test

hi-ml-histopathology/testhisto/testhisto/mocks/tiles_datamodule.py
```diff
@@ -0,0 +1,165 @@
...
+                    f"_{slide_id}",
+                    f"_{slide_id}.{tile_x}x_{tile_y}y",
+                    f"_{slide_id}/train_images/{tile_x}x_{tile_y}y.png",
+                    f"_{slide_id}/train_images/{tile_x}x_{tile_y}y_mask.png",
+                    tile_x,
+                    tile_y,
+                    1.0,
+                    data_provider,
+                    isup_grade,
+                    gleason_score,
```
vale-salvatelli, line 152: Minor: building a dictionary using as keys the column names defined in rows 111-121 would have been a bit more readable and less prone to mismatch in the fields order, a df can be created from the dictionary. just an idea, no need to change now.

hi-ml-histopathology/testhisto/testhisto/mocks/tiles_datamodule.py
```diff
@@ -0,0 +1,165 @@
...
+            group_column=dataset.SLIDE_ID_COLUMN,
+        )
+        return (
+            MockTilesDataset(self.root_path, dataset_df=splits.train),
+            MockTilesDataset(self.root_path, dataset_df=splits.val),
+            MockTilesDataset(self.root_path, dataset_df=splits.test),
+        )
+
+
+class MockTilesGenerator(MockHistoDataGenerator):
```
vale-salvatelli, line 53: Maybe also this should be `MockPandaTilesGenerator` because we recreate some of the Panda dataset characteristics

hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
```diff
@@ -255,38 +314,37 @@ def test_container(container_type: Type[LightningContainer], use_gpu: bool) -> N
...
 
-    test_data_loader = data_module.test_dataloader()
-    for batch_idx, batch in enumerate(test_data_loader):
-        batch = move_batch_to_expected_device(batch, use_gpu)
-        outputs_dict = module.test_step(batch, batch_idx)
-        loss = outputs_dict[ResultsKey.LOSS]  # noqa
-        assert loss.shape == ()
-        assert isinstance(loss, Tensor)
-        break
+@pytest.mark.parametrize("use_gpu", [True, False])
```
vale-salvatelli, line 329: I would parametrize over a few parameters of MockDeepSMILE, for example cache_mode, cache_location, is_finetune. Also encoder_type might be useful but I think it would add to much scope to this PR

hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
```diff
@@ -255,38 +314,37 @@ def test_container(container_type: Type[LightningContainer], use_gpu: bool) -> N
...
+    module = container.create_model()
+
+    module.trainer = MagicMock(world_size=1)  # type: ignore
+    module.log = MagicMock()  # type: ignore
+    if use_gpu:
+        module.cuda()
+
+    assert_train_step(module, data_module, use_gpu)
+    assert_validation_step(module, data_module, use_gpu)
+    assert_test_step(module, data_module, use_gpu)
```
vale-salvatelli, line 347: PR looks good overall - great work! My major comment is about avoiding replicating logic in the mock classes, otherwise we will use the usefulness of these tests

vale-salvatelli (changes requested):

hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
vale-salvatelli: got you, makes sense. `we can maybe pick a random number of tiles per slide in the range of n_tiles ` I think this is a great idea, it would make the mock dataset much closer to the real use case and a number of issues can come from the fact that not all the slides have the same number of tiles so better to test the case

hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
kenza-bouzid: Thank you for the valuable feedback. I wasn't familiar with patching, I am reading into it and will make the changes.

dccastro (changes requested):
Great job, @kenza-bouzid! It's clear you put a lot of work into this, and, as @vale-salvatelli said, this will certainly be useful going forward. Most of my comments are minor suggestions for clarity.

I'm also wondering if you've thought of writing unit tests for the mock classes themselves, to make sure the generated dataset is up to the spec?

hi-ml-histopathology/testhisto/testhisto/datamodules/test_slides_datamodule.py
```diff
@@ -24,17 +24,17 @@ def mock_wsi_root_dir(tmpdir_factory: pytest.TempdirFactory) -> py.path.local:
     tmp_root_dir = tmpdir_factory.mktemp("mock_wsi")
```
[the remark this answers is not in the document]
kenza-bouzid: Oh good point, I missed this part of the documentation. I chose TempdirFactory so that we create the mock data only once per session, but  TempPathFactory is definitely a better option, I had to alter the typing annotations in the other classes to support py.path.local consequently. Thanks for pointing it out!

hi-ml-histopathology/testhisto/testhisto/mocks/base_datamodule.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def create_mock_metadata_dataframe(self) -> pd.DataFrame:
+        """Create a mock dataframe with random metadata."""
+        raise NotImplementedError
+
+    def get_dataloader(self) -> Optional[DataLoader]:
+        if self.mock_type == MockHistoDataType.PATHMNIST:
+            return self._get_pathmnist_dataloader()
+        elif self.mock_type == MockHistoDataType.FAKE:
+            return None
```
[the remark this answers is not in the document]
kenza-bouzid: In the case of  MockHistoDataType.FAKE we don't really need a dataloader (as in pathmnist) cause we're just gonna create a tile with the same random "fake" value   (see lines 134-150). Should I maybe add a commentt to clarify it?

hi-ml-histopathology/testhisto/testhisto/mocks/base_datamodule.py
```diff
@@ -0,0 +1,103 @@
...
+        else:
+            raise NotImplementedError
+
+    def _get_pathmnist_dataloader(self) -> DataLoader:
+        """Get a dataloader for pathmnist dataset. It returns tiles of shape (batch_size, 3, 28, 28).
+        :return: A dataloader to sample pathmnist tiles.
+        """
+        info = INFO["pathmnist"]
+        DataClass = getattr(medmnist, info["python_class"])
+        data_transform = transforms.Compose([transforms.ToTensor()])
```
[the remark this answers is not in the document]
kenza-bouzid: Definitely, I was experimenting with other transforms for scaling and normalization and forgot to remove the compose

hi-ml-histopathology/testhisto/testhisto/mocks/slides_datamodule.py
```diff
@@ -237,11 +173,11 @@ def _create_multi_resolution_wsi(self, mock_image: np.ndarray) -> List[np.ndarra
         levels = [mock_image[:: 2 ** i, :: 2 ** i] for i in range(self.n_levels)]
         return levels
 
-    def generate_mock_wsi(self) -> None:
+    def generate_mock_histo_data(self) -> None:
         """Create mock wsi and save them as tiff files"""
-        for sample_counter in range(self.n_samples):
+        for sample_counter in range(self.n_slides):
             tiles, _ = next(iter(self.dataloader)) if self.dataloader else None, None
```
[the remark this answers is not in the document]
kenza-bouzid: Sounds great!

hi-ml-histopathology/testhisto/testhisto/mocks/slides_datamodule.py
```diff
@@ -237,11 +173,11 @@ def _create_multi_resolution_wsi(self, mock_image: np.ndarray) -> List[np.ndarra
...
-            mock_image, dump_tiles = self.create_wsi_from_stitched_tiles(tiles[0])
+            mock_image, dump_tiles = self._create_wsi_from_stitched_tiles(tiles[0])
```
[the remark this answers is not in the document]
kenza-bouzid: Good catch! thanks.

hi-ml-histopathology/testhisto/testhisto/mocks/tiles_datamodule.py
```diff
@@ -0,0 +1,165 @@
...
+                self.ISUP_GRADE,
+                self.GLEASON_SCORE,
+            ]
+        )
+        grid = [
+            f"{i}_{j}" for i in range(self.img_size // self.tile_size) for j in range(self.img_size // self.tile_size)
+        ]
+        coords = list(
+            map(lambda x: list(map(int, x.split("_"))), np.random.choice(grid, size=self.n_tiles, replace=False))
+        )
```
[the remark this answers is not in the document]
kenza-bouzid: because [np.random.choice](https://numpy.org/doc/stable/reference/random/generated/numpy.random.choice.html) only takes a 1D array as input, it was just a hacky way to sample random distinct 2D coordinates, but I like your way even better! Thanks for the suggestion!

hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
kenza-bouzid: I tried running that locally, but it crashed my VM (too many cominations). I don't know on which nodes the CI runs but it might make it super slow. We might need to parametrize more but not all of them.

dccastro (approved):

PR #303, opened 2022-04-13
ENH: Add `is_caching` parameter in DeepMIL container

harshita-s:
In this PR:
- `is_caching` is added as a param in the DeepMIL containers to enable or disable caching during training. If caching is disabled, random subsampling of tiles is possible at each epoch.
- When fine-tuning, caching will be disabled by default.


hi-ml-histopathology/src/histopathology/utils/download_utils.py
```diff
@@ -20,12 +20,12 @@ def download_file_if_necessary(run_id: str, remote_dir: Path, download_dir: Path
     """
     aml_workspace = get_workspace()
     os.chdir(fixed_paths.repository_root_directory())
-    local_path = download_dir / run_id.split(":")[1] / "outputs" / filename
+    local_path = download_dir / run_id.split(":")[1] / "outputs/test" / filename
```
dccastro, line 23: Why is it necessary to hardcode this "test/" subdirectory here? I'd avoid this, if possible, as I believe this is meant as a general-purpose function. If the desired file is now generated in a different directory, I suggest changing the caller to pass `filename="test/my_file"`.

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILECrck.py
```diff
@@ -95,8 +95,17 @@ def setup(self) -> None:
             self.downloader.download_checkpoint_if_necessary()
 
         self.encoder = self.get_encoder()
-        self.encoder.cuda()
-        self.encoder.eval()
+        if not self.is_finetune:
```
dccastro, line 98: May I propose to avoid replicating this logic in each child container? How about the following:
```python
class BaseMIL:
  def setup(self):
    # without the "if self.encoder_type ..." part
    self.encoder = self.get_encoder()
    ...

class DeepSMILECrck(BaseMIL):  # likewise for DeepSMILEPanda
  def setup(self):
    if self.encoder_type == SSLEncoder.__name__:
      ...
    super().setup()
```

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -92,6 +96,15 @@ def setup(self) -> None:
         self.encoder = self.get_encoder()
         if not self.is_finetune:
             self.encoder.eval()
+        # Fine-tuning requires tiles to be loaded on-the-fly, hence, caching is disabled by default.
+        if self.is_finetune:
+            self.is_caching = False
+        if self.is_caching:
+            self.cache_mode = CacheMode.MEMORY
+            self.precache_location = CacheLocation.CPU
+        else:
```
dccastro, line 105: How about having `is_caching` override the caching options only in the negative case? That way, users can still configure `cache_mode` and `precache_location` if `is_caching is False`.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -92,6 +96,15 @@ def setup(self) -> None:
         self.encoder = self.get_encoder()
         if not self.is_finetune:
             self.encoder.eval()
+        # Fine-tuning requires tiles to be loaded on-the-fly, hence, caching is disabled by default.
```
dccastro, line 99: [optional] Consider moving the configuration below into `__init__()` to ensure that these parameters are correctly set as soon as the container is created.
```python
class BaseMIL:
  def __init__(self, **kwargs):
    super().__init__(**kwargs)
    if self.is_finetune:
      ...
```

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -95,6 +94,15 @@ def setup(self) -> None:
         self.encoder = self.get_encoder()
         if not self.is_finetune:
             self.encoder.eval()
+        # Fine-tuning requires tiles to be loaded on-the-fly, hence, caching is disabled by default.
```
dccastro, line 97: Also replicated here.

dccastro (changes requested):
Thanks for enabling this new no-caching + no-finetuning scenario! Made a few small suggestions to reduce replication and retain flexibility.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -142,10 +155,11 @@ def create_model(self) -> DeepMILModule:
         self.data_module = self.get_data_module()
         # Encoding is done in the datamodule, so here we provide instead a dummy
         # no-op IdentityEncoder to be used inside the model
-        if self.is_finetune:
+        if not self.is_caching:
```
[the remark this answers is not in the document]
harshita-s: Thanks for the suggestion! I have now swapped the conditions in `create_model()`, `get_transform()` and `get_dataloader_kwargs()`

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILECrck.py
harshita-s: Thanks, this is a good idea. I have made this change now.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
harshita-s: I agree, removed the positive case now. I have changed the default for `precache_location` in BaseMIL

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
harshita-s: Done, thanks

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
harshita-s: I've reverted this as per our discussion.

dccastro (approved):
This is looking great, @harshita-s; thanks for working on it!

ant0nsc (approved):

PR #304, opened 2022-04-14
ENH: Refactor DeepMIL modules to support slide dataloader

kenza-bouzid:
Refactor DeepMIL modules to support loading data with SlidesDataModule. 

Main contributions include:
* Refactoring of DeepMILModule into BaseDeepMILModule, TilesDeepMILModule and SlidesDeepMILModule
* Refactoring of deepmil containers with multi inheritance
* Update mock data generators
* Unit test for slide based mock panda slides deep smile


hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -93,6 +88,27 @@ def setup(self) -> None:
...
 
+    def download_ssl_checkpoint(self, run_id: str) -> CheckpointDownloader:
+        downloader = CheckpointDownloader(
+            aml_workspace=get_workspace(),
+            run_id=run_id,
+            checkpoint_filename="last.ckpt",
+            download_dir="outputs/",
+            remote_checkpoint_dir=Path(self.checkpoint_folder_path)
+        )
+        os.chdir(fixed_paths.repository_root_directory().parent)
```
ant0nsc, line 99: While you're modifying that, can we get rid of this os.chdir? Changing working directory is brittle and error prone. Downloads can go into `self.outputs_folder` instead

hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -183,23 +183,20 @@ def configure_optimizers(self) -> optim.Optimizer:
...
         return getattr(self, f'{stage}_metrics')
 
+    def _compute_bag_labels_logits_and_attn_maps(self, batch: Dict) -> Tuple[Tensor, Tensor, List]:
+        """Compute bag labels, logits and attention maps depending on the nature of batch.
+        Subclasses (TilesDeepMILModule and SlidesDeepMILModule) should specify how this is handled depending on if
+        the tiles are fixed or generated on the fly from whole slide images.
+        """
+        raise NotImplementedError
+
+    def _update_results_with_data_specific_info(self, batch: dict, results: dict) -> None:
```
ant0nsc, line 193: here and in line 186: Methods that are expected to be overridden in subclasses should not be marked private.

hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -285,3 +272,68 @@ def test_epoch_end(self, epoch_results: EpochResultsType) -> None:  # type: igno
...
+        results.update({ResultsKey.SLIDE_ID: batch[TilesDataset.SLIDE_ID_COLUMN],
+                        ResultsKey.TILE_ID: batch[TilesDataset.TILE_ID_COLUMN],
+                        ResultsKey.IMAGE_PATH: batch[TilesDataset.PATH_COLUMN]})
+
+        if (TilesDataset.TILE_X_COLUMN in batch.keys()) and (TilesDataset.TILE_Y_COLUMN in batch.keys()):
+            results.update({ResultsKey.TILE_X: batch[TilesDataset.TILE_X_COLUMN],
+                           ResultsKey.TILE_Y: batch[TilesDataset.TILE_Y_COLUMN]}
+                           )
+        else:
+            if is_global_rank_zero():
```
ant0nsc, line 309: I think this could be done concisely with `from pytorch_lightning.utilities import rank_zero_warn`

hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -285,3 +272,68 @@ def test_epoch_end(self, epoch_results: EpochResultsType) -> None:  # type: igno
...
+                                "input tiles dataset.")
+
+
+class SlidesDeepMILModule(BaseDeepMILModule):
+    """Base class for slides based deep multiple-instance learning."""
+    def __init__(self, tiles_count: int, **kwargs: Any) -> None:
+        self.tiles_count = tiles_count
+        super().__init__(**kwargs)
+
+    def _compute_bag_labels_logits_and_attn_maps(self, batch: Dict) -> Tuple[Tensor, Tensor, List]:
```
ant0nsc, line 320: Code here is still rather similar to the method in TilesDeepMILModule apart from the label computation, can this be done better?

hi-ml-histopathology/testhisto/testhisto/mocks/slides_generator.py
```diff
@@ -10,19 +11,26 @@
...
 
+class TilesPositioningType(Enum):
+    DIAGONAL = 0
+    RANDOM = 1
+
+
 class MockPandaSlidesGenerator(MockHistoDataGenerator):
     """Generator class to create mock WSI on the fly. A mock WSI resembles to:
                                 [**      ]
                                 [  **    ]
```
ant0nsc, line 27: I think the comment is no longer valid if "random" is used?

hi-ml-histopathology/testhisto/testhisto/mocks/tiles_generator.py
```diff
@@ -55,22 +57,23 @@ def create_mock_metadata_dataframe(self) -> pd.DataFrame:
...
 
             data_provider = np.random.choice(self.DATA_PROVIDERS_VALUES)
             isup_grade = isup_grades[slide_id]
             gleason_score = np.random.choice(self.ISUP_GRADE_MAPPING[isup_grade])
 
-            # pick a random n_tiles for each slide
-            n_tiles = np.random.randint(self.n_tiles // 2 + 1, 3 * self.n_tiles // 2)
+            # pick a random n_tiles for each slide without exceeding the max n_tiles allowed
+            n_tiles: int = min(
+                np.random.randint(self.n_tiles // 2 + 1, 3 * self.n_tiles // 2), (self.img_size // self.tile_size) ** 2
```
dccastro, line 73: [optional] Minor suggestion to define `max_n_tiles` separately for clarity.

hi-ml-histopathology/testhisto/testhisto/mocks/tiles_generator.py
```diff
@@ -94,17 +97,27 @@ def create_mock_metadata_dataframe(self) -> pd.DataFrame:
 
         df = pd.DataFrame(data=mock_metadata)
         df.to_csv(os.path.join(self.tmp_path, PandaTilesDataset.DEFAULT_CSV_FILENAME), index=False)
+        self.n_tiles: int = tiles_count
```
dccastro, line 100: Isn't `self.n_tiles` already defined? Would it make sense to distinguish tiles per slide from total tiles?

hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
```diff
@@ -320,18 +338,13 @@ def test_container(container_type: Type[LightningContainer], use_gpu: bool) -> N
     if use_gpu:
         module.cuda()
 
-    assert_train_step(module, data_module, use_gpu)
-    assert_validation_step(module, data_module, use_gpu)
-    assert_test_step(module, data_module, use_gpu)
+    assert_train_step(module, data_module, use_gpu)  # type: ignore
```
dccastro, line 341: Why are we ignoring types here? Would this be fixed by narrowing the type annotation for `container_type` (e.g. `Type[BaseMILTiles]`)? Or at least declaring `module: BaseDeepMILModule`?

dccastro (commented):
Good job with all this refactoring—happy to see a lot duplication disappear! As you've seen, additional changes will be necessary to reconcile the conflicts with https://github.com/microsoft/hi-ml/pull/303.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -59,31 +63,22 @@ class BaseMIL(LightningContainer):
...
     encoding_chunk_size: int = param.Integer(0, doc="If > 0 performs encoding in chunks, by loading"
                                                     "enconding_chunk_size tiles per chunk")
     # local_dataset (used as data module root_path) is declared in DatasetParams superclass
 
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.best_checkpoint_filename = "checkpoint_max_val_auroc"
+        self.best_checkpoint_filename_with_suffix = self.best_checkpoint_filename + ".ckpt"
+        self.checkpoint_folder_path = "outputs/checkpoints/"
+        self.callbacks = self.get_monitoring_callbacks(monitor="val/auroc", mode="max")
```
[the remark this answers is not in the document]
kenza-bouzid: I thought that this might be useful in the future in case a child class needs some extra callbacks, or do you think that will probably not be the case?

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -229,3 +201,155 @@ def get_path_to_best_checkpoint(self) -> Path:
...
+    step: int = param.Integer(None, bounds=(0, None),
+                              doc="Step size to define the offset between tiles."
+                              "If None (default), it takes the same value as tile_size."
+                              "If step < tile_size, it creates overlapping tiles."
+                              "If step > tile_size, it skips some chunks in the wsi.")
+    random_offset: bool = param.Boolean(False, doc="If True, randomize position of the grid, instead of starting at"
+                                                   "the top-left corner,")
+    pad_full: bool = param.Boolean(False, doc="If True, pad image to the size evenly divisible by tile_size")
+    background_val: int = param.Integer(255, bounds=(0, None),
+                                        doc="The background constant to ignore background tiles.")
```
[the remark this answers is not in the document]
kenza-bouzid: Yes! I'll clarify it.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -229,3 +201,155 @@ def get_path_to_best_checkpoint(self) -> Path:
...
+    filter_mode: str = param.String("min", doc="mode must be in ['min', 'max', 'random']. If total number of tiles is"
+                                               "greater than tile_count, then sort by intensity sum, and take the "
+                                               "smallest (for min), largest (for max) or random (for random) subset, "
+                                               "defaults to 'min' (which assumes background is high value).")
+
+    def get_transform(self, image_key: str) -> Callable:
+        # TODO how to deal with intensity scaling: I added this transform because I was getting this error
```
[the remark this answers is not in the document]
kenza-bouzid: we actually had a quick chat with Valentina about this and she wasn't sure if this is normalized or not. Thanks for the clarification, I will simply cast it then. I have to double check if we can do that at the WSIReader level to keep it consistent.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -229,3 +201,155 @@ def get_path_to_best_checkpoint(self) -> Path:
...
+
+    def create_model(self) -> SlidesDeepMILModule:
+        self.setup_model_creation()
+        pooling_layer, num_features = self.get_pooling_layer()
+        deepmil_module = SlidesDeepMILModule(tiles_count=self.tile_count,
+                                             encoder=self.model_encoder,
+                                             # Here we can't use self.data_module.train_dataset.LABEL_COLUMN as labels.
+                                             # because label_column in (PANDA)SlidesDataset is "isup_grade" which is a
+                                             # string "ISUP_i" i={0,..,5}. However the model needs an integer label
+                                             # SlideKey.LABEL is used as an alternative but we should maybe cast
```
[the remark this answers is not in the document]
kenza-bouzid: Great, I'll use SlideKey.LABEL for tiles as well!

hi-ml-histopathology/src/histopathology/models/deepmil.py
kenza-bouzid: Definitely, I wasn't sure if I should only override get_bag_labels as we had in mind refactoring this loop in the next steps by vectorizing the calls. However, as @dccastro and Max mentioned earlier, this is probably not necessary unless we prove that it's a bottleneck. 

I will go for @dccastro suggestion for now then by just overriding get_bag_label(labels)

hi-ml-histopathology/testhisto/testhisto/mocks/slides_generator.py
kenza-bouzid: I added this "if tiles positioning is diagonal. Tiles are positioned randomly on the WSI grid when tiles positioning type is random." as clarification. Do you think I should just remove the illustration? It's helpful to better understand the purpose of other parameters.

hi-ml-histopathology/testhisto/testhisto/mocks/tiles_generator.py
kenza-bouzid: This is a bit subtle, cause self.n_tiles is used a bit differently depending on the type of tiles positioning... I should definitely use different parameters for clarity

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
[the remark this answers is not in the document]
dccastro: If a child class wants additional callbacks, can't they just do the following?
```python
def get_callbacks():
  extra_callbacks = [...]
  return super().get_callbacks() + extra_callbacks
```

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -229,3 +201,155 @@ def get_path_to_best_checkpoint(self) -> Path:
...
+                                            num_features=num_features,
+                                            dropout_rate=self.dropout_rate,
+                                            class_weights=self.data_module.class_weights,
+                                            l_rate=self.l_rate,
+                                            weight_decay=self.weight_decay,
+                                            adam_betas=self.adam_betas,
+                                            is_finetune=self.is_finetune,
+                                            class_names=self.class_names,
+                                            outputs_handler=self.get_outputs_handler())
+        assert deepmil_module.outputs_handler is not None
```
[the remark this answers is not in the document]
dccastro: Fair enough, though here it's indeed the same one.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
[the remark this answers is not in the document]
dccastro: I meant we should use `TileKey` in the tiles pipeline:
https://github.com/microsoft/hi-ml/blob/8722c8b574286f0a6eb182990742b2850512c762/hi-ml-histopathology/src/histopathology/utils/naming.py#L24-L37

hi-ml-histopathology/testhisto/testhisto/models/test_deepmil.py
dccastro: Because this test is in `test_deepmil.py`, I'd think it's safe to assume we'll only test `BaseMIL` containers. We can always generalise the test later if necessary.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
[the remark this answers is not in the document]
kenza-bouzid: I get this error, not sure what's happening under the hood in TilesDataset, I left it as before or does it worth investigating why "label" is missing?
```
>       for bag_idx in range(len(batch[self.label_column])):
E       KeyError: <TileKey.LABEL: 'label'>
```

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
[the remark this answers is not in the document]
kenza-bouzid: yes, agree! I wasn't sure about the difference between self.callbacks and self.get_callbacks() 
because in the previous implementation self.callbacks was ModelCheckpointCallback only and self.get_callbacks() returns a list of callbacks including super().get_callbacks().
Can you confirm that there is no need to set self.callbacks as ModelCheckpointCallback only? in the latest update I set self.callbacks = self.get_callbacks()

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -65,58 +64,35 @@ class BaseMIL(LightningContainer):
...
-                                     "(enables random subsampling of tiles).")
     # local_dataset (used as data module root_path) is declared in DatasetParams superclass
 
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         self.best_checkpoint_filename = "checkpoint_max_val_auroc"
         self.best_checkpoint_filename_with_suffix = self.best_checkpoint_filename + ".ckpt"
-        self.checkpoint_folder_path = "outputs/checkpoints/"
-        self.callbacks = self.get_monitoring_callbacks(monitor="val/auroc", mode="max")
+        self.checkpoint_folder_path = f"{self.outputs_folder}/checkpoints/"
```
dccastro, line 73: I believe this attribute is now used with different purposes. In `download_ssl_checkpoint()`, this needs to be as before (`"outputs/checkpoints/"`), because it's actually the _remote_ checkpoint location on AML, not a local path. And see my other suggestion for the checkpoint callback (using the `LightningContainer.checkpoint_folder` property instead). I suggest we get rid of `self.checkpoint_folder_path`.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -59,31 +63,22 @@ class BaseMIL(LightningContainer):
...
+        self.best_checkpoint_filename = "checkpoint_max_val_auroc"
+        self.best_checkpoint_filename_with_suffix = self.best_checkpoint_filename + ".ckpt"
+        self.checkpoint_folder_path = "outputs/checkpoints/"
+        self.callbacks = self.get_monitoring_callbacks(monitor="val/auroc", mode="max")
+
     @property
     def cache_dir(self) -> Path:
-        raise NotImplementedError
+        return Path(
+            f"/tmp/innereye_cache1/{self.__class__.__name__}-{self.encoder_type}/"
```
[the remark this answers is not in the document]
dccastro: Sorry, my comment was a bit misleading! I meant we could set it to another temporary subdirectory, e.g. `/tmp/himl_cache`.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -138,7 +154,16 @@ def get_pooling_layer(self) -> Tuple[nn.Module, int]:
...
+    def get_outputs_handler(self) -> DeepMILOutputsHandler:
+        return DeepMILOutputsHandler(outputs_root=self.outputs_folder,
+                                     n_classes=self.data_module.train_dataset.N_CLASSES,
+                                     tile_size=self.tile_size,
+                                     level=1,
+                                     class_names=self.class_names,
+                                     primary_val_metric=MetricsKey.AUROC,
+                                     maximise=True)
+
+    def setup_model_creation(self) -> None:
```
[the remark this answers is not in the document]
dccastro: Suggestion: rename this method to `def get_model_encoder(self) -> TileEncoder:` and return `model_encoder` instead of setting it as an attribute. It's only used in `create_model()`, and this makes the method's purpose clear without relying on side-effects.

hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -317,22 +312,11 @@ def __init__(self, tiles_count: int, **kwargs: Any) -> None:
...
-            bag_labels_list.append(label)  # no need to do majority voting to get the bag_label
-            logit, attn = self(images)
-            bag_logits_list.append(logit.view(-1))
-            bag_attn_list.append(attn)
-        bag_logits = torch.stack(bag_logits_list)
-        bag_labels = torch.stack(bag_labels_list).view(-1)
-        return bag_logits, bag_labels, bag_attn_list
+    @staticmethod
+    def get_bag_label(labels: Tensor) -> Tensor:
+        return labels
```
dccastro, line 317: Great! Minor suggestion to add a comment clarifying why here we return just `labels` for the slides pipeline.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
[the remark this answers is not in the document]
dccastro: Yes, I believe it was only ever used inside `BaseMIL.get_callbacks()`. I'm in favour of removing any unnecessary attributes :)

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -59,47 +62,36 @@ class BaseMIL(LightningContainer):
...
+        super().__init__(**kwargs)
+        self.best_checkpoint_filename = "checkpoint_max_val_auroc"
+        self.best_checkpoint_filename_with_suffix = self.best_checkpoint_filename + ".ckpt"
+        self.checkpoint_folder_path = f"{self.outputs_folder}/checkpoints/"
+        self.callbacks = self.get_callbacks()
+
     @property
     def cache_dir(self) -> Path:
-        raise NotImplementedError
+        return Path(f"himl_cache/{self.__class__.__name__}-{self.encoder_type}/")
```
ant0nsc, line 78: Do the cache results actually end up in the AML outputs? If so, we should find a better place for the cached data

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -146,75 +138,28 @@ def get_pooling_layer(self) -> Tuple[nn.Module, int]:
...
-            # We ensure num_devices is not 0 for non-GPU machines
-            # to avoid division by zero error when computing `workers_per_gpu`
-            num_devices = max(torch.cuda.device_count(), 1)
-            assert num_cpus is not None  # for mypy
-            workers_per_gpu = num_cpus // num_devices
-            dataloader_kwargs = dict(num_workers=workers_per_gpu, pin_memory=True)
-        return dataloader_kwargs
+    def setup_model_creation(self) -> None:
+        self.model_encoder = self.encoder
+        if self.is_finetune:
```
ant0nsc, line 152: is_finetune is not defined in this class, only in the subclasses.

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -240,3 +185,164 @@ def get_path_to_best_checkpoint(self) -> Path:
...
+        return None
+
+
+class BaseMILTiles(BaseMIL):
+    """BaseMILTiles is an abstract subclass of BaseMIL for running MIL experiments on tiles datasets. It is responsible
+    for instantiating the full DeepMIL model in tiles settings. Subclasses should define their datamodules and
+    configure experiment-specific parameters.
+    """
+    # Model parameters:
+    is_finetune: bool = param.Boolean(False, doc="If True, fine-tune the encoder during training. If False (default), "
```
ant0nsc, line 218: see comment above - you are using is_finetune already in BaseMIL

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -65,58 +64,35 @@ class BaseMIL(LightningContainer):
...
 
     def download_ssl_checkpoint(self, run_id: str) -> CheckpointDownloader:
         downloader = CheckpointDownloader(
             aml_workspace=get_workspace(),
             run_id=run_id,
-            checkpoint_filename="last.ckpt",
-            download_dir="outputs/",
+            checkpoint_filename=LAST_CHECKPOINT_FILE_NAME_WITH_SUFFIX,
+            download_dir=self.outputs_folder,
             remote_checkpoint_dir=Path(self.checkpoint_folder_path)
```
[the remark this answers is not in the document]
dccastro: [optional] With your other change below to `ModelCheckpoint.dirpath` in `get_callbacks()`, it looks like it would be safe to now just inline the value of `self.checkpoint_folder_path` here and get rid of the attribute. What do you think?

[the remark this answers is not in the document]
kenza-bouzid: What about get_path_to_best_checkpoint? Should I inline self.checkpoint_folder_path lines 173/179/184 as well?

dccastro (approved):

ant0nsc (approved):

PR #311, opened 2022-04-20
BUG: Fix function to compute bounding box

### LINKED ISSUE #310: Tiles offset in PANDA is incorrect
The absolute (level 0 of the slides) coordinates of the tiles generated by [`create_panda_tiles_dataset`](https://github.com/microsoft/hi-ml/blob/main/hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py#L79) are not correct. This are computed in

https://github.com/microsoft/hi-ml/blob/8722c8b574286f0a6eb182990742b2850512c762/hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py#L79

because the tile locations are stored as `x, y` (horizontal, vertical)

https://github.com/microsoft/hi-ml/blob/8722c8b574286f0a6eb182990742b2850512c762/hi-ml-histopathology/src/histopathology/preprocessing/tiling.py#L82

but the sample location seems to be stored as `y, x`, according to my debugging experiments. The loading code looks fine:

https://github.com/microsoft/hi-ml/blob/8722c8b574286f0a6eb182990742b2850512c762/hi-ml-histopathology/src/histopathology/datasets/panda_dataset.py#L118-L120

so I suspect the error is somewhere else. I thought it might come from the logic used to compute bounding boxes

https://github.com/microsoft/hi-ml/blob/8722c8b574286f0a6eb182990742b2850512c762/hi-ml-histopathology/src/histopathology/datasets/panda_dataset.py#L98-L112

so I looked at the test for this and found that 1) is not very comprehensive and 2) is incorrect as it passes the width and height to create an array, instead of height and width:

https://github.com/microsoft/hi-ml/blob/8722c8b574286f0a6eb182990742b2850512c762/hi-ml/testhiml/testhiml/test_box_utils.py#L15-L19

Passing

```python
array([[0, 1, 1, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0]])
```

to `get_bounding_box` results in

```python
(x_min, y_min) == (0, 1)
(width, height) == (1, 2)
```

but we wanted

```python
(x_min, y_min) == (1, 0)
(width, height) == (2, 1)
```

so the bug that needs to be fixed is in `get_bounding_box`!

fepegar:
Related to #310.


hi-ml/testhiml/testhiml/test_box_utils.py
```diff
@@ -8,27 +8,72 @@
...
     # passing an identity matrix will return a bounding box with the same shape as the original,
     # and xmin and ymin will both be zero
-    mask_eye = np.eye(length_x)
+    length = 5
+    mask_eye = np.eye(length)
     bbox_eye = get_bounding_box(mask_eye)
     assert isinstance(bbox_eye, Box)
-    assert bbox_eye.w == length_x
-    assert bbox_eye.h == length_x
+    assert bbox_eye.w == length
```
dccastro, line 36: [optional] As a `dataclass`, `Box` implements `__eq__()`. Might be a bit clearer and less verbose to instantiate `expected_box = Box(...)` and then `assert whatever_box == expected_box` (also in the other tests below).

dccastro (changes requested):

hi-ml/src/health_ml/utils/box_utils.py
```diff
@@ -123,14 +124,22 @@ def get_bounding_box(mask: np.ndarray) -> Box:
...
-    y_min, y_max = ys.min(), ys.max()
-    width = x_max - x_min + 1
-    height = y_max - y_min + 1
-    return Box(x_min, y_min, width, height)
+        raise TypeError(f"Expected a 2D array but got an array with shape {mask.shape}")
+
+    slices = ndimage.find_objects(mask > 0)
+    if not slices:
+        raise RuntimeError("The input mask is empty")
+    assert len(slices) == 1
```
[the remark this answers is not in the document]
dccastro: My bad. I was thinking that `ndimage.find_objects()` would return _connected components_, but the docs suggest "objects" are defined by the label values. So nevermind :)

dccastro (dismissed):
Good stuff; thanks @fepegar!

dccastro (approved):

hi-ml/src/health_ml/utils/box_utils.py
```diff
@@ -2,6 +2,7 @@
 from typing import Optional, Sequence, Tuple
```
harshita-s, line 2: Could you add the Microsoft header here?

hi-ml/testhiml/testhiml/test_box_utils.py
```diff
@@ -8,27 +8,78 @@
 from health_ml.utils.box_utils import Box, get_bounding_box
```
harshita-s, line 8: Same, please add the header with copyright info

hi-ml/src/health_ml/utils/box_utils.py
fepegar: If these files are missing the header, I suspect many others will too. Do you mind if we address that issue in a different PR?

harshita-s: I think we could address it where we spot, what do you think?

hi-ml/testhiml/testhiml/test_box_utils.py
fepegar: I think this file already has it.

hi-ml/src/health_ml/utils/box_utils.py
fepegar: Done.

harshita-s (approved):

dccastro (approved):

fepegar:
Thanks for your reviews!

PR #313, opened 2022-04-20
ENH: Add ability to override which checkpoint to use for inference

ant0nsc:
Models that use a custom logic for choosing best checkpoints can't presently use those for inference. This PR adds a method to the container to control that.

### DIFF
--- a/hi-ml/src/health_ml/deep_learning_config.py
+++ b/hi-ml/src/health_ml/deep_learning_config.py
@@ -127,12 +127,11 @@ class WorkflowParams(param.Parameterized):
     This class contains all parameters that affect how the whole training and testing workflow is executed.
     """
     random_seed: int = param.Integer(42, doc="The seed to use for all random number generators.")
-    weights_url: List[str] = param.List(default=[], class_=str,
-                                        doc="If provided, a set of urls from which checkpoints will be downloaded"
-                                            "and used for inference.")
-    local_weights_path: List[Path] = param.List(default=[], class_=Path,
-                                                doc="A list of checkpoints paths to use for inference, "
-                                                    "when the job is running outside Azure.")
+    weights_url: str = param.String(default="",
+                                    doc="If provided, use this URL to download checkpoints for inference.")
+    local_weights_path: Optional[Path] = \
+        param.ClassSelector(class_=Path, allow_None=True, default=None,
+                            doc="Checkpoint paths to use for inference, when the job is running outside Azure.")
     crossval_count: int = param.Integer(default=1, bounds=(0, None),
                                         doc="The number of splits to use when doing cross-validation. "
                                             "Use 1 to disable cross-validation")
@@ -158,7 +157,7 @@ class WorkflowParams(param.Parameterized):
     CROSSVAL_COUNT_ARG_NAME = "crossval_count"
 
     def validate(self) -> None:
-        if sum([bool(param) for param in [self.weights_url, self.local_weights_path]]) > 1:
+        if self.weights_url and self.local_weights_path:
             raise ValueError("Cannot specify more than one of local_weights_path, weights_url.")
 
         if self.crossval_count > 1:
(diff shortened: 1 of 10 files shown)


hi-ml/src/health_ml/lightning_container.py
```diff
@@ -135,6 +136,16 @@ def before_training_on_all_ranks(self) -> None:
         """
         pass
 
+    def get_checkpoint_to_test(self) -> Path:
+        """Returns the path of the model checkpoint that should be used for testing/inference. By default, this will
+        return the checkpoint that is written in the last training epoch. Override this method if you implement a
+        custom checkpointing logic, for example if you added a model checkpoint callback that looks at validation
+        accuracuy.
```
harshita-s, line 143: small typo in 'accuracy'

hi-ml/src/health_ml/utils/checkpoint_handler.py
```diff
@@ -0,0 +1,139 @@
...
+        if is_global_rank_zero():
+            checkpoints = list(self.container.checkpoint_folder.rglob("*"))
+            logging.info(f"Available checkpoints: {len(checkpoints)}")
+            for f in checkpoints:
+                logging.info(f)
+        return find_recovery_checkpoint_on_disk_or_cloud(self.container.checkpoint_folder)
+
+    def get_checkpoint_to_test(self) -> Path:
+        """
+        Find the model checkpoint that should be used for inference. If a run recovery is provided, or if the model
```
harshita-s, line 73: Checking if the run recovery case is included somewhere (earlier we had a case with recovery checkpoints `if self.run_recovery or self.has_continued_training`) If this isn't required now, perhaps the docs need to be updated?

ant0nsc: Thanks, updated.

hi-ml/src/health_ml/run_ml.py
```diff
@@ -175,15 +169,12 @@ def is_crossval_disabled_or_child_0(self) -> bool:
             return self.container.crossval_index == 0
         return True
 
-    def run_inference(self, checkpoint_paths: List[Path]) -> None:
+    def run_inference(self, checkpoint_path: Path) -> None:
         """
         Run inference on the test set for all models.
 
         :param checkpoint_paths: The path to the checkpoint that should be used for inference.
```
mebristo, line 176: Param name needs updating to checkpoint_path (singular)

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -128,15 +127,11 @@ class WorkflowParams(param.Parameterized):
     This class contains all parameters that affect how the whole training and testing workflow is executed.
     """
     random_seed: int = param.Integer(42, doc="The seed to use for all random number generators.")
-    weights_url: List[str] = param.List(default=[], class_=str,
-                                        doc="If provided, a set of urls from which checkpoints will be downloaded"
-                                            "and used for inference.")
-    local_weights_path: List[Path] = param.List(default=[], class_=Path,
-                                                doc="A list of checkpoints paths to use for inference, "
-                                                    "when the job is running outside Azure.")
-    model_id: str = param.String(default="",
```
mebristo, line 137: Why do we get rid of model_id in this PR?

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -20,8 +20,7 @@
 from health_ml.utils.common_utils import (CHECKPOINT_FOLDER,
                                           create_unique_timestamp_id,
                                           DEFAULT_AML_UPLOAD_DIR,
-                                          DEFAULT_LOGS_DIR_NAME,
-                                          parse_model_id_and_version)
```
mebristo, line 24: I think this file is the only place we were calling this function? If so, can we delete it now?

hi-ml/src/health_ml/deep_learning_config.py
ant0nsc: because it was not used.

harshita-s (approved):

PR #321, opened 2022-04-22
BUG: Fix code to create the PANDAS tiles dataset

### LINKED ISSUE #314: Default padding color is black when loading slides
When a WSI is loaded, the corresponding mask is used to load only the region of interest. If a `margin` > 0 is passed to `LoadPandaROId`, the slides are padded by MONAI's `WSIReader` using black as default color.

https://github.com/microsoft/hi-ml/blob/8722c8b574286f0a6eb182990742b2850512c762/hi-ml-histopathology/src/histopathology/datasets/panda_dataset.py#L110-L123

The background in WSIs is normally almost white, so padding with black might mess up the preprocessing.

fepegar:
Fixes #314 by setting `margin` to 0.
Fixes also #310, but I will close that once the dataset has been generated.
Related to https://github.com/Project-MONAI/MONAI/issues/4156.

The script to create the tiles dataset now uses more descriptive file names for the tiles, using `left`, `top`, `right` and `bottom` for the file names and the CSV columns, instead of `x` and `y`.


hi-ml-histopathology/src/histopathology/datasets/panda_dataset.py
```diff
@@ -115,9 +118,11 @@ def __call__(self, data: Dict) -> Dict:
         # but relative region size in pixels at the chosen level
         scale = mask_obj.resolutions['level_downsamples'][self.level]
         scaled_bbox = level0_bbox / scale
-        get_data_kwargs = dict(location=(level0_bbox.x, level0_bbox.y),
-                               size=(scaled_bbox.w, scaled_bbox.h),
-                               level=self.level)
+        get_data_kwargs = dict(
+            location=(level0_bbox.y, level0_bbox.x),
```
fepegar, line 122: Note the swapped x and y here!

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -60,109 +79,114 @@ def save_image(array_chw: np.ndarray, path: Path) -> PIL.Image:
...
+        TileKey.TILE_ID.value: tile_id,
+        TileKey.IMAGE.value: image_tile_filename,
+        TileKey.MASK.value: mask_tile_filename,
+        TileKey.TILE_LEFT.value: tile_box.x,
+        TileKey.TILE_TOP.value: tile_box.y,
+        TileKey.TILE_RIGHT.value: tile_box.x + tile_box.w,
+        TileKey.TILE_BOTTOM.value: tile_box.y + tile_box.h,
+        'data_provider': slide_metadata['data_provider'],
+        'slide_isup_grade': slide_metadata['isup_grade'],
+        'slide_gleason_score': slide_metadata['gleason_score'],
```
fepegar, line 136: Couldn't find Enums for these.

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -217,14 +246,69 @@ def main(panda_dir: Union[str, Path], root_output_dir: Union[str, Path], level:
...
+    parser.add_argument(
+        "--overwrite",
+        action="store_true",
+        default=False,
+    )
+    parser.add_argument(
+        "--filter-slide",  # process only this slide
+        type=str,
+        default="",
+    )
```
fepegar, line 300: This was useful for debugging; I can remove it if necessary.

dccastro: Should be fine to leave it in, IMO. I also had a similar param when developing this script.

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
dccastro: Correct; metadata fields are free-form and specific to each dataset. That said, they're listed in `PandaDataset.METADATA_COLUMNS`. Not sure it's worth using that here, though.

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -60,109 +79,114 @@ def save_image(array_chw: np.ndarray, path: Path) -> PIL.Image:
...
+
+        for i in tqdm(range(n_tiles), f"Tiles ({slide_id[:6]}…)", unit="img", disable=not tile_progress):
+            tile_metadata = save_tile(
+                sample,
+                image_tiles[i],
+                mask_tiles[i],
+                tile_boxes[i],
+                slide_dir,
+            )
+            tile_metadata['occupancy'] = occupancies[i]
```
dccastro, line 182: [optional] Should these keys be from `TileKey`?

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -60,109 +79,114 @@ def save_image(array_chw: np.ndarray, path: Path) -> PIL.Image:
...
+            tile_metadata['image'] = os.path.join(slide_dir.name, tile_metadata['image'])
```
dccastro, line 183: [optional] Just noticed this could be simplified to `slide_dir / tile_metadata['image']` (same below), if you felt like it.

hi-ml-histopathology/src/histopathology/utils/naming.py
```diff
@@ -32,6 +32,10 @@ class TileKey(str, Enum):
     SPLIT = 'split'
     TILE_X = 'tile_x'
     TILE_Y = 'tile_y'
+    TILE_LEFT = 'left'
```
dccastro, line 35: Do you envision us keeping the `TILE_X/Y` keys in addition to these new ones? Or should we refactor them out of existence later on?

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -217,14 +246,69 @@ def main(panda_dir: Union[str, Path], root_output_dir: Union[str, Path], level:
...
-         root_output_dir="/datadrive",
-         level=1,
-         tile_size=224,
-         margin=64,
-         occupancy_threshold=0.05,
-         parallel=True,
-         overwrite=False)
+    parser = ArgumentParser()
+    parser.add_argument(
+        "--container-dir",
```
dccastro, line 255: What do you mean by 'container' in this context? Why not pass the PANDA folder directly?

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -217,14 +246,69 @@ def main(panda_dir: Union[str, Path], root_output_dir: Union[str, Path], level:
...
+        type=str,
+        required=True,
+    )
+    parser.add_argument(
+        "--level",
+        type=int,
+        default=1,
+    )
+    parser.add_argument(
+        "--size",
```
dccastro, line 271: [optional] Maybe `--tile_size`, for added clarity and consistency?

hi-ml-histopathology/src/histopathology/datasets/panda_dataset.py
vale-salvatelli: Interesting we never realized we were swapping the coordinates - is it something we can test?

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -217,14 +246,69 @@ def main(panda_dir: Union[str, Path], root_output_dir: Union[str, Path], level:
...
+        type=int,
+        default=224,
+    )
+    parser.add_argument(
+        "--margin",
+        type=int,
+        default=64,
+    )
+    parser.add_argument(
+        "--occupancy",
```
dccastro, line 281: [optional] How about `--occupancy-threshold` or `--min-occupancy`?

dccastro (approved):
Looks good to merge, just added some suggestions and clarification questions. Was also wondering if you considered generating a small metadata file in the root output folder with the main tiling parameters (level, tile size, margin, occupancy threshold, etc.)?

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -60,109 +79,114 @@ def save_image(array_chw: np.ndarray, path: Path) -> PIL.Image:
...
-    abs_tile_locations = (sample['scale'] * tile_locations + sample['location']).astype(int)
+    top, left = sample['location']
+    offset = np.array((left, top))
+
+    abs_tile_locations = (sample['scale'] * tile_locations + offset).astype(int)
+    tile_size_scaled = int(tile_size * sample['scale'])
+    tile_boxes = [Box(x, y, tile_size_scaled, tile_size_scaled) for x, y in abs_tile_locations]
 
-    return image_tiles, mask_tiles, abs_tile_locations, occupancies, n_discarded
+    return image_tiles, mask_tiles, tile_boxes, occupancies, num_discarded
```
vale-salvatelli, line 109: shall we test this function? it contains a lot of logic around offset/scale

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -60,109 +79,114 @@ def save_image(array_chw: np.ndarray, path: Path) -> PIL.Image:
...
+        'slide_gleason_score': slide_metadata['gleason_score'],
     }
 
     return tile_metadata
 
 
 def process_slide(sample: dict, level: int, margin: int, tile_size: int, occupancy_threshold: int,
-                  output_dir: Path, tile_progress: bool = False) -> None:
-    slide_id = sample['image_id']
+                  output_dir: Path, tile_progress: bool = False, filter_slide: str = '') -> None:
```
vale-salvatelli, line 143: not sure I understand what filter_slide is used for? Should this be a list?

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -60,109 +79,114 @@ def save_image(array_chw: np.ndarray, path: Path) -> PIL.Image:
...
-            failed_tiles_file.close()
-            if tiles_failure > 0:
-                # TODO what we want to do with slides that have some failed tiles?
-                logging.warning(f"{slide_id} is incomplete. {tiles_failure} tiles failed.")
-        except Exception as e:
-            traceback.print_exc()
-            warnings.warn(f"An error occurred while processing slide {slide_id}: {e}")
+            shutil.rmtree(slide_dir)
+            return
+        logging.info(f"Tiling slide {slide_id} ...")
```
vale-salvatelli, line 169: The following block was in the try/except before. In the current version we assume the error can come only from loading the slide. If I remember correctly, we had errors also on saving. It might be safer to wrap also this block in a try/except.

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -217,14 +246,69 @@ def main(panda_dir: Union[str, Path], root_output_dir: Union[str, Path], level:
...
-         margin=64,
-         occupancy_threshold=0.05,
-         parallel=True,
-         overwrite=False)
+    parser = ArgumentParser()
+    parser.add_argument(
+        "--container-dir",
+        type=str,
+        required=True,
+        help="Folder with the 'datasets' container",
```
vale-salvatelli, line 258: I would provide an example

hi-ml-histopathology/src/histopathology/preprocessing/loading.py
```diff
@@ -99,12 +99,12 @@ def __call__(self, data: Dict) -> Dict:
...
         # cuCIM/OpenSlide takes absolute location coordinates in the level 0 reference frame,
         # but relative region size in pixels at the chosen level
-        origin = (level0_bbox.x, level0_bbox.y)
+        origin = (level0_bbox.y, level0_bbox.x)
         scale = image_obj.resolutions['level_downsamples'][self.level]
         scaled_bbox = level0_bbox / scale
 
         data[self.image_key], _ = self.reader.get_data(image_obj, location=origin, level=self.level,
-                                                       size=(scaled_bbox.w, scaled_bbox.h))
+                                                       size=(scaled_bbox.h, scaled_bbox.w))
```
vale-salvatelli, line 107: Now I understand why the axis were swapped before - why the need to swap axis in this transform? 

I couldn't find a test for this transform, this could be something worth to add as the transform was used in more than one place (e.g.  for the creation of TCGA-PRAD https://github.com/microsoft/hi-ml/blob/main/hi-ml-histopathology/src/histopathology/preprocessing/create_tiles_dataset.py)

hi-ml-histopathology/src/histopathology/utils/naming.py
```diff
@@ -32,6 +32,10 @@ class TileKey(str, Enum):
     SPLIT = 'split'
     TILE_X = 'tile_x'
     TILE_Y = 'tile_y'
```
vale-salvatelli, line 34: Maybe add a TODO to remove TILE_X. TILE_Y once all the datasets have been recreated according to this new philosophy

hi-ml-histopathology/src/histopathology/datasets/panda_dataset.py
fepegar: > Interesting we never realized we were swapping the coordinates - is it something we can test?

I guess we could write a test for this using a synthetic WSI. Can we write WSIs?

kenza-bouzid: Yep, you can use [MockPandaSlidesGenerator](https://github.com/microsoft/hi-ml/blob/39911d217c919d8213ad36c9c776f69369d98509/hi-ml-histopathology/testhisto/testhisto/mocks/slides_generator.py#L24)

fepegar: Thanks, @kenza-bouzid! I'll take a look.

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -40,15 +56,18 @@ def select_tile(mask_tile: np.ndarray, occupancy_threshold: float) \
         raise ValueError("Tile occupancy threshold must be between 0 and 1")
     foreground_mask = mask_tile > 0
     occupancy = foreground_mask.mean(axis=(-2, -1))
-    return (occupancy > occupancy_threshold).squeeze(), occupancy.squeeze()
+    selected = occupancy > occupancy_threshold
+    return selected[:, 0], occupancy[:, 0]
```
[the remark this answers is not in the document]
fepegar: Because squeezing removes all dimensions when only one tile is to be selected, which raises an error later on.

```python
In [1]: import numpy as np

In [2]: selected = np.zeros((128, 1))

In [3]: selected[:, 0].shape
Out[3]: (128,)

In [4]: selected.squeeze().shape
Out[4]: (128,)

In [5]: selected = np.zeros((1, 1))

In [6]: selected[:, 0].shape
Out[6]: (1,)

In [7]: selected.squeeze().shape
Out[7]: ()
```

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
```diff
@@ -40,15 +56,18 @@ def select_tile(mask_tile: np.ndarray, occupancy_threshold: float) \
...
 
 
-def get_tile_descriptor(tile_location: Sequence[int]) -> str:
-    return f"{tile_location[0]:05d}x_{tile_location[1]:05d}y"
+def get_tile_descriptor(tile_box: Box) -> str:
+    left, top = tile_box.x, tile_box.y
+    right, bottom = left + tile_box.w, top + tile_box.h
+    return f"left_{left:05d}_top_{top:05d}_right_{right:05d}_bottom_{bottom:05d}"
```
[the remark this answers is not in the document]
fepegar: Yes.

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
fepegar: The idea of this PR was to write a quick fix for PANDAS. It turned out to be not that quick due to the bugs. A similar function is used in the more generic script `create_tiles_dataset`. Discussing with @dccastro, we thought we could revisit and merge these scripts if we ever needed to generate more tiles datasets, which might be unlikely as we're moving towards online cropping. Then, we would move this logic outside of the script and write tests to cover it. Does that make sense?

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
fepegar: I used it for debugging. It's useful to generate a dataset of a single slide. I'll add a help message.

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
fepegar: I meant "folder on which an Azure blob container is mounted". I can't remember why I switched to specifying the container, but I'll revert back.

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
fepegar: I got no errors on saving. They might have come from the issue with `squeeze`. I have narrowed down the try/except blocks to identify why errors happen and properly decide what to do with them.

hi-ml-histopathology/src/histopathology/preprocessing/loading.py
harshita-s: Not sure if anything changes in https://github.com/microsoft/hi-ml/blob/main/hi-ml-histopathology/src/histopathology/preprocessing/create_tiles_dataset.py ? This was used for tiling other slide datasets

hi-ml-histopathology/src/histopathology/datasets/panda_dataset.py
fepegar: I would prioritise merging this in case it fixes the issue reported by @harshita-s, then we can think of adding some tests if you like. Do you agree?

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
fepegar: Not really because `slide_dir` is an absolute path. But I've made some edits to avoid using `os`. Thanks!

hi-ml-histopathology/src/histopathology/preprocessing/loading.py
fepegar: > Now I understand why the axis were swapped before - why the need to swap axis in this transform?

Because the reader expects (y, x) coordinates.

> I couldn't find a test for this transform, this could be something worth to add as the transform was used in more than one place (e.g. for the creation of TCGA-PRAD https://github.com/microsoft/hi-ml/blob/main/hi-ml-histopathology/src/histopathology/preprocessing/create_tiles_dataset.py)

As I said above, I would wait to see if we need to create more tiles datasets before we add tests. Maybe we can add an issue to the backlog.

hi-ml-histopathology/src/histopathology/utils/naming.py
fepegar: As @vale-salvatelli, I think we should deprecate them at some point.

hi-ml-histopathology/src/histopathology/preprocessing/loading.py
fepegar: @harshita-s I'm not sure what you mean; I haven't touched that file.

fepegar: Some more context about the coordinate swapping:
- https://github.com/microsoft/hi-ml/issues/310

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
dccastro: Ah yes, this is to produce the relative paths that go in the CSV

fepegar:
> Looks good to merge, just added some suggestions and clarification questions. Was also wondering if you considered generating a small metadata file in the root output folder with the main tiling parameters (level, tile size, margin, occupancy threshold, etc.)?

Sorry, I had missed this comment. I think none of these are actually needed for our purposes. The main concern was the tile size, but that is now explicit thanks to the bounding boxes being included in the CSV.
But I've added a log file in which all this is recorded anyway. It starts by mentioning the command used to generate the dataset, so in theory that should be enough to recover this information, if ever needed.
What do you think?

```console
$ cat /tmp/PANDA_tiles/panda_tiles_level1_224/2022-04-25_09-43-49.log     
2022-04-25 09:43:49 fernando-comp-westeu root[33137] INFO Command: "/home/azureuser/git/RadiomicsNN/innereye-deeplearning/hi-ml/hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py --panda-dir /home/azureuser/datasets/panda --output-dir /tmp/PANDA_tiles --overwrite --no-parallel --margin 0 --filter-slide 0005f7aaab2800f6170c399693a96917"
2022-04-25 09:43:49 fernando-comp-westeu root[33137] INFO Creating dataset of level-1 224x224 PANDA tiles at: /tmp/PANDA_tiles/panda_tiles_level1_224
2022-04-25 09:43:49 fernando-comp-westeu root[33137] INFO >>> Slide dir /tmp/PANDA_tiles/panda_tiles_level1_224/0005f7aaab2800f6170c399693a96917
2022-04-25 09:43:49 fernando-comp-westeu root[33137] INFO Loading slide 0005f7aaab2800f6170c399693a96917 ...
2022-04-25 09:43:50 fernando-comp-westeu root[33137] INFO Tiling slide 0005f7aaab2800f6170c399693a96917 ...
2022-04-25 09:43:50 fernando-comp-westeu root[33137] INFO Discarded 107/168 tiles (63.69 %)
[...]
```

hi-ml-histopathology/src/histopathology/preprocessing/loading.py
harshita-s: > @harshita-s I'm not sure what you mean; I haven't touched that file.

Was just wondering if we need to change that file as well (Daniel and Valentina suggested that it should be changed but this will be addressed in a later PR)

fepegar: Ok, thanks for clarifying!

dccastro: @fepegar, I think @harshita-s is wondering whether your changes in this PR affect the other script (e.g. via `LoadROId`, as @vale-salvatelli mentioned). We may need to revisit `create_tiles_dataset.py` somewhat soon e.g. for TCGA-PRAD. I'd be in favour of merging this now and discussing separately how to harmonise the scripts going forward.

dccastro:
> Sorry, I had missed this comment. I think none of these are actually needed for our purposes. The main concern was the tile size, but that is now explicit thanks to the bounding boxes being included in the CSV. But I've added a log file in which all this is recorded anyway. It starts by mentioning the command used to generate the dataset, so in theory that should be enough to recover this information, if ever needed. What do you think?
That's great; should be enough in case we need that info later.

hi-ml-histopathology/src/histopathology/preprocessing/loading.py
fepegar: Sounds good!

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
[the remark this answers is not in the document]
vale-salvatelli: fair enough

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
vale-salvatelli: as long as you can create the full dataset without errors looks OK to me - separate try/except would have been OK to capture different errors

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
vale-salvatelli: I would still be in favor of writing a test just for this - if something is still wrong in the logic there will be errors in the recreated tiles. I am OK to wait for tests that concerns future datasets

hi-ml-histopathology/src/histopathology/preprocessing/loading.py
vale-salvatelli: I am OK with getting in delaying this test if @dccastro is going to refactor and look into TCGA-PRAD soon (before we forget about this change). Adding a warning or a TODO in `create_tiles_dataset` would not harm

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
fepegar: I've updated the code to make error catching more granular. There are 4 slides with an empty mask and around 100 slides without an associated mask. All those slides are currently being ignored. For these cases, we could resort to computing the mask ourselves, but that might be out of the scope of this PR.

hi-ml-histopathology/src/histopathology/preprocessing/create_panda_tiles_dataset.py
fepegar: I've added a test.

vale-salvatelli (approved):

fepegar:
Thank you all for reviewing!

PR #322, opened 2022-04-26
STYLE: Rename panda_tiles to panda_tiles_small

mebristo:
The AML dataset panda_tiles (which is a subset of 61 PANDA_tiles files) will be renamed to panda_tiles_small, in order not to confuse with the full-size PANDA_tiles dataset


kenza-bouzid (approved):
Looks good!

fepegar (approved):
LGTM.

For my education, why is there an `other` folder? I've seen some duplicated files with respect to the rest of the library.

PR #327, opened 2022-05-04
Add code to generate large slides montage

fepegar:
As discussed with @dccastro and @harshita-s, this is a draft PR to quickly share the code I used to generate the large slides montages/mosaics (we need to find a name for them), in case new images are needed this week.

To do:

- Remove hard-coded variables and logic
- Add some tests, maybe using the new tiny datasets we have for testing
- Add docstrings
- Move to base dataset

Feel free to push to this branch if you make some changes.


fepegar:
It looks like I created this branch off `310-fix-panda-tiles-dataset` instead of `main`. I will fix all this next week :)

dccastro:
Thanks for implementing this functionality, @fepegar! Could we perhaps move the thumbnail and montage stuff to a separate module as standalone functions, so we can easily reuse across datasets? Also not convinced it makes sense for them to be part of the dataset class itself. What do you think?

fepegar:
Closing this for now. We can reopen at some point if needed.

fepegar:
(Thanks for taking a look anyway, @dccastro!)

PR #333, opened 2022-05-06
ENH: Add template for multimodal package

ant0nsc:
* Add a folder and Python namespace for multimodal work
* Add a simple PR workflow
* Add an empty template folder "new_project_template" from which we can start future projects quicker
* Move GPU test runner from hi-ml to hi-ml-azure and ensure it only depends on minimal set of packages


.github/workflows/multimodal-pr.yml
```diff
@@ -0,0 +1,178 @@
...
+          cd ${{ env.folder }}
+
+          # Install local package in editable mode
+          make pip_local
+
+          # Run tests
+          make pip_test
+          make pytest
+
+      #- name: Run GPU tests
```
mebristo, line 127: Delete commented lines?

.github/workflows/multimodal-pr.yml
```diff
@@ -0,0 +1,178 @@
...
+          cd ${{ env.folder }}
+          make pip_build
+          make build
+
+      - name: Upload package artifacts
+        uses: ./.github/actions/upload_package_artifacts
+        with:
+          folder: ${{ env.folder }}
+
+  publish-pypi-pkg:
```
mebristo, line 158: will multimodal by published as a package to PyPI?

new_project_template/.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Setup hi-ml-multimodal package",
```
mebristo, line 8: Should the name of this be something more generic like 'setup package'?

new_project_template/Makefile
```diff
@@ -0,0 +1,59 @@
...
+	python setup.py sdist bdist_wheel
+
+# run flake8, assuming test requirements already installed
+flake8:
+	flake8 --count --statistics .
+
+# run mypy, assuming test requirements already installed
+mypy:
+	mypy --install-types --non-interactive setup.py
+	mypy --install-types --non-interactive -p health_multimodal
```
mebristo, line 43: Again I think these 2 lines could be removed since they refer to multimodal?

new_project_template/README.md
```diff
@@ -0,0 +1,21 @@
...
+This is a folder with a template set of files to create a new project in hi-ml.
+
+To use it:
+
+* Copy the template folder to the desired project folder
+* In the new location, rename the folders `health_multimodal` and `test_multimodal` to be the desired Python namespaces
+* Search for all occurences of `multimodal` and replace with the desired project name. Don't use a simple query/replace
+  strategy but do check the context!
+* Move the workflow definition `new-project-pr.yml` to `.github/workflows` and give it a name that matches the project.
+
```
mebristo, line 12: Should we mention something about updating the *_requirements.txt files if necessary?

mebristo (changes requested):
I've left a few comments about replacing references to 'multimodal' in the new package template with something more generic since I think it could cause confusion and maybe even errors down the road if all references aren't properly removed.. Let me know your thoughts. Ps I haven't commented all the instances, just the first few I came across, to avoid repeating the same comment

multimodal/package_description.md
```diff
@@ -0,0 +1,28 @@
...
+You can install the latest version from `pypi` via
+
+```shell
+pip install hi-ml-multimodal
+```
+
+## Documentation
+
+The detailed package documentation, with examples and API reference, is on
+[readthedocs](https://hi-ml.readthedocs.io/en/latest/).
```
Shruthi42, line 18: Should we point to project specific links on Read The Docs? I see one for hi-ml-histopathology (https://hi-ml.readthedocs.io/en/latest/histopathology.html) but not for hi-ml.

Also, not too important now, but how do we want to point to older versions of our docs, given that each package will move independently? Maybe we can come up with some nomenclature to make it clear what package each release/version based on.

multimodal/setup.py
```diff
@@ -0,0 +1,88 @@
...
+    if build_number:
+        # In github workflows, we may pull in hi-ml-multimodal as a dependency. Usually, we have a condition like
+        # hi-ml-multimodal>=0.1.5. This means that a package version from PyPi would trump the local wheels. For this
+        # reason, use an extremely large version number to give the local wheel priority.
+        version = '99.99.post' + build_number
+    else:
+        default_random_version_number = floor(random() * 10_000_000_000)
+        version = f'99.99.post{str(default_random_version_number)}'
+
+package_name = 'hi-ml-multimodal'
```
Shruthi42, line 56: I'm wondering, why do we call the package hi-ml-multimodal but not the project folder?

multimodal/package_description.md
ant0nsc: Yes, this should point to something specific - but for now, these are all just templates that need to be filled out later once we have actual docs.

multimodal/setup.py
ant0nsc: Happy to talk about that - we can unify that. I just wanted to have something that is nicer with folder name autocomplete. If everything has a "hi-ml-" prefix, there's lots of extra typing to be done.

new_project_template/README.md
ant0nsc: yup.

.github/workflows/multimodal-pr.yml
ant0nsc: I've removed publishing for now.

new_project_template/Makefile
ant0nsc: should be clearer now with the "newproject" suffix.

multimodal/setup.py
```diff
@@ -0,0 +1,88 @@
...
+# If this was triggered by a tagged commit then GITHUB_REF will be: 'ref/tags/new_tag'.
+# Extract this tag and use it as a version string
+# See also:
+# https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+# https://github.com/pypa/gh-action-pypi-publish
+GITHUB_REF_TAG_COMMIT = 'refs/tags/v'
+
+github_ref = os.getenv('GITHUB_REF')
+if github_ref and github_ref.startswith(GITHUB_REF_TAG_COMMIT):
+    version = github_ref[len(GITHUB_REF_TAG_COMMIT):]
```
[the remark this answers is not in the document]
ant0nsc: yes, that's indeed the case. It's a minor issue, in my view. Also Pypi has some additional logic to not accept identical packages, afaik.

multimodal/environment.yml
```diff
@@ -0,0 +1,19 @@
...
+  - pip:
+    # Run requirements for hi-ml-azure
+    - azureml-sdk==1.36.0
+    - azureml-tensorboard==1.36.0
+    - conda-merge==0.1.5
+    - pandas==1.3.4
+    - param==1.12
+    - pysocks==1.6.0
+    - ruamel.yaml==0.16.12
+    - tensorboard==2.6.0
```
Shruthi42, line 17: Could we remove these dependencies for now? We've been refactoring the code to make sure it has as few dependencies as possible. Right now, the only dependencies should be einops, torch and transformers (torch only, not tensorflow).
We don't have any dependency on azureml yet, so maybe we could wait to add this and other dependencies in the future?

ant0nsc: removed now because the GPU tests are gone.

multimodal/environment.yml
```diff
@@ -0,0 +1,10 @@
+name: multimodal
+channels:
+  - defaults
+  - pytorch
+dependencies:
+  - pip=20.1.1
+  - python=3.7.3
+  - pip:
+    # Test requirements
+    - pytest==6.2.2
```
Shruthi42, line 10: This is out of sync with the test requirements, don't we need mypy and Flake8 as well? One thought is that we could directly use the requirements.txt file here, I believe something like this should work:

```suggestion
  - pip:
    # Test requirements
    - -r requirements_test.txt
```

hi-ml-azure/run_pytest.py
```diff
@@ -46,6 +53,10 @@ class RunPytestConfig(param.Parameterized):
         default="30m", doc="The maximum runtime that is allowed for this job in AzureML. This is given as a floating"
         "point number with a string suffix s, m, h, d for seconds, minutes, hours, day. Examples: '3.5h', '2d'"
     )
+    sys_path: str = param.String(
```
mebristo, line 56: I think the name of this suggests that it will replace the entire sys.path rather than be appended to it?

multimodal/environment.yml
ant0nsc: updated and left comments in the file for reference.

mebristo (approved):

.github/workflows/multimodal-pr.yml
```diff
@@ -0,0 +1,125 @@
...
+          # Install local package in editable mode
+          make pip_local
+
+      - name: Install hi-ml in editable mode
+        run: |
+          cd hi-ml
+
+          # Install local package in editable mode
+          make pip_local
+
```
Shruthi42, line 115: Shouldn't we also remove these two steps for now, since we've also removed hi-ml-azure and hi-ml from the dependency list?

Shruthi42 (approved):

PR #337, opened 2022-05-11
ENH: Validate the number of requested nodes for an AML job

mebristo:
Add function that checks the number of nodes that have been specified for an AML job and checks that number is less than or equal to the maximum number of nodes allowed for the cluster.

Also small refactoring to separate out another cluster validation step into its own function. Tests added for all of the above.

### DIFF
--- a/hi-ml-azure/src/health_azure/himl.py
+++ b/hi-ml-azure/src/health_azure/himl.py
@@ -86,7 +86,7 @@ def validate_num_nodes(compute_cluster: ComputeTarget, num_nodes: int) -> None:
     Check that the user hasn't requested more nodes than the maximum number of nodes allowed by
     their compute cluster
 
-    :param compute_cluster_name: An AML ComputeTarget representing the cluster whose upper node limit
+    :param compute_cluster: An AML ComputeTarget representing the cluster whose upper node limit
         should be checked
     :param num_nodes: The number of nodes that the user has requested
     """
--- a/hi-ml-azure/src/health_azure/himl.py
+++ b/hi-ml-azure/src/health_azure/himl.py
@@ -81,33 +81,33 @@ class AzureRunInfo:
     folder will be uploaded to blob storage regularly during the script run."""
 
 
-def validate_num_nodes(existing_compute_clusters: Dict[str, ComputeTarget], compute_cluster_name: str, num_nodes: int
-                       ) -> None:
+def validate_num_nodes(compute_cluster: ComputeTarget, num_nodes: int) -> None:
     """
-    Check that the user hasn't requested more nodes than the maximum number of nodes allowed by their compute cluster
+    Check that the user hasn't requested more nodes than the maximum number of nodes allowed by
+    their compute cluster
 
-    :param existing_compute_clusters: A list of AML ComputeTarget objects in a given AML Workspace
-    :param compute_cluster_name: The name of the specific compute cluster whose upper node limit should be checked
+    :param compute_cluster_name: An AML ComputeTarget representing the cluster whose upper node limit
+        should be checked
     :param num_nodes: The number of nodes that the user has requested
     """
-    compute_cluster = existing_compute_clusters[compute_cluster_name]
     max_cluster_nodes: int = compute_cluster.scale_settings.maximum_node_count
     if num_nodes > max_cluster_nodes:
         raise ValueError(
-            f"You have requested {num_nodes} nodes, which is more than your compute cluster's maximum of "
-            f"{max_cluster_nodes} nodes.")
+            f"You have requested {num_nodes} nodes, which is more than your compute cluster "
+            f"({compute_cluster.name})'s maximum of {max_cluster_nodes} nodes.")
 
 
-def validate_compute_name(existing_compute_clusters: Dict[str, ComputeTarget], compute_cluster_name: str) -> None:
+def validate_compute_name(existing_compute_targets: Dict[str, ComputeTarget], compute_target_name: str) -> None:
     """
-    Check that the specified compute cluster is one of the available existing compute clusters in the Workspace
+    Check that a specified compute target is one of the available existing compute targets in a Workspace
 
-    :param existing_compute_clusters: A list of AML ComputeTarget objects in a given AML Workspace
-    :param compute_cluster_name: The name of the specific compute cluster whose name to lookup in existing clusters
+    :param existing_compute_targets: A list of AML ComputeTarget objects available to a given AML Workspace
+    :param compute_cluster_name: The name of the specific compute target whose name to look up in existing
+        compute targets
     """
-    if compute_cluster_name not in existing_compute_clusters:
-        raise ValueError(f"Could not find the compute target {compute_cluster_name} in the AzureML workspace. ",
-                         f"Existing clusters: {list(existing_compute_clusters.keys())}")
+    if compute_target_name not in existing_compute_targets:
+        raise ValueError(f"Could not find the compute target {compute_target_name} in the AzureML workspace. ",
+                         f"Existing compute targets: {list(existing_compute_targets)}")
 
 
 def validate_compute_cluster(workspace: Workspace, compute_cluster_name: str, num_nodes: int) -> None:
@@ -121,7 +121,8 @@ def validate_compute_cluster(workspace: Workspace, compute_cluster_name: str, nu
     """
     existing_compute_clusters: Dict[str, ComputeTarget] = workspace.compute_targets
     validate_compute_name(existing_compute_clusters, compute_cluster_name)
-    validate_num_nodes(existing_compute_clusters, compute_cluster_name, num_nodes)
+    compute_cluster = existing_compute_clusters[compute_cluster_name]
+    validate_num_nodes(compute_cluster, num_nodes)
 
 
 def create_run_configuration(workspace: Workspace,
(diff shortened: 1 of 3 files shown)


mebristo:
Failing test is unrelated to this PR

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -81,6 +81,49 @@ class AzureRunInfo:
...
+def validate_num_nodes(existing_compute_clusters: Dict[str, ComputeTarget], compute_cluster_name: str, num_nodes: int
+                       ) -> None:
+    """
+    Check that the user hasn't requested more nodes than the maximum number of nodes allowed by their compute cluster
+
+    :param existing_compute_clusters: A list of AML ComputeTarget objects in a given AML Workspace
+    :param compute_cluster_name: The name of the specific compute cluster whose upper node limit should be checked
+    :param num_nodes: The number of nodes that the user has requested
+    """
+    compute_cluster = existing_compute_clusters[compute_cluster_name]
```
ant0nsc, line 93: The parameters for this function could be simplified. You are passing in a dictionary and a key, but you use it only to do the lookup. Would be more natural to just pass in the compute cluster.

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -119,6 +120,103 @@ def test_write_run_recovery_file(mock_run: mock.MagicMock) -> None:
     assert expected_run_recovery_id != recovery_file_text
 
 
+@pytest.fixture(scope="module")
+def dummy_max_num_nodes_available() -> int:
+    return randint(2, 10)
+
+
+@pytest.fixture(scope="module")
+def dummy_compute_cluster_name() -> str:
```
ant0nsc, line 129: I don't get the point of using a constant in a fixture.

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -119,6 +120,103 @@ def test_write_run_recovery_file(mock_run: mock.MagicMock) -> None:
...
+    return 'dummy_cluster'
+
+
+@pytest.fixture(scope="module")
+def mock_scale_settings(dummy_max_num_nodes_available: int) -> MagicMock:
+    return MagicMock(maximum_node_count=dummy_max_num_nodes_available)
```
ant0nsc, line 135: Maybe add comments here and below what you are trying to mock here? (like: Mocking a ComputeCluster object which has a property maximum_node_count...)

ant0nsc (approved):

ant0nsc:
@mebristo , one more high-level thought around tests here: What if the structure of the AML cluster object changed, retiring a property for example? Would we catch that in the tests? Maybe it would be good to, in a first test, work with an actual Cluster object, and read out the property. Check that the validation works as expected with the real object. Then later mock only the property that holds the number of nodes if you need to do more detailed checks.

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -119,6 +120,103 @@ def test_write_run_recovery_file(mock_run: mock.MagicMock) -> None:
...
+    return MagicMock(compute_targets={dummy_compute_cluster_name: mock_compute_cluster})
+
+
+@pytest.mark.fast
+def test_validate_num_nodes(dummy_max_num_nodes_available: int, dummy_compute_cluster_name: str,
+                            mock_workspace: MagicMock) -> None:
+    # If number of requested nodes <= max available nodes, nothing should happen
+    num_nodes_requested = dummy_max_num_nodes_available // 2
+    existing_compute_clusters = mock_workspace.compute_targets
+    print(f"Existing compute cluster scale settings: {existing_compute_clusters['dummy_cluster'].scale_settings}")
```
fepegar, line 154: Do we need print statements in tests?

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -119,6 +120,103 @@ def test_write_run_recovery_file(mock_run: mock.MagicMock) -> None:
...
+    himl.validate_num_nodes(existing_compute_clusters, dummy_compute_cluster_name, num_nodes_requested)
+    num_nodes_requested = dummy_max_num_nodes_available
+    himl.validate_num_nodes(existing_compute_clusters, dummy_compute_cluster_name, num_nodes_requested)
+
+    # But if number of nodes requested > max available, a ValueError should be raised
+    num_nodes_requested = randint(dummy_max_num_nodes_available, 5000)
```
fepegar, line 160: Isn't it possible that `num_nodes_requested == dummy_max_num_nodes_available` here? In that case, the error wouldn't be raised and the test would fail. Right?

fepegar: IIUC, we should use `dummy_max_num_nodes_available + 1` instead.

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -119,6 +120,103 @@ def test_write_run_recovery_file(mock_run: mock.MagicMock) -> None:
...
+    with pytest.raises(ValueError) as e:
+        himl.validate_num_nodes(existing_compute_clusters, dummy_compute_cluster_name, num_nodes_requested)
+    expected_error_msg = f"You have requested {num_nodes_requested} nodes, which is more than your compute cluster's"
+    f"maximum of {dummy_max_num_nodes_available} nodes"
+    assert expected_error_msg in str(e)
```
fepegar, line 165: Have you considered using the `match` kwarg of `pytest.raises` instead?

fepegar (changes requested):

hi-ml-azure/testazure/testazure/test_himl.py
ant0nsc: I see no issues with those

hi-ml-azure/testazure/testazure/test_himl.py
ant0nsc: nice one, I did not know about that!

fepegar: Neither did I! I found it googling around. I wasn't very comfortable hard-coding messages from somewhere else in the codebase and tried to find an alternative method, but I guess there's no other way to do it.

mebristo: I don't understand how this helps? We still have to provide the message, but with match we specify it inside the call to `raises`, rather than outside?

hi-ml-azure/testazure/testazure/test_himl.py
mebristo: Because it is shared between multiple tests

hi-ml-azure/testazure/testazure/test_himl.py
mebristo: Yes, good catch!

hi-ml-azure/testazure/testazure/test_himl.py
fepegar: I think `match` is generally more powerful as it uses regex (it calls `re.search`). But it's not critical here.

hi-ml-azure/testazure/testazure/test_himl.py
mebristo: Happy to remove, probably not a very useful statement anyway

hi-ml-azure/testazure/testazure/test_himl.py
mebristo: I see, thanks. Happy to update (for the chance to try out a feature I didn't know about!)

fepegar (approved):

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -81,33 +81,33 @@ class AzureRunInfo:
...
-                       ) -> None:
+def validate_num_nodes(compute_cluster: ComputeTarget, num_nodes: int) -> None:
     """
-    Check that the user hasn't requested more nodes than the maximum number of nodes allowed by their compute cluster
+    Check that the user hasn't requested more nodes than the maximum number of nodes allowed by
+    their compute cluster
 
-    :param existing_compute_clusters: A list of AML ComputeTarget objects in a given AML Workspace
-    :param compute_cluster_name: The name of the specific compute cluster whose upper node limit should be checked
+    :param compute_cluster_name: An AML ComputeTarget representing the cluster whose upper node limit
```
ant0nsc, line 89: argument name is no longer in sync

hi-ml-azure/testazure/testazure/test_himl.py
ant0nsc: yes, but why make it a fixture, and not a plain constant in that file?

ant0nsc (approved):

hi-ml-azure/testazure/testazure/test_himl.py
mebristo: It was just for consistency. Happy to make this a constant if you prefer?

mebristo:
> @mebristo , one more high-level thought around tests here: What if the structure of the AML cluster object changed, retiring a property for example? Would we catch that in the tests? Maybe it would be good to, in a first test, work with an actual Cluster object, and read out the property. Check that the validation works as expected with the real object. Then later mock only the property that holds the number of nodes if you need to do more detailed checks.

Good idea - I've added a test for this

mebristo:
> Can you add a link?

https://github.com/microsoft/hi-ml/pull/340

PR #355, opened 2022-05-18
ENH: Add outputs for visualization of attention heatmaps on DSA

harshita-s:
In this PR, columns for tile coordinates (left, top, right, bottom) are added to `test_output.csv` to enable the visualization of attention heatmaps overlay with WSI on DSA.


hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -308,7 +308,15 @@ def update_results_with_data_specific_info(self, batch: dict, results: dict) ->
                         ResultsKey.TILE_ID: batch[TilesDataset.TILE_ID_COLUMN],
                         ResultsKey.IMAGE_PATH: batch[TilesDataset.PATH_COLUMN]})
 
-        if (TilesDataset.TILE_X_COLUMN in batch.keys()) and (TilesDataset.TILE_Y_COLUMN in batch.keys()):
+        if (TileKey.TILE_TOP in batch.keys()) and (TileKey.TILE_LEFT in batch.keys()) \
```
dccastro, line 311: [Optional] A couple of less verbose alternatives (for here and in `metrics_utils.py`):
```python
# Test if set of keys is subset of all batch keys
if {TileKey.TILE_TOP, TileKey.TILE_LEFT, TileKey.TILE_RIGHT, TileKey.TILE_BOTTOM} <= batch.keys():
```
or more explicitly:
```python
if all(key in batch for key in [TileKey.TILE_TOP, TileKey.TILE_LEFT, TileKey.TILE_RIGHT, TileKey.TILE_BOTTOM]):
```

hi-ml-histopathology/src/histopathology/utils/metrics_utils.py
```diff
@@ -155,8 +155,14 @@ def plot_heatmap_overlay(slide: str,
 
     # for each tile in the bag
     for tile_idx in range(len(results[ResultsKey.IMAGE_PATH][slide_idx])):
-        tile_coords = np.transpose(np.array([results[ResultsKey.TILE_X][slide_idx][tile_idx].cpu().numpy(),
-                                             results[ResultsKey.TILE_Y][slide_idx][tile_idx].cpu().numpy()]))
+        if (ResultsKey.TILE_TOP in results.keys()) and (ResultsKey.TILE_LEFT in results.keys()) \
+                and (ResultsKey.TILE_RIGHT in results.keys()) and (ResultsKey.TILE_BOTTOM in results.keys()):
+            tile_coords = np.transpose(np.array([results[ResultsKey.TILE_LEFT][slide_idx][tile_idx].cpu().numpy(),
```
dccastro, line 160: [Optional] I think calling `np.stack([x, y], axis=1)` would make the intent and output shape clearer than `np.transpose(np.array([x, y]))` (also below and in `test_metrics_utils.py`).

hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -308,7 +308,15 @@ def update_results_with_data_specific_info(self, batch: dict, results: dict) ->
...
+                and (TileKey.TILE_RIGHT in batch.keys()) and (TileKey.TILE_BOTTOM in batch.keys()):
+            results.update({ResultsKey.TILE_TOP: batch[TileKey.TILE_TOP],
+                            ResultsKey.TILE_LEFT: batch[TileKey.TILE_LEFT],
+                            ResultsKey.TILE_RIGHT: batch[TileKey.TILE_RIGHT],
+                            ResultsKey.TILE_BOTTOM: batch[TileKey.TILE_BOTTOM]}
+                           )
+        # the condition below ensures compatibility with older tile datasets
+        elif (TilesDataset.TILE_X_COLUMN in batch.keys()) and (TilesDataset.TILE_Y_COLUMN in batch.keys()):
```
dccastro, line 319: In this case, could we not do instead `ResultsKey.TILE_LEFT/TOP: batch[TilesDataset.TILE_X/Y_COLUMN]`? Then we may be able to deprecate/delete `ResultsKey.TILE_X/Y` entirely, remove the `if/else` in `metrics_utils.py` an `test_metrics_utils.py`, and just check for the existence of `ResultsKey.TILE_RIGHT/BOTTOM` where necessary to get the tile size. @fepegar, do you anticipate any issues with this?

hi-ml-histopathology/src/histopathology/utils/metrics_utils.py
fepegar: Optional: There's quite a lot repetition and nested indexing, which makes this code a bit hard to read and understand. For example, x and y could be defined in the if/else block and stacking could happen after.

hi-ml-histopathology/src/histopathology/models/deepmil.py
fepegar: That sounds like a good idea, but it's hard for me to predict issues happening in other parts of the codebase.

hi-ml-histopathology/testhisto/testhisto/utils/test_metrics_utils.py
```diff
@@ -88,7 +88,27 @@ def assert_equal_lists(pred: List, expected: List) -> None:
                  [Tensor([200, 424, 200, 424]),
                   Tensor([200, 200, 424, 424]),
                   Tensor([200, 200, 424, 424]),
-                  Tensor([200, 200, 424, 424])]
+                  Tensor([200, 200, 424, 424])],
+             ResultsKey.TILE_LEFT:
```
fepegar, line 92: Can't we now get rid of the previous TILE_X and TILE_Y results in this test?

hi-ml-histopathology/testhisto/testhisto/utils/test_metrics_utils.py
```diff
@@ -226,9 +246,14 @@ def test_location_selected_tiles(level: int) -> None:
...
-            np.array([test_dict[ResultsKey.TILE_X][slide_idx][tile_idx].cpu().numpy(),  # type: ignore
-                      test_dict[ResultsKey.TILE_Y][slide_idx][tile_idx].cpu().numpy()]))  # type: ignore
+        if (ResultsKey.TILE_TOP in test_dict.keys()) and (ResultsKey.TILE_LEFT in test_dict.keys()) \
+                and (ResultsKey.TILE_RIGHT in test_dict.keys()) and (ResultsKey.TILE_BOTTOM in test_dict.keys()):
+            tile_coords = np.transpose(np.array([test_dict[ResultsKey.TILE_LEFT][slide_idx][tile_idx].cpu().numpy(),
+                                                 test_dict[ResultsKey.TILE_TOP][slide_idx][tile_idx].cpu().numpy()]))
+        else:
+            tile_coords = np.transpose(
+                np.array([test_dict[ResultsKey.TILE_X][slide_idx][tile_idx].cpu().numpy(),  # type: ignore
+                          test_dict[ResultsKey.TILE_Y][slide_idx][tile_idx].cpu().numpy()]))  # type: ignore
```
fepegar, line 256: I think our comments about previous snippets apply here as well.

hi-ml-histopathology/src/histopathology/models/deepmil.py
fepegar: There's also `a.issubset(b)`, which is, I think, more explicit than `a <= b` for sets.

hi-ml-histopathology/src/histopathology/models/deepmil.py
harshita-s: @dccastro how should we deal with the `else` when co-ordinates are not available? We will still have to check if `TilesDataset.TILE_X/Y_COLUMN` are available for the dataset, so there will be an if-else

harshita-s: @dccastro thinking more about this, we still have default columns as `TilesDataset.TILE_X/Y_COLUMN` and we need to check for co-ordinates in the previously created datasets not having LEFT, TOP etc. , so we need the `elif`. Agreed that we can delete `ResultsKey.TILE_X/Y` for consistent outputs

hi-ml-histopathology/testhisto/testhisto/utils/test_metrics_utils.py
harshita-s: Thanks, done

hi-ml-histopathology/src/histopathology/utils/metrics_utils.py
harshita-s: Thanks, the nesting (condition) is now removed as we don't have `ResultsKey.TILE_X/Y` now, as suggested

fepegar (approved):

hi-ml-histopathology/src/histopathology/models/deepmil.py
dccastro: Yes, I agree we still need this `elif` here for the column names; my point was regarding the `ResultsKey`s, which you've already addressed. 🙂

dccastro (approved):

PR #361, opened 2022-05-19
ENH: Add pl_limit_batches to TrainerConfigs.

kenza-bouzid:
Enable pytorch lightning flag "fast_dev_run" in our pipeline to be able to limit the number of train/test/val batches at once for fast debugging.

### DIFF
--- a/hi-ml/src/health_ml/deep_learning_config.py
+++ b/hi-ml/src/health_ml/deep_learning_config.py
@@ -397,6 +397,10 @@ class TrainerParams(param.Parameterized):
         param.Number(default=None,
                      doc="PyTorch Lightning trainer flag 'limit_test_batches': Limit the test dataset to the "
                          "given number of batches if integer, or proportion of test dataset if float.")
+    pl_limit_batches: Optional[IntOrFloat] = \
+        param.Number(default=None,
+                     doc="PyTorch Lightning trainer flag 'fast_dev_run': Limit the test dataset to the "
+                         "given number of batches if integer, or proportion of test dataset if float.")
     pl_profiler: Optional[str] = \
         param.String(default=None,
                      doc="The value to use for the 'profiler' argument for the Lightning trainer. "
--- a/hi-ml/src/health_ml/deep_learning_config.py
+++ b/hi-ml/src/health_ml/deep_learning_config.py
@@ -21,7 +21,7 @@ from health_ml.utils.common_utils import (CHECKPOINT_FOLDER,
                                           create_unique_timestamp_id,
                                           DEFAULT_AML_UPLOAD_DIR,
                                           DEFAULT_LOGS_DIR_NAME)
-from health_ml.utils.type_annotations import IntOrFloat, TupleFloat2
+from health_ml.utils.type_annotations import IntOrFloat, TupleFloat2, IntOrBool
 
 
 @unique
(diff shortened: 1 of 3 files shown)


hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -397,6 +397,10 @@ class TrainerParams(param.Parameterized):
         param.Number(default=None,
                      doc="PyTorch Lightning trainer flag 'limit_test_batches': Limit the test dataset to the "
                          "given number of batches if integer, or proportion of test dataset if float.")
+    pl_limit_batches: Optional[IntOrBool] = \
+        param.Number(default=False,
+                     doc="PyTorch Lightning trainer flag 'fast_dev_run': Runs n if set to ``n`` (int) else 1 if set to"
+                         "``True`` batch(es) of train, val and test to" "find any bugs (ie: a sort of unit test).")
```
mebristo, line 403: I see that this matches the description in PL's Trainer class, but I wonder if we can make it clearer by explaining that the default is False - i.e. use all batches? Also I find their statement 'to find any bugs (ie: a sort of unit test).' unnecessary - what do you think?

kenza-bouzid: Sounds great! I'll update that.

mebristo:
Great spot that we can use this parameter to simplify our command line args. Since we aren't removing the component params (e.g. `--pl_limi_train_batches` etc, I wonder what is the behaviour if `--pl_limit_train_batches` (e.g.) and `--fast_dev_run` are specified with different values? Do we want to keep all 4 of these params? If so, can you add some tests to determine which parameter overrides the other? (In fact adding tests for this new param is a good idea in general!)

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -397,6 +397,12 @@ class TrainerParams(param.Parameterized):
         param.Number(default=None,
                      doc="PyTorch Lightning trainer flag 'limit_test_batches': Limit the test dataset to the "
                          "given number of batches if integer, or proportion of test dataset if float.")
+    pl_limit_batches: Optional[IntOrBool] = \
```
ant0nsc, line 400: IntOrBool is not in sync with param.Number. I'd use IntOrFloat

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -397,6 +397,12 @@ class TrainerParams(param.Parameterized):
...
+        param.Number(default=False,
+                     doc="PyTorch Lightning trainer flag 'fast_dev_run': Runs n if set to ``n`` (int) else 1 if set to"
```
ant0nsc, line 402: I would leave out the ability to set that to True/False - param.Number will prevent that anyway.

hi-ml/src/health_ml/utils/type_annotations.py
```diff
@@ -7,6 +7,7 @@
 
 T = TypeVar('T')
 IntOrFloat = Union[int, float]
+IntOrBool = Union[int, bool]
```
ant0nsc, line 10: I think this is no longer needed then

ant0nsc (approved):

hi-ml/src/health_ml/deep_learning_config.py
kenza-bouzid: but Pytorch Lightning trainer expects fast_dev_run to be Union[int, bool]. Should we maybe use just Int and set defaults to 0? a boolean is just 0 or 1 after all

mebristo (approved):

PR #365, opened 2022-05-20
ENH: Run WSI pipeline without a fixed tile_count 

kenza-bouzid:
Main changes:
* Enable training with tile_count = None.
* Rename save_tiles parameter to save_output_tiles.
*  Move train/val/test_dataloader fucntions to base class HistoDataModule


hi-ml-histopathology/src/histopathology/utils/wsi_utils.py
```diff
@@ -3,7 +3,7 @@
 
 from typing import List
 from histopathology.utils.naming import SlideKey
-from torch.utils.data.dataloader import default_collate
+from health_ml.utils.bag_utils import multibag_collate
 
 
 def image_collate(batch: List[dict]) -> List[dict]:
```
harshita-s, line 9: do we want to have a test for `image_collate` as codecov has some warnings

kenza-bouzid: Yes, just discussed that on Teams with Daniel. I will add a test for that.

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -85,6 +85,15 @@ def get_splits(self) -> Tuple[_SlidesOrTilesDataset, _SlidesOrTilesDataset, _Sli
...
         raise NotImplementedError
 
+    def train_dataloader(self) -> DataLoader:
+        return self._get_dataloader(self.train_dataset, shuffle=True, stage=ModelKey.TRAIN, **self.dataloader_kwargs)
+
+    def val_dataloader(self) -> DataLoader:
+        return self._get_dataloader(self.val_dataset, shuffle=True, stage=ModelKey.VAL, **self.dataloader_kwargs)
+
+    def test_dataloader(self) -> DataLoader:
+        return self._get_dataloader(self.test_dataset, shuffle=True, stage=ModelKey.TEST, **self.dataloader_kwargs)
```
vale-salvatelli, line 95: Now that I look at these - I think it's important we shuffle the data before splitting in train/test/validation but for reproducibility we should not randomly shuffling val and test - thoughts?

hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -331,17 +331,18 @@ def get_bag_label(labels: Tensor) -> Tensor:
 
     def update_results_with_data_specific_info(self, batch: dict, results: dict) -> None:
         # WARNING: This is a dummy input until we figure out tiles coordinates retrieval in the next iteration.
+        tile_counts = [tiles.shape[0] for tiles in batch[SlideKey.IMAGE]]
```
vale-salvatelli, line 334: why do we expect a different number of tiles per slides now?

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
kenza-bouzid: Yes definitely! I wanted to point it out as well!

hi-ml-histopathology/src/histopathology/models/deepmil.py
kenza-bouzid: For inference, if we want to use all tiles in a slide it might differ from a slide to another. -> The whole purpose of this PR

vale-salvatelli:
> > Question: Is `tile_count` same as `max_bag_size` or `max_bag_size_inf`? Why do we have two different params now if these represent the same thing?
> 
> Great question, we can maybe remove tile_count to stay consistent in both pipelines. Now that we are able to run the pipeline with all tiles, we can maybe change line 298 in base_module.py to: `tile_count=self.max_bag_size if stage == ModelKey.TRAIN else self.max_bag_size_inf` in the TileInGridd Transform. @vale-salvatelli what do you think?

I agree these parameters are the same. @kenza-bouzid what you suggest sounds good to me

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
kenza-bouzid: we should also maybe consider setting random_offset to False for val/test. what do you think?

vale-salvatelli: shall we change? @harshita-s would you agree?

hi-ml-histopathology/src/histopathology/models/deepmil.py
vale-salvatelli: I see! Are some tests capturing this change? (we now want all outputs NOT to assume all slides have the same number of tiles)

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
harshita-s: Sure we can change it, I saw a warning on AML as well

hi-ml-histopathology/src/histopathology/models/deepmil.py
kenza-bouzid: I am writing a test for the collate function, what else should be tested in your opinion? maybe add a test in test_slides_datamodule?

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
vale-salvatelli: "we should also maybe consider setting random_offset to False for val/test. what do you think?" -> yes, good point

PR #367, opened 2022-05-20
ENH: Add log text reader function

peterhessey:
Adds a util function for reading the text contained in log files. Designed to work for boht `std_log.txt` (new runtime) and `70_driver_log.txt`. Used only once so far in hi-ml but has use in other tasks in InnerEye, such as [here](https://github.com/microsoft/InnerEye-Inference/issues/35).

### DIFF
--- a/docs/source/examples.md
+++ b/docs/source/examples.md
@@ -44,7 +44,7 @@ python sample.py -n 103 --azureml
 then the function `submit_to_azure_if_needed` will perform all the required actions to run this script in AzureML and exit. Note that:
 
 * code after `submit_to_azure_if_needed` is not run locally, but it is run in AzureML.
-* the print statement prints to the AzureML console output and is available in the `Output + logs` tab of the experiment in the `70_driver_log.txt` file, and can be downloaded from there.
+* the print statement prints to the AzureML console output and is available in the `Output + logs` tab of the experiment in the `70_driver_log.txt` if using the old AzureML runtime, or `std_log.txt` if using the new AzureML runtime, and can be downloaded from there.
 * the command line arguments are passed through (apart from --azureml) when running in AzureML.
 * a new file: `most_recent_run.txt` will be created containing an identifier of this AzureML run.
 
--- a/hi-ml-azure/src/health_azure/utils.py
+++ b/hi-ml-azure/src/health_azure/utils.py
@@ -76,6 +76,7 @@ CONDA_CHANNELS = "channels"
 CONDA_DEPENDENCIES = "dependencies"
 CONDA_PIP = "pip"
 
+VALID_LOG_FILE_PATHS = [Path("user_logs/std_log.txt"), Path("azureml-logs/70_driver_log.txt")]
 
 # By default, define several environment variables that work around known issues in the software stack
 DEFAULT_ENVIRONMENT_VARIABLES = {
@@ -1459,31 +1460,32 @@ def download_files_from_run_id(
     torch_barrier()
 
 
-def get_driver_log_file_text(run: Run) -> Optional[str]:
-    """Given a run object, return the text stored in the driver log file of that run. Detects log files produced
+def get_driver_log_file_text(run: Run, download_file=True) -> Optional[str]:
+    """
+    Return the text stored in the driver log file of that run. Detects log files produced
     by both old and new runtimes.
 
     :param run: Run object representing the current run.
-    :return: Driver log file text if a file exists, None otherwise.
+    :param download_file: Boolean flag that control if the log file should be downloaded from the run or not.
+    :return: Driver log file text if a file exists, `None` otherwise.
     """
     with tempfile.TemporaryDirectory() as tmp_dir_name:
-        run.download_files(
-            prefix="azureml-logs",
-            output_directory=tmp_dir_name,
-            append_prefix=False
-        )
-        run.download_files(
-            prefix="user_logs",
-            output_directory=tmp_dir_name,
-            append_prefix=False
-        )
 
-        for log_file_name in ["70_driver_log.txt", "std_log.txt"]:
-            driver_log_path = Path(tmp_dir_name) / log_file_name
-            if driver_log_path.exists():
-                return driver_log_path.read_text()
+        for log_file_path in VALID_LOG_FILE_PATHS:
+            if download_file:
+                run.download_files(
+                    prefix=str(log_file_path),
+                    output_directory=tmp_dir_name,
+                    append_prefix=False,
+                )
+            if log_file_path.is_file():
+                return log_file_path.read_text()
 
-    logging.warning("Tried to get driver log file text when no such file exists.")
+    files_as_str = ', '.join([f"'{str(log_file_path)}'" for log_file_path in VALID_LOG_FILE_PATHS])
+    logging.warning(
+        "Tried to get driver log file for run {run.id} text when no such file exists. Expected to find "
+        f"one of the following: {files_as_str}"
+    )
     return None
 
 
--- a/hi-ml-azure/src/health_azure/utils.py
+++ b/hi-ml-azure/src/health_azure/utils.py
@@ -1466,7 +1466,7 @@ def get_driver_log_file_text(run: Run, download_file: bool = True) -> Optional[s
 
     :param run: Run object representing the current run.
     :param download_file: If ``True``, download log file from the run.
-    :return: Driver log file text if a file exists, `None` otherwise.
+    :return: Driver log file text if a file exists, ``None`` otherwise.
     """
     with tempfile.TemporaryDirectory() as tmp_dir_name:
 
@@ -1481,7 +1481,7 @@ def get_driver_log_file_text(run: Run, download_file: bool = True) -> Optional[s
             if tmp_log_file_path.is_file():
                 return tmp_log_file_path.read_text()
 
-    files_as_str = ', '.join([f"'{str(log_file_path)}'" for log_file_path in VALID_LOG_FILE_PATHS])
+    files_as_str = ', '.join(f"'{log_file_path}'" for log_file_path in VALID_LOG_FILE_PATHS)
     logging.warning(
         "Tried to get driver log file for run {run.id} text when no such file exists. Expected to find "
         f"one of the following: {files_as_str}"
--- a/hi-ml-azure/src/health_azure/utils.py
+++ b/hi-ml-azure/src/health_azure/utils.py
@@ -1459,6 +1459,27 @@ def download_files_from_run_id(
     torch_barrier()
 
 
+def get_driver_log_file_text(run: Run) -> Optional[str]:
+    """Given a run object, return the text stored in the driver log file of that run. Detects log files produced 
+    by both old and new runtimes.
+
+    :param run: Run object representing the current run.
+    :return: Driver log file text if a file exists, None otherwise.
+    """
+    with tempfile.TemporaryDirectory() as tmp_dir_name:
+        run.download_files(prefix="azureml-logs", output_directory=tmp_dir_name,
+                            append_prefix=False)
+        run.download_files(prefix="user_logs", output_directory=tmp_dir_name,
+                            append_prefix=False)
+
+        for log_file_name in ["70_driver_log.txt", "std_log.txt"]:
+            driver_log_path = Path(tmp_dir_name) / log_file_name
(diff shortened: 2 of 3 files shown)


hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1459,6 +1459,27 @@ def download_files_from_run_id(
     torch_barrier()
 
 
+def get_driver_log_file_text(run: Run) -> Optional[str]:
+    """Given a run object, return the text stored in the driver log file of that run. Detects log files produced 
+    by both old and new runtimes.
```
fepegar: Maybe first four words can be removed, as their redundant given the function signature.
Also, I think the summary of the docstring should not span multiple lines (PEP8). Clarifications can come below.

peterhessey:
Also re. unit tests: mocking `Run` objects properly seems like a daunting task somewhat out of the scope of this issue. Fortunately this function has been tested in end-to-end tests using both runtimes (AML seems to pick at random which it's going to use) and tests pass. However, the test case where no log file is present is _not_ covered by these tests.

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1459,6 +1459,27 @@ def download_files_from_run_id(
...
+
+    :param run: Run object representing the current run.
+    :return: Driver log file text if a file exists, None otherwise.
```
fepegar, line 1467: You could surround "None" with double backticks to improve Sphinx rendering.

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1459,6 +1459,27 @@ def download_files_from_run_id(
...
+    """
+    with tempfile.TemporaryDirectory() as tmp_dir_name:
+        run.download_files(prefix="azureml-logs", output_directory=tmp_dir_name,
+                            append_prefix=False)
+        run.download_files(prefix="user_logs", output_directory=tmp_dir_name,
+                            append_prefix=False)
+
+        for log_file_name in ["70_driver_log.txt", "std_log.txt"]:
```
fepegar, line 1481: Are these values defined somewhere else? It would be nice to not hard-code them.

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1459,6 +1459,27 @@ def download_files_from_run_id(
...
+            driver_log_path = Path(tmp_dir_name) / log_file_name
+            if driver_log_path.exists():
```
fepegar, line 1483: I suggest `is_file` instead of `exists`.

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1459,6 +1459,27 @@ def download_files_from_run_id(
...
+                return driver_log_path.read_text()
+
+    logging.warning("Tried to get driver log file text when no such file exists.")
```
fepegar, line 1486: It would be nice to have the run ID and the expected file path in this message.

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -875,20 +876,13 @@ def spawn() -> Tuple[int, List[str]]:
...
-        for f in driver_log_files:
-            if f in files:
-                run.download_file(f, output_file_path=str(driver_log))
-                break
-        else:
-            raise ValueError("The run does not contain any of the driver log files")
-        log_text = driver_log.read_text()
-        return log_text
+
+        # TODO: upgrade to walrus operator when upgrading python version to 3.8+
```
fepegar, line 880: ❤️

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -875,20 +876,13 @@ def spawn() -> Tuple[int, List[str]]:
...
+        # if log_text := get_driver_log_file_text(run=run):
+        log_text = get_driver_log_file_text(run=run)
+        if log_text:
```
fepegar, line 883: Not sure about this, but is it maybe better to explicitly say `if log_text is None`?

fepegar: I think it is. Imagine the log exists, but it's blank. This would still unexpectedly fail.

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -875,20 +876,13 @@ def spawn() -> Tuple[int, List[str]]:
...
+            return log_text
+        raise ValueError("The run does not contain any of the driver log files")
```
fepegar, line 885: As above, maybe adding more info in the message would be nice.

fepegar (changes requested):

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1459,6 +1459,34 @@ def download_files_from_run_id(
...
 
+def get_driver_log_file_text(run: Run) -> Optional[str]:
+    """Given a run object, return the text stored in the driver log file of that run. Detects log files produced
+    by both old and new runtimes.
+
+    :param run: Run object representing the current run.
+    :return: Driver log file text if a file exists, None otherwise.
+    """
+    with tempfile.TemporaryDirectory() as tmp_dir_name:
+        run.download_files(
```
ant0nsc, line 1470: I would not download the whole azureml-logs/user_logs folders, they also contain potentially large files, restrict to only the single file we need.

ant0nsc (changes requested):
We have 2 occurences of 70_driver_log.txt in code, and one in documentation. Could you clean up those as part of this PR?

hi-ml-azure/src/health_azure/utils.py
ant0nsc: or check the render_and_run_test_script function for an alternative way of seeing which files are available

hi-ml-azure/src/health_azure/utils.py
peterhessey: I agree with the suggestion to remove the first 4 words. However I'm not clear on why the summary should not span multiple lines? There are many instances of summaries over multiple lines in this file.

hi-ml-azure/src/health_azure/utils.py
peterhessey: Unfortunately not, I've defined them as a constant at the top of the file now though

hi-ml-azure/src/health_azure/utils.py
peterhessey: The second occurrence of `70_driver_log.txt` is in `docs/source/examples/2/results.py` which imports `health_azure` as a PyPI package - it will needs to be changed in a separate PR after this one is released in a new PyPI package.

hi-ml-azure/src/health_azure/utils.py
fepegar: > I'm not clear on why the summary should not span multiple lines

It's recommended by [PEP 257](https://peps.python.org/pep-0257/#multi-line-docstrings) (not PEP 8, as I said before):

> **[Multi-line Docstrings](https://peps.python.org/pep-0257/#multi-line-docstrings)**
> 
> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line. The summary line may be on the same line as the opening quotes or on the next line. The entire docstring is indented the same as the quotes at its first line (see example below).
> Insert a blank line after all docstrings (one-line or multi-line) that document a class – generally speaking, the class’s methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line.

> There are many instances of summaries over multiple lines in this file

There are many issues with docstrings across our repos, I think it's good to start working on improving them or, at least, making sure we don't create more.

peterhessey: Gotcha, will put a more concise summary now!

docs/source/examples.md
```diff
@@ -44,7 +44,7 @@ python sample.py -n 103 --azureml
 then the function `submit_to_azure_if_needed` will perform all the required actions to run this script in AzureML and exit. Note that:
 
 * code after `submit_to_azure_if_needed` is not run locally, but it is run in AzureML.
-* the print statement prints to the AzureML console output and is available in the `Output + logs` tab of the experiment in the `70_driver_log.txt` file, and can be downloaded from there.
+* the print statement prints to the AzureML console output and is available in the `Output + logs` tab of the experiment in the `70_driver_log.txt` if using the old AzureML runtime, or `std_log.txt` if using the new AzureML runtime, and can be downloaded from there.
```
fepegar, line 47: I'm curious. How does one know which runtime is using?

peterhessey: So far the only way I've been able to tell is a notification at the top of the page while the run is running.

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1459,6 +1460,35 @@ def download_files_from_run_id(
     torch_barrier()
 
 
+def get_driver_log_file_text(run: Run, download_file: bool = True) -> Optional[str]:
+    """
+    Returns text stored in run log driver file.
+
+    :param run: Run object representing the current run.
+    :param download_file: Boolean flag that control if the log file should be downloaded from the run or not.
```
fepegar, line 1468: Suggestion: remove "Boolean flag to control". I'd just say maybe "If True, download the log file from the run.", where True must be surrounded by _two_ backticks.

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1459,6 +1460,35 @@ def download_files_from_run_id(
...
+    :return: Driver log file text if a file exists, `None` otherwise.
```
fepegar, line 1469: Double backticks must be used, not single. See https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1459,6 +1460,35 @@ def download_files_from_run_id(
...
+                run.download_files(
+                    prefix=str(log_file_path),
+                    output_directory=tmp_dir_name,
+                    append_prefix=False,
+                )
+            tmp_log_file_path = tmp_dir_name / log_file_path
+            if tmp_log_file_path.is_file():
+                return tmp_log_file_path.read_text()
+
+    files_as_str = ', '.join([f"'{str(log_file_path)}'" for log_file_path in VALID_LOG_FILE_PATHS])
```
fepegar, line 1484: 1. You don't need to use `str` in the f-string, the conversion will happen automatically.
2. You can lose the square brackets (some linter will raise errors about this)

hi-ml-azure/src/health_azure/utils.py
peterhessey: Why two backticks?

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -875,20 +877,21 @@ def spawn() -> Tuple[int, List[str]]:
...
+
+        # TODO: upgrade to walrus operator when upgrading python version to 3.8+
+        # if log_text := get_driver_log_file_text(run=run):
+        log_text = get_driver_log_file_text(run=run)
+        if log_text is not None:
+            return log_text
+
+        raise ValueError(
+            "The run does not contain any of the following log files: "
+            f"{[str(log_file_path) for log_file_path in VALID_LOG_FILE_PATHS]}"
```
fepegar, line 893: As above.

fepegar, line 893: Style suggestion for readability, although it can be argued.

```python
if log_text is None:
    raise ValueError(
        "The run does not contain any of the following log files: "
        f"{[str(log_file_path) for log_file_path in VALID_LOG_FILE_PATHS]}"
    )
return log_text
```

fepegar (changes requested):

hi-ml-azure/src/health_azure/utils.py
fepegar: Left a reference in my comment below.

docs/source/examples.md
fepegar: Thanks! @ant0nsc, do you know more about this?

fepegar (approved):

ant0nsc (approved):

PR #348, opened 2022-05-17
BUG: Make files with shebang executable

fepegar:
Picked up by `pre-commit`.

PR #347, opened 2022-05-17
STYLE: Ensure all files end with a newline

fepegar:
Picked up by `pre-commit`.

Issue #342, opened 2022-05-13
Add linters and flake8 as pre-commit hooks


ant0nsc:
mypy, flake8, pylint?

Pre-commit hooks should be easy to install on each machine

ant0nsc:
Closed by #371

PR #298, opened 2022-04-10
DOC: Extend coding guidelines

ant0nsc:

PR #371, opened 2022-05-23
ENH: Add pre-commit config file

fepegar:
Related to:
- #298
- #342
- #345
- #346
- #347
- #348
- #349

TODO:
- [x] Add `pre-commit` to requirements
- [ ] Add installation instructions in documentation, maybe in #298?
- [ ] Add `pre-commit` bot: https://pre-commit.ci. I will ask Microsoft to approve the app first

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,7 +20,7 @@ repos:
   - id: check-shebang-scripts-are-executable  # Checks that scripts with shebangs are executable
 
 - repo: https://github.com/PyCQA/flake8
-  rev: 3.9.2
+  rev: 4.0.1
   hooks:
   - id: flake8
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,30 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+# See https://github.com/pre-commit/pre-commit-hooks/blob/master/.pre-commit-config.yaml for an example with more hooks
+
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v4.2.0
+  hooks:
+  - id: check-ast                  # Simply check whether the files parse as valid python
+  - id: check-yaml                 # Attempts to load all yaml files to verify syntax
+
+  - id: check-added-large-files    # Prevent giant files from being committed
+  - id: check-docstring-first      # Checks a common error of defining a docstring after code
+  - id: check-merge-conflict       # Check for files that contain merge conflict strings
+  - id: check-vcs-permalinks       # Ensures that links to vcs websites are permalinks
+  - id: detect-private-key         # Detects the presence of private keys
+  - id: end-of-file-fixer          # Ensures that a file is either empty, or ends with one newline
+  - id: trailing-whitespace        # This hook trims trailing whitespace
+  - id: mixed-line-ending          # Replaces or checks mixed line ending
+  - id: check-shebang-scripts-are-executable  # Checks that scripts with shebangs are executable
+
+- repo: https://github.com/PyCQA/flake8
+  rev: 3.9.2
+  hooks:
+  - id: flake8
+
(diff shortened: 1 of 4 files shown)


fepegar:
Where do you think the requirement should be specified?

ant0nsc:
> Where do you think the requirement should be specified?

sorry, I don't get your question?

fepegar:
> sorry, I don't get your question?

The `pre-commit` library must be pip-installed for development, but I'm not sure where the requirement (e.g., `pre-commit==2.19.0`) should be specified. I mean, in which environment YAML file or requirements txt file within this repository I should add that line. Does that make sense?

ant0nsc:
> The `pre-commit` library must be pip-installed for development, but I'm not sure where the requirement (e.g., `pre-commit==2.19.0`) should be specified. I mean, in which environment YAML file or requirements txt file within this repository I should add that line. Does that make sense?

Got it. Unfortunately in multiple places, root level `test_requirements.txt`, and `requirements_test.txt` and `environment.yml` in histo folder.

vale-salvatelli:
Each package has its own flake8/mypy.ini, therefore each package should have their own pre-commit config file. 

Related to that, make sure the config points to our own defined flake8.ini to avoid inconsistencies.

vale-salvatelli:
Each package has its own flake8/mypy.ini, therefore each package should have their own pre-commit config file. 

Related to that, make sure the config points to our own defined flake8.ini to avoid inconsistencies.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+- repo: https://github.com/PyCQA/flake8
+  rev: 4.0.1
+  hooks:
+  - id: flake8
+
+- repo: https://github.com/pre-commit/mirrors-autopep8
+  rev: v1.5.7
+  hooks:
+  - id: autopep8
```
vale-salvatelli, line 30: why not adding also a hook for mypy?

vale-salvatelli, line 30: why not adding also a hook for mypy?

fepegar: We can. @ant0nsc feared that it might make things slow.

fepegar:
Given that all the flake8 files are identical and that adding multiple configs in the repo doesn't feel right in terms on maintainability etc., I suggest we start with this simple setup and move on to different ones if necessary.

vale-salvatelli:
> Given that all the flake8 files are identical and that adding multiple configs in the repo doesn't feel right in terms on maintainability etc., I suggest we start with this simple setup and move on to different ones if necessary.

seems reasonable if the files are identical. Let's just make sure it points towards at least one flake8.ini? I don't see any in the root where the config is currently placed

.pre-commit-config.yaml
fepegar: As below, I suggest leaving it out for now and adding it later on if we want to.

fepegar:
> seems reasonable if the files are identical. Let's just make sure it points towards at least one flake8.ini? I don't see any in the root where the config is currently placed

Addressed in e34c759.

ant0nsc (approved):

.pre-commit-config.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+# The structure of this was suggested by the author of pre-commit and maintainer of flake8
+# See https://stackoverflow.com/a/66485642/3956024
+- repo: https://github.com/PyCQA/flake8
+  rev: 4.0.1
+  hooks:
+  - id: flake8
+    name: flake8 ./hi-ml/
+    alias: flake8-hi-ml
+    files: ^hi-ml/
```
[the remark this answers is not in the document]
fepegar: Doesn't that happen by default?

vale-salvatelli (approved):

.pre-commit-config.yaml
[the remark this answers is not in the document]
fepegar: Resolving this as I'm pretty sure flake8 won't look at other types of files.

PR #375, opened 2022-05-24
BUG: VSCode uses wrong value for Python path in terminal

ant0nsc:


fepegar (approved):
Works for me. Thanks, Anton!

peterhessey (approved):

PR #378, opened 2022-05-25
ENH: Add new DeepSMILETilesPanda config

mebristo:
Add new DeepSMILETilesPanda class and rename old one


hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -94,7 +116,171 @@ def get_data_module(self) -> PandaTilesDataModule:
...
+                                           self.pool_out_dim)
+        elif self.pool_type == GatedAttentionLayer.__name__:
+            pooling_layer = GatedAttentionLayer(num_encoding,
+                                                self.pool_hidden_dim,
+                                                self.pool_out_dim)
+        elif self.pool_type == MeanPoolingLayer.__name__:
+            pooling_layer = MeanPoolingLayer()
+        elif self.pool_type == MaxPoolingLayer.__name__:
+            pooling_layer = MaxPoolingLayer()
+        elif self.pool_type == TransformerPooling.__name__:
```
harshita-s, line 268: `TransformerPooling` is repeated here, but we have two different transformer pooling layers in cpath. We need to rename one of these.

hi-ml-histopathology/src/histopathology/datamodules/panda_module_baseline.py
```diff
@@ -0,0 +1,34 @@
+"""
```
harshita-s, line 1: Copyright header is missing

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -3,27 +3,49 @@
 #  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
```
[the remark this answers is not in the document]
vale-salvatelli: 2. I agree with @harshita-s that renaming the new config would bring in less changes. I like the word `Benchmark`

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -94,7 +116,171 @@ def get_data_module(self) -> PandaTilesDataModule:
...
+                                      EncodeTilesBatchd(image_key, self.encoder, chunk_size=self.encoding_chunk_size)
+                                      ])
+        else:
+            # Use same transforms as demonstrated in
+            # https://github.com/Project-MONAI/tutorials/blob/master/pathology/multiple_instance_learning/panda_mil_train_evaluate_pytorch_gpu.py
+            transform_train = Compose([
+                                      LoadTilesBatchd(image_key, progress=False),
+                                      transform_dict_adaptor(RandomHorizontalFlip(p=0.5), image_key, image_key),
+                                      transform_dict_adaptor(RandomVerticalFlip(p=0.5), image_key, image_key),
+                                      transform_dict_adaptor(RandomRotationByMultiplesOf90(), image_key, image_key),
```
vale-salvatelli, line 171: Has this config been tested? I believe ` transform_dict_adaptor(RandomRotationByMultiplesOf90(), image_key, image_key)` is throwing a pickle error. We discussed to try out the MONAI or torch transform implementation to avoid it.

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
vale-salvatelli: I think is MaxPoolingLayer that is repeated

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
harshita-s: @vale-salvatelli  this is the TilesDeepSmilePanda config where it works. IT was throwing an error on the Slides config, which is now changed (under review in cpath)

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -94,7 +116,171 @@ def get_data_module(self) -> PandaTilesDataModule:
         )
 
     def get_slides_dataset(self) -> Optional[PandaDataset]:
-        return PandaDataset(root=self.local_datasets[1])                             # type: ignore
+        return PandaDataset(root=self.local_datasets[1])  # type: ignore
+
+
+class DeepSMILETilesPanda(DeepSMILETilesPandaBasic):
```
vale-salvatelli, line 122: 1. I would definitely recommend putting the benchmark config, whatever is called, in a different module. This would make it easier to find for the user. 
2. Technically - the benchmark config to reproduce Myronenko et al. 2021 should be the one that starts from slides (therefore inherits from DeepSMILESlidesPanda)

hi-ml-histopathology/src/histopathology/datamodules/panda_module_baseline.py
vale-salvatelli, line 1: I don't have strong opinions on which tag we use to define this benchmark/baseline/experiment but I think important we stick to the same all over the repo - here the module and the class are called baseline for example but this name is not used in the container class.

hi-ml-histopathology/src/histopathology/models/deepmil.py
```diff
@@ -323,6 +323,18 @@ def update_results_with_data_specific_info(self, batch: dict, results: dict) ->
...
+class PandaTilesDeepMILModule(TilesDeepMILModule):
+    def __init__(self, n_epochs: int, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+        self.save_hyperparameters()
+        self.n_epochs = n_epochs
+
+    def configure_optimizers(self) -> Dict[str, Any]:  # type: ignore
+        optimizer = optim.AdamW(self.parameters(), lr=self.l_rate, weight_decay=self.weight_decay)
+        scheduler = optim.lr_scheduler.CosineAnnealingLR(optimizer=optimizer, T_max=self.n_epochs, eta_min=0)
+        return {"optimizer": optimizer, "lr_scheduler": scheduler}
```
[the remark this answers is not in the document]
harshita-s: Agreed, the optimizer and learning rate scheduler specific to paper, need to be renamed

mebristo:
I have opened a new PR to replace this one since the scope is different (slides instead of tiles): https://github.com/microsoft/hi-ml/pull/385

PR #383, opened 2022-05-26
ENH: Upload Codecov results also when tests fail

ant0nsc:
Also retiring the testing of the `*.tar.gz` packages

### DIFF
--- a/.github/workflows/hi-ml-pr.yml
+++ b/.github/workflows/hi-ml-pr.yml
@@ -146,7 +146,6 @@ jobs:
     strategy:
       matrix:
         folder: [ hi-ml, hi-ml-azure ]
-        packageName: [ '*.whl', '*.tar.gz' ]
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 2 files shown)


fepegar:
Could you please give some context? If tests fail, loads of lines won't be covered. Why would we want to upload the coverage in that case?

ant0nsc:
@fepegar At present, we don't get any coverage information if any single test fails. So the workflow will need to be to first get all tests to pass, and then tweak coverage if needed. I much prefer to see as many problems as possible during a single workflow run, so that I can try to fix them all in one swoop. Yes, the coverage information will not be 100% complete, but at least I have something to work with.

fepegar:
But won't that confuse Codecov in terms of coverage differences?

For example:
1. Coverage is 80%
2. Tests fail and only 5% of lines were covered, coverage is uploaded
3. Tests are fixed and coverage goes back to 80%. Would it show an artificial difference in coverage of 75%? Maybe I'm missing something.

Another question: can `if always()` be omitted?

ant0nsc:
As far as I understand it, even if tests fail, they would still produce coverage output? Coverage just checks which parts of the code are executed, if a subsequent assert is failing that's orthogonal.

No, you can't omit the `always()` because the default is `success()`

fepegar (approved):

peterhessey (approved):

PR #390, opened 2022-05-31
ENH: Add visualization parameters to DSA annotations script

fepegar:
This PR adds the `rescale` and `colormap` parameters to the script in the `girder` module, to improve flexibility and facilitate experimentation.

## Defaults

`Greens` is the default colormap because

1. It goes from white to a dark colour. This is nice because the slides background is typically white as well.
2. Green has the best contrast to the typical tissue colour (~magenta)

Rescaling is not performed by default because it might artificially emphasise attention values that are only slight larger than others. E.g., compare (0.04, 0.06, 0.95) and (0.32, 0.33, 0.34) with their rescaled equivalents: (0, 0.02, 1) and (0, 0.5, 1). The differences between values of the second array are very exaggerated. However, as pointed out in our meetings, this might not be relevant because the values depend on the number of tiles anyway, so we might as well simply use the logits. To be considered in the future.

## Screenshots

Default (no rescaling, `Greens`):

Screenshot 2022-05-31 101513

Rescaling, `Greens:

Screenshot 2022-05-31 101534

No rescaling, `viridis`:

Screenshot 2022-05-31 101555

Rescaling, `viridis`:

Screenshot 2022-05-31 101612

### DIFF
--- a/hi-ml-histopathology/src/histopathology/utils/girder.py
+++ b/hi-ml-histopathology/src/histopathology/utils/girder.py
@@ -290,6 +290,7 @@ class Item:
     :raises ValueError: If no ID or JSON is passed.
     :raises ValueError: If both an ID and JSON are passed.
     """
+
     def __init__(self, dsa: DigitalSlideArchive, id: Optional[str] = None, json: Optional[Dict] = None):
         self._dsa = dsa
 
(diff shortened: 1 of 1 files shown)


vale-salvatelli (approved):

fepegar:
> Thanks @fepegar for the rescaling and visualization examples. Could we also display the 0-1 scale somewhere? This could be useful as different colormaps appear differently, e.g., Greens have lighter color for smaller values, but it is vice-versa for Viridis. One question that we had was which colors are the higher values, perhaps the scale could make this clearer?

I agree a colour bar would definitely be useful. I haven't seen this in DSA, I'm not sure it's supported. But I'll contact the developers and add it in a future PR, if possible.

PR #399, opened 2022-06-01
ENH: Add gpu tests coverage to codecov reports

kenza-bouzid:
Add pytest_after_submission_hook to download pytest gpu test results and add it to codecov report.

### DIFF
--- a/.github/workflows/histopathology-pr.yml
+++ b/.github/workflows/histopathology-pr.yml
@@ -16,7 +16,7 @@ on:
 env:
   pythonVersion: 3.7
   folder: hi-ml-histopathology
-  module: histopathology
+  module_for_coverage_reporting: histopathology
   HIML_TENANT_ID: ${{ secrets.HIML_TENANT_ID }}
   HIML_RESOURCE_GROUP: ${{ secrets.HIML_RESOURCE_GROUP }}
   HIML_SUBSCRIPTION_ID: ${{ secrets.HIML_SUBSCRIPTION_ID }}
@@ -116,7 +116,7 @@ jobs:
           branch_prefix="refs/heads/"
           full_branch_name=$GITHUB_REF
           branch_name_without_prefix=${full_branch_name#$branch_prefix}
-          python hi-ml-azure/run_pytest.py --mark=gpu --cluster=pr-gpu --conda_env=${{ env.folder }}/environment.yml --folder=${{ env.folder }} --module=${{ env.module }} --experiment="$branch_name_without_prefix"
+          python hi-ml-azure/run_pytest.py --mark=gpu --cluster=pr-gpu --conda_env=${{ env.folder }}/environment.yml --folder=${{ env.folder }} --coverage_module=${{ env.module_for_coverage_reporting }} --experiment="$branch_name_without_prefix"
 
       - name: Upload gpu tests coverage report to Codecov
         # Coverage should also be uploaded if tests still fail
--- a/.github/workflows/histopathology-pr.yml
+++ b/.github/workflows/histopathology-pr.yml
@@ -16,6 +16,7 @@ on:
 env:
   pythonVersion: 3.7
   folder: hi-ml-histopathology
+  module: histopathology
   HIML_TENANT_ID: ${{ secrets.HIML_TENANT_ID }}
   HIML_RESOURCE_GROUP: ${{ secrets.HIML_RESOURCE_GROUP }}
   HIML_SUBSCRIPTION_ID: ${{ secrets.HIML_SUBSCRIPTION_ID }}
@@ -115,7 +116,8 @@ jobs:
           branch_prefix="refs/heads/"
           full_branch_name=$GITHUB_REF
           branch_name_without_prefix=${full_branch_name#$branch_prefix}
-          python hi-ml-azure/run_pytest.py --mark=gpu --cluster=pr-gpu --conda_env=${{ env.folder }}/environment.yml --folder=${{ env.folder }} --experiment="$branch_name_without_prefix"
+          python hi-ml-azure/run_pytest.py --mark=gpu --cluster=pr-gpu --conda_env=${{ env.folder }}/environment.yml
+          --folder=${{ env.folder }} --module=${{ env.module }} --experiment="$branch_name_without_prefix"
 
       - name: Upload gpu tests coverage report to Codecov
         # Coverage should also be uploaded if tests still fail
--- a/.github/workflows/histopathology-pr.yml
+++ b/.github/workflows/histopathology-pr.yml
@@ -115,4 +115,12 @@ jobs:
           branch_prefix="refs/heads/"
           full_branch_name=$GITHUB_REF
           branch_name_without_prefix=${full_branch_name#$branch_prefix}
-          python hi-ml-azure/run_pytest.py --mark=gpu --cluster=pr-gpu --conda_env=${{ env.folder }}/environment.yml --folder=${{ env.folder }} --experiment="$branch_name_without_prefix"
+          python hi-ml-azure/run_pytest.py --mark=gpu --cluster=pr-gpu --conda_env=${{ env.folder }}/environment.yml
+          --folder=${{ env.folder }} --experiment="$branch_name_without_prefix"
+
+      - name: Upload gpu tests coverage report to Codecov
+        # Coverage should also be uploaded if tests still fail
+        if: always()
+        uses: codecov/codecov-action@v2
+        with:
+            flags: ${{ env.folder }}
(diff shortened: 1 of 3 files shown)


hi-ml-azure/run_pytest.py
```diff
@@ -42,6 +46,11 @@ class RunPytestConfig(param.Parameterized):
         doc="The file or folder of tests that should be run. This value is used as the first argument to start "
         "pytest, so it can also be a specific test like 'my_test.py::any_test'",
     )
+    module: str = param.String(
+        default="",
+        doc="The module of tests that should be run. This value is used as an argument to --cov of pytest to run code"
```
fepegar, line 52: I think you're missing a space at the end.

hi-ml-azure/run_pytest.py
```diff
@@ -59,18 +68,22 @@ class RunPytestConfig(param.Parameterized):
...
     AzureML, this will make the job fail.
 
     :param pytest_mark: The PyTest mark to use for filtering out the tests to run.
     :param folder_to_test: The folder with tests that should be run.
+    :param module_to_test: The module for which test code coverage should be run.
     """
     results_file = Path(OUTPUT_FOLDER) / PYTEST_RESULTS_FILE
     pytest_args = [folder_to_test, f"--junitxml={str(results_file)}"]
 
+    if module_to_test:
```
fepegar, line 85: Does this mean that this argument could be an empty string? None? I think this should be documented.

hi-ml-azure/run_pytest.py
```diff
@@ -81,6 +94,57 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
         raise ValueError(f"PyTest failed with exit code: {status_code}")
 
 
+def download_run_output_file(blob_path: Path, destination: Path, run: Run) -> Path:
+    """
+    Downloads a single file from the run's default output directory: DEFAULT_AML_UPLOAD_DIR ("outputs").
```
fepegar, line 101: Maybe not a good idea to mention the value of that variable. If it changes, this docstring will be out of date.

hi-ml-azure/run_pytest.py
```diff
@@ -81,6 +94,57 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
...
+    For example, if blobs_path = "foo/bar.csv", then the run result file "outputs/foo/bar.csv" will be downloaded
+    to <destination>/bar.csv (the directory will be stripped off).
```
fepegar, line 103: I would add this explanation in the `blobs_path` parameter line.

hi-ml-azure/run_pytest.py
```diff
@@ -81,6 +94,57 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
...
+    :param blob_path: The name of the file to download.
+    :param run: The AzureML run to download the files from
```
fepegar, line 105: Period is missing.

hi-ml-azure/run_pytest.py
```diff
@@ -81,6 +94,57 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
...
+    :param destination: Local path to save the downloaded blob to.
+    :return: Destination path to the downloaded file(s)
+    """
+    blobs_prefix = str((DEFAULT_AML_UPLOAD_DIR / blob_path).as_posix())
```
fepegar, line 109: Why do you need to call `as_posix`?

hi-ml-azure/run_pytest.py
```diff
@@ -81,6 +94,57 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
...
+        raise ValueError(f"Unable to download file '{blobs_prefix}' from run {run.id}") from ex
+    return destination
+
+
+def download_pytest_result(run: Run, destination_folder: Path = Path.cwd()) -> Path:
+    """
+    Downloads the pytest result file that is stored in the output folder of the given AzureML run.
+    If there is no pytest result file, throw an Exception.
+    :param run: The run from which the files should be read.
+    :param destination_folder: The folder into which the PyTest result file is downloaded.
```
fepegar, line 124: Inconsistent pytest vs PyTest

hi-ml-azure/run_pytest.py
```diff
@@ -81,6 +94,57 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
...
+    :return: The path (folder and filename) of the downloaded file.
+    """
+    logging.info(f"Downloading pytest result file: {PYTEST_RESULTS_FILE}")
+    try:
+        return download_run_output_file(Path(PYTEST_RESULTS_FILE), destination=destination_folder, run=run)
+    except ValueError:
+        raise ValueError(f"No pytest result file {PYTEST_RESULTS_FILE} was found for run {run.id}")
```
fepegar: Maybe you can raise from the previous, as in line 113. I guess that would give more context.

hi-ml-azure/run_pytest.py
```diff
@@ -81,6 +94,57 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
...
+    destination = destination / blob_path.name
+    logging.info(f"Downloading single file from run {run.id}: {blobs_prefix} -> {str(destination)}")
+    try:
+        run.download_file(blobs_prefix, str(destination), _validate_checksum=True)
+    except Exception as ex:
+        raise ValueError(f"Unable to download file '{blobs_prefix}' from run {run.id}") from ex
+    return destination
+
+
+def download_pytest_result(run: Run, destination_folder: Path = Path.cwd()) -> Path:
```
fepegar: I'm not sure about this, but it might be nicer to set the default destination folder to None in the definition, and call cwd in the body if None.

ant0nsc (changes requested):
as discussed in a call, the present code does not download the coverage file

hi-ml-azure/run_pytest.py
```diff
@@ -81,6 +94,57 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
...
+    """
+    logging.info(f"Downloading pytest result file: {PYTEST_RESULTS_FILE}")
+    try:
+        return download_run_output_file(Path(PYTEST_RESULTS_FILE), destination=destination_folder, run=run)
+    except ValueError:
+        raise ValueError(f"No pytest result file {PYTEST_RESULTS_FILE} was found for run {run.id}")
+
+
+def pytest_after_submission_hook(azure_run: Run) -> None:
+    """
```
fepegar, line 135: One liner docstrings should have the quotation marks in the same line, before and after the string.

hi-ml-azure/run_pytest.py
```diff
@@ -81,6 +94,57 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
...
+    A function that will be called right after pytest gpu tests submission.
```
fepegar, line 136: Suggestion: replace "a function" with "hook".

hi-ml-azure/run_pytest.py
```diff
@@ -107,6 +171,7 @@ def run_pytest(folder_to_test: str, pytest_mark: str) -> None:
                 snapshot_root_directory=git_repo_root_folder(),
                 conda_environment_file=config.conda_env,
                 experiment_name=config.experiment,
-                max_run_duration=config.max_run_duration
+                max_run_duration=config.max_run_duration,
+                after_submission=pytest_after_submission_hook
```
fepegar, line 177: This is a good example of why it's a good idea to add a trailing comma after the last argument. So I suggest adding a comma after "hook".

fepegar (changes requested):
Looks good in general, added some minor suggestions.

hi-ml-azure/run_pytest.py
```diff
@@ -42,6 +47,12 @@ class RunPytestConfig(param.Parameterized):
         doc="The file or folder of tests that should be run. This value is used as the first argument to start "
         "pytest, so it can also be a specific test like 'my_test.py::any_test'",
     )
+    module: str = param.String(
```
ant0nsc, line 50: Name and description are slightly misleading. Maybe it should be called "coverage_module", because which tests we run is completely independent of what we collect coverage on.

ant0nsc: Also say that if this is empty, no coverage is collected.

.github/workflows/histopathology-pr.yml
```diff
@@ -16,6 +16,7 @@ on:
 env:
   pythonVersion: 3.7
   folder: hi-ml-histopathology
+  module: histopathology
```
ant0nsc, line 19: I'm not sure this is worth defining the constant here, you're using it once? Also, name is misleading, it should be "module_for_coverage_reporting" or something like that.

ant0nsc (approved):

hi-ml-azure/run_pytest.py
ant0nsc: Let's cross that bridge when we get there, and keep this PR simple and merge it asap

fepegar (approved):
Approved with a couple of minor docs points.

PR #400, opened 2022-06-01
BUG: TilesPandaImageNetMIL runs hang on Azure ML

harshita-s:
In this PR, `max_bag_size=56` and `max_bag_size_inf=0` are added to the constructor of `DeepSMILETilesPanda`. The runs get completed on Azure ML by using these bag sizes.

### DIFF
--- a/hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
+++ b/hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
@@ -77,8 +77,6 @@ class DeepSMILETilesPanda(BaseMILTiles, BaseDeepSMILEPanda):
     def setup(self) -> None:
         if self.encoder_type == SSLEncoder.__name__:
             self.downloader = self.download_ssl_checkpoint(innereye_ssl_checkpoint_binary)
-        if self.is_finetune:
-            self.batch_size = 2
         BaseMILTiles.setup(self)
 
     def get_data_module(self) -> PandaTilesDataModule:
@@ -142,8 +140,6 @@ class DeepSMILESlidesPanda(BaseMILSlides, BaseDeepSMILEPanda):
     def setup(self) -> None:
         if self.encoder_type == SSLEncoder.__name__:
             self.downloader = self.download_ssl_checkpoint(innereye_ssl_checkpoint_binary)
-        if self.is_finetune:
-            self.batch_size = 2
         BaseMILSlides.setup(self)
 
     def get_dataloader_kwargs(self) -> dict:


hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -67,6 +67,9 @@ def __init__(self, **kwargs: Any) -> None:
         default_kwargs = dict(
             # declared in BaseMILTiles:
             is_caching=False,
+            max_bag_size=56,
+            max_bag_size_inf=0,
+            batch_size=8,
```
kenza-bouzid, line 72: Small suggestion: maybe declare this in BaseDeepSMILEPanda that works for both tiles and slides?

harshita-s: Good point. It is already declared in `DeepSMILESlidesPanda`, could declare it only once in `BaseDeepSMILEPanda`

harshita-s: done, thanks

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -76,6 +77,8 @@ def __init__(self, **kwargs: Any) -> None:
     def setup(self) -> None:
         if self.encoder_type == SSLEncoder.__name__:
             self.downloader = self.download_ssl_checkpoint(innereye_ssl_checkpoint_binary)
+        if self.is_finetune:
+            self.batch_size = 2
```
vale-salvatelli, line 81: why fixing the batch size here? I see two issues:
- this is an unexpected counter effect of setup - why setup should overwrite one of the input parameter?
- the optimal batch_size will depend on a number of factors (eg number of tiles, tiles size, instance memory), we shouldn't hardcode one.

An alternative solution can be in the init to suggest a certain batch_size - given number of tiles, tiles size, instance memory

vale-salvatelli: "this didn't override the original batch_size when is_finetune was set" for my own understanding can you explain why? I guess this is a problem that doesn't affect only batch_size, does it mean we need to change something in the constructor to make sure params are correctly passed?

harshita-s: I wanted that whenever --is_finetune is set, it automatically takes batch_size=2 and overrides other batch sizes, else it will give OOM. I tried using this condition in the constructor, but it didn't override the batch size if we give a different batch size in the command line.

harshita-s: @kenza-bouzid I think I may have tried both as I was trying to work this out, but I've noted your suggestion :)

PR #404, opened 2022-06-06
BUG: Correctly setting up encoder for SSLMIL 

harshita-s:
In this PR, 
- `encoders.py` is modified to correctly load an SSL encoder. 
- Test is added for SSL encoder.
- The best SSL checkpoint is changed for PANDA.

### DIFF
--- a/hi-ml-histopathology/src/histopathology/configs/run_ids.py
+++ b/hi-ml-histopathology/src/histopathology/configs/run_ids.py
@@ -1,6 +1,6 @@
 # TODO: remove innereye references
 innereye_ssl_checkpoint = "hsharma_panda_explore:hsharma_panda_explore_1638437076_357167ae"
-innereye_ssl_checkpoint_binary = "hsharma_panda_tiles_ssl:hsharma_panda_tiles_ssl_1639766433_161e03b9"
+innereye_ssl_checkpoint_binary = "PANDA_SimCLR:PANDA_SimCLR_1654023538_7a0f1d00"
 innereye_ssl_checkpoint_crck_4ws = "simclr_8_gpus:CRCK_SimCLR_1647951559_d2a19df5"
 innereye_ssl_checkpoint_crck_radiomics = "ModifyOldSSLCheckpoint:704b1af8-7c75-46ed-8460-d80a0e603194"
 
--- a/hi-ml-histopathology/src/histopathology/configs/run_ids.py
+++ b/hi-ml-histopathology/src/histopathology/configs/run_ids.py
@@ -1,8 +1,8 @@
 # TODO: remove innereye references
-innereye_ssl_checkpoint = "hsharma_panda_explore:hsharma_panda_explore_1638437076_357167ae"
-innereye_ssl_checkpoint_binary = "PANDA_SimCLR:PANDA_SimCLR_1654023538_7a0f1d00"
-innereye_ssl_checkpoint_crck_4ws = "simclr_8_gpus:CRCK_SimCLR_1653673515_42d53d78"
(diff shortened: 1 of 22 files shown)


vale-salvatelli:
> In this PR, `setup_feature_extractor` is modified to correctly load a custom encoder (e.g., SSL encoder). Also, the best SSL checkpoint is changed for TCGA-CRCk.

the checkpoint seems to be changed for PANDA, not TCGA. Is it a mistake in this description?

harshita-s:
> > In this PR, `setup_feature_extractor` is modified to correctly load a custom encoder (e.g., SSL encoder). Also, the best SSL checkpoint is changed for TCGA-CRCk.
> 
> the checkpoint seems to be changed for PANDA, not TCGA. Is it a mistake in this description?

Yes, thanks for spotting! Changed the description now

hi-ml-histopathology/src/histopathology/utils/layer_utils.py
```diff
@@ -23,7 +23,16 @@ def setup_feature_extractor(pretrained_model: nn.Module,
...
+        layers = list(pretrained_model.children())
+        if len(layers) <= 1:
+            # If `layers` contains only one element (the entire model)
+            # Fetch the model and find its children (individual layers)
+            # Remove the last classification layer
+            layers = list(list(pretrained_model.children())[0].children())[:-1]
+        else:
+            # If `layers` contains multiple elements, these are the individual model layers
+            # Remove the last classification layer
+            layers = list(pretrained_model.children())[:-1]
```
vale-salvatelli, line 35: good catch - can we extend the test to capture this case? this will ensure we don't encounter the same bug in the future (and it will avoid we decrease test coverage with this PR)

hi-ml-histopathology/src/histopathology/configs/run_ids.py
```diff
@@ -1,6 +1,6 @@
 # TODO: remove innereye references
 innereye_ssl_checkpoint = "hsharma_panda_explore:hsharma_panda_explore_1638437076_357167ae"
-innereye_ssl_checkpoint_binary = "hsharma_panda_tiles_ssl:hsharma_panda_tiles_ssl_1639766433_161e03b9"
+innereye_ssl_checkpoint_binary = "PANDA_SimCLR:PANDA_SimCLR_1654023538_7a0f1d00"
```
ant0nsc, line 3: minor thing: Can you try if you can drop the "PANDA_SimCLR:" prefix?

hi-ml-histopathology/src/histopathology/utils/layer_utils.py
harshita-s: @vale-salvatelli  To capture this case, a test needs to be added for the SSL encoder in `test_encoders.py`. Not sure why this was initially not added, is it because it requires a checkpoint path to download the weights? Which checkpoint path should be given here, can a test SSL run be used?

vale-salvatelli: could you check with Melissa? she has maybe a test run that could be used. Otherwise, instead of downloading, we could patch the SSL encoder with an encoder initialized in the test - we don't care in this case if the weights are random but just that we're selecting the right layer

hi-ml-histopathology/src/histopathology/configs/run_ids.py
harshita-s: Dropped, it works without the prefix.

hi-ml-histopathology/src/histopathology/utils/layer_utils.py
harshita-s: Thanks. I have added a test for the SSL encoder.

dccastro (approved):
Great stuff, thanks for addressing this!

vale-salvatelli (approved):

PR #409, opened 2022-06-07
STYLE: Fix file without end line

fepegar:
Picked up by `pre-commit`.

PR #410, opened 2022-06-07
STYLE: Ensure new line at the end of JSON files

fepegar:
This is to make sure that pre-commit doesn't complain about JSON files missing new lines.

See https://github.com/microsoft/hi-ml/pull/409#issuecomment-1148419642 for more context.


hi-ml-histopathology/src/histopathology/utils/report_utils.py
```diff
@@ -109,14 +109,14 @@ def collect_crossval_metrics(parent_run_id: str, download_dir: Path, aml_workspa
     metrics_json = download_dir / "aml_metrics.json"
     if not overwrite and metrics_json.is_file():
         print(f"AML metrics file already exists at {metrics_json}")
-        metrics_df = pd.read_json(metrics_json)
+        metrics_df = pd.read_json(metrics_json, lines=True)
```
mebristo, line 112: Is it ok that we don't specify `orient='records'` here?

ant0nsc (approved):

hi-ml-histopathology/src/histopathology/utils/report_utils.py
fepegar: Good catch. It worked in my tiny test, but I guess it's safest to add it here. I wonder if that's why the tests are failing. we'll see.

hi-ml-histopathology/src/histopathology/utils/report_utils.py
```diff
@@ -116,7 +117,7 @@ def collect_crossval_metrics(parent_run_id: str, download_dir: Path, aml_workspa
                                                   aml_workspace=aml_workspace)
         metrics_json.parent.mkdir(parents=True, exist_ok=True)
         print(f"Writing AML metrics file to {metrics_json}")
-        metrics_df.to_json(metrics_json)
+        df_to_json(metrics_df, metrics_json)
```
mebristo, line 120: Not a blocker for this PR but I wonder whether we should create this file by default in the runner, rather than only when a regression test folder is specified, or when the report is created

mebristo (approved):

hi-ml-histopathology/src/histopathology/utils/report_utils.py
fepegar: [This is now outdated as I've refactored the PR]

hi-ml-histopathology/src/histopathology/utils/report_utils.py
fepegar: I think that's unrelated to this PR. I'd suggest opening a discussion thread about this, if you like.

PR #417, opened 2022-06-07
DOC: Add documentation on DSA

fepegar:


ajsharda17 (approved):

docs/source/dsa.md
```diff
@@ -0,0 +1,80 @@
...
+By default, the application runs at port 8080.
+To use HTTPS, an SSL certificate must be obtained and port 443 may be specified instead within the Docker configuration.
+The DSA uses [Cherrypy][8] as the underlying server engine.
+Therefore, Cherrypy must be [configured][9] to use the SSL certificate installed in the VM.
+Ports in a VM can be opened using [network security group rules][7].
+
+### Storage
+
+The datasets we use are securely stored in [Azure Blob Storage][3] containers.
+Containers can be mounted on the host machine using [BlobFuse][13] and connected to Girder using [assetstores][12].
```
ant0nsc, line 25: Not clear to me. So I first have to mount the container, and then start assetstore? or does assetstore do that for me?

docs/source/dsa.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Authentication
+
+We have [added][14] Microsoft as a provider in the [OAuth2 Login][15] plugin for Girder.
+An app must be [registered][16] so that [Azure Active Directory][17] (Azure AD) can be used for authentication.
```
ant0nsc, line 30: For newbies, what is an app? But even knowing what an app is, I have trouble parsing that sentence. Once I have the app, what do I do with it?

docs/source/dsa.md
```diff
@@ -0,0 +1,80 @@
...
+
+The [Girder RESTful API][19] may be used to upload annotations to DSA items programmatically.
+An example of a use case is creating annotations to visualize attention maps generated by a deep learning model.
+The `girder` module in this repository includes tools to download training results from Azure Machine Learning (Azure ML) and
+upload annotations to a deployed DSA.
+For example:
+
+```shell
+SCRIPT="hi-ml-histopathology/src/histopathology/utils/girder.py"
+python $SCRIPT --run-id "Experiment_transformer_Gleason_run_0" --rescale
```
ant0nsc, line 46: As is, this will fail because it doesn't know which DSA to talk to. I would at least say that `--dsa-url` and api key must be provided on the commandline or in environment variables. Same for workspace config file. Does the script respond to "--help" to print all options? If yes, this should be mentioned.

ant0nsc, line 46: Out of scope for this PR, but just to say that we have not used "-" as the separator for commandline args throughout most of the codebase, but "_". This may be frowned upon as not pythonic, but has the huge advantage that we can use auto-generated parsers, which eliminate a classic source of errors.

docs/source/dsa.md
fepegar: Rephrased in 1d97167.

docs/source/dsa.md
fepegar: Rephrased in a31c1ac.

docs/source/dsa.md
fepegar: Addressed in f095998. Sensitive strings were randomly generated.

fepegar:
Thanks, @ant0nsc! Let me know if this looks better. Happy to keep iterating if necessary.

docs/source/index.rst
```diff
@@ -40,6 +40,7 @@ The `hi-ml` toolbox provides
    :caption: Histopathology
 
    histopathology.md
+   dsa.md
```
vale-salvatelli, line 43: if this is the title that appears in the hierarchy maybe I'd call it digital_slide_archive

vale-salvatelli (approved):

docs/source/index.rst
fepegar: That's just the file name. The docs look like this: https://hi-ml--417.org.readthedocs.build/en/417/dsa.html

ant0nsc (approved):

fepegar:
Thanks for reviewing!

PR #432, opened 2022-06-10
BUG: Stop looking for environment file if already running in AML

mebristo:
Currently, if the user specifies a relative path as `--conda_env`, it will be found by the first process which will then start training, and change the cwd to 'outputs'. This means that spawned processes can't find the relative conda_env path. This PR enables spawned processes to still find the conda env path

### DIFF
--- a/hi-ml/src/health_ml/runner.py
+++ b/hi-ml/src/health_ml/runner.py
@@ -27,9 +27,9 @@ from health_azure import AzureRunInfo, submit_to_azure_if_needed  # noqa: E402
 from health_azure.datasets import create_dataset_configs  # noqa: E402
 from health_azure.logging import logging_to_stdout   # noqa: E402
 from health_azure.paths import is_himl_used_from_git_repo  # noqa: E402
-from health_azure.utils import (get_workspace, is_local_rank_zero, merge_conda_files,  # noqa: E402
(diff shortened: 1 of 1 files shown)


hi-ml/src/health_ml/runner.py
```diff
@@ -245,7 +245,13 @@ def after_submission_hook(azure_run: Run) -> None:
...
         try:
             if self.experiment_config.cluster:
-                env_file = choose_conda_env_file(env_file=self.experiment_config.conda_env)
+                # Enable spawned processes to find relative env file regardless of CWD
+                conda_env_file = self.experiment_config.conda_env
+                if conda_env_file is not None:
+                    if str(self.project_root) not in str(conda_env_file):
+                        conda_env_file = self.project_root / conda_env_file
+
+                env_file = choose_conda_env_file(env_file=conda_env_file)
```
ant0nsc, line 254: I think this can be done simpler. We need to prevent lines 248 to be executed in AzureML altogether. Modify line 247 to be `if self.experiment_config.cluster and not is_running_in_azureml():`

ant0nsc (changes requested):
I think this can be achieved simpler, please see my comment.

ant0nsc (approved):

fepegar (approved):
I edited the PR title for consistency, I hope that's ok.

PR #440, opened 2022-06-16
BUG: Default parameters in `DeepSMILECRCk` container 

harshita-s:
We use caching with pre-trained encoders in DeepMIL for TCGA-CRCk. This is now enabled by default in `DeepSMILECRCk` container through the `is_caching` parameter, by setting it to `True`. Also, `num_top_slides` is set to 0 by default, to avoid errors arising from caching in outputs handler for TCGA-CRCk.

### DIFF
--- a/hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILECrck.py
+++ b/hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILECrck.py
@@ -38,7 +38,8 @@ class DeepSMILECrck(BaseMILTiles):
             num_transformer_pool_heads=4,
             encoding_chunk_size=60,
             is_finetune=False,
-            is_caching=False,
+            is_caching=True,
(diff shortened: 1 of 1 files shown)


kenza-bouzid (approved):
LGTM, please keep in mind that setting num_top_slides to 0 disables top and bottom tiles plots and not slides outputs. Slides outputs (thumbnails and attention heatmaps) are disabled by default if a SlidesDataset is not provided which is the case for TCGACrck.

harshita-s:
> LGTM, please keep in mind that setting num_top_slides to 0 disables top and bottom tiles plots and not slides outputs. Slides outputs (thumbnails and attention heatmaps) are disabled by default if a SlidesDataset is not provided which is the case for TCGACrck.

Thank you @kenza-bouzid. I have modified the description now.

maxilse (approved):

PR #443, opened 2022-06-16
ENH: Make building Monai-based models reproducible

ant0nsc:
- Control randomness in MONAI transforms
- Ensure PyTorch Lightning seeds the dataloader worker process


hi-ml-histopathology/testhisto/testhisto/datasets/test_panda_slides.py
```diff
@@ -0,0 +1,76 @@
...
+    container.tile_size = tile_size
+    container.max_bag_size = num_tiles
+    container.local_datasets = [tmp_path]
+
+    def get_items_from_new_loader(num_items: int) -> Any:
+        data_module = container.get_data_module()
+        data_module.dataloader_kwargs = {**data_module.dataloader_kwargs,
+                                         "multiprocessing_context": None,
+                                         "num_workers": 0}
+        loader = data_module.train_dataloader()
```
kenza-bouzid, line 50: You might want to test the val_dataloader and test_dataloader as well especially since we apply different transforms. It's true that we don't apply any random transforms for val and test but testing it might help us catch any source of randomness during inference (e.g if someone accidently uses a random transform or shuffle the val/test dataloaders without properly fixing the seeds).

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
```diff
@@ -312,6 +312,10 @@ def _load_dataset(self, slides_dataset: SlidesDataset, stage: ModelKey) -> Datas
             transforms = Compose([base_transform, self.transforms_dict[stage]]).flatten()
         else:
             transforms = base_transform
+        if self.seed is not None:
+            # The tiling transform is randomized. Make them deterministic. This call needs to be
+            # done on the final Compose, not at the level of the individual randomized transforms.
+            transforms.set_random_state(self.seed)
```
kenza-bouzid, line 318: I wonder if we still need to set_random_state for the previous  transform_train and transform_inf. They are going to be composed with the base transforms line 312 and what matters is to set_random_state on the final Compose line.318, is that right? so no need to fix the seed at the container level if it's controlled by the datamodule?
Note: This is even better because we control sources of randomness related to monai transforms with a single line of code in a much safer way (e.g. it's less likely to make changes in the base datamodule vs containers that change constantly during experiments)
Also no need to add extra comments in configs as I suggest it before if we are able to fix it at the base data module level.

kenza-bouzid (commented):
Great work, thank you for putting this together, I left some few minor suggestions.

hi-ml-histopathology/src/histopathology/configs/classification/DeepSMILEPanda.py
```diff
@@ -157,6 +157,7 @@ def get_data_module(self) -> PandaSlidesDataModule:
             tile_size=self.tile_size,
             step=self.step,
             random_offset=self.random_offset,
+            seed=self.get_effective_random_seed(),
```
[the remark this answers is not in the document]
ant0nsc: yup, thanks!

hi-ml-histopathology/src/histopathology/datamodules/base_module.py
ant0nsc: Very good point indeed. I've simplified the code

hi-ml-histopathology/environment.yml
```diff
@@ -34,6 +34,7 @@ dependencies:
       - coloredlogs==15.0.1
       - cucim==22.04.00
       - girder-client==3.1.14
+      - itk==5.2.1.post1
```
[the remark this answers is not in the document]
ant0nsc: removed

hi-ml-histopathology/testhisto/testhisto/datasets/test_panda_slides.py
```diff
@@ -0,0 +1,91 @@
...
+    container.max_bag_size = num_tiles
+    container.local_datasets = [tmp_path]
+
+    def test_data_items_are_equal(loader_fn_names: List[str]) -> None:
+        """Creates a new data module from the container, and checks if all the data loaders specified in
+        `loader_fn_names` return the same set of items when enumerated twice."""
+        data_module = container.get_data_module()
+        data_module.dataloader_kwargs = {**data_module.dataloader_kwargs,
+                                         "multiprocessing_context": None,
+                                         "num_workers": 0}
```
kenza-bouzid, line 59: souldn't we test the case with multi_workers as well? to make sure the changes (in model_trainer) are effective

hi-ml-histopathology/testhisto/testhisto/datasets/test_panda_slides.py
```diff
@@ -0,0 +1,91 @@
...
+            print(f"Checking {loader_fn_name}")
+            item1 = get_item_from_new_loader(loader_fn_name)
+            item2 = get_item_from_new_loader(loader_fn_name)
+            assert item1[SlideKey.SLIDE_ID] == item2[SlideKey.SLIDE_ID], "Order of slides must match"
+            assert len(item1[SlideKey.IMAGE]) == len(item2[SlideKey.IMAGE]), "Length of images must match"
+            for i in range(len(item1[SlideKey.IMAGE])):
+                image1 = item1[SlideKey.IMAGE][i]
+                image2 = item2[SlideKey.IMAGE][i]
+                if len(np.unique(image1)) == 1:
+                    assert False, "Something is wrong here, image1 only has a single value"
```
kenza-bouzid, line 78: Is it really something wrong? a tile can be complete background and therefore have a single value. It might be because max_bag_size is bigger than the effective foreground in the WSI, in which case, we should rather make sure that we get a background tile deterministically. torch.allclose(image1, image2) should be valid in this case as well.
This is a special case where the mock WSI have exactly num_tiles of foreground that we use for tiling as well.
But in real life, not all slides will have the expected num_tiles and might include some background tiles.

hi-ml-histopathology/testhisto/testhisto/datasets/test_panda_slides.py
```diff
@@ -0,0 +1,91 @@
...
+try:
+    from cucim import CuImage  # noqa: F401
+    has_cucim = True
+except:  # noqa: E722
+    has_cucim = False
+
+
+@pytest.mark.skipif(not has_cucim, reason="Test requires CUCIM library")
+@pytest.mark.gpu
+def test_panda_reproducibility(tmp_path: Path) -> None:
```
kenza-bouzid, line 30: Is it somehow possible to write a DDP test as well? We might want to make sure that everything is properly seeded in DDP context.  You can use `run_distributed` from utils_testhisto.

hi-ml/src/health_ml/model_trainer.py
```diff
@@ -215,8 +215,8 @@ def model_train(checkpoint_path: Optional[Path],
     # Create the trainer object. Backup the environment variables before doing that, in case we need to run a second
     # training in the unit tests.
     old_environ = dict(os.environ)
-    # Set random seeds just before training
-    seed_everything(container.get_effective_random_seed())
+    # Set random seeds just before training. Ensure that dataloader workers are also seed correctly.
+    seed_everything(container.get_effective_random_seed(), workers=True)
```
kenza-bouzid, line 219: See comment above about testing with num_workkers>0

kenza-bouzid (commented):
Looking good, added some suggestions!

hi-ml-histopathology/testhisto/testhisto/datasets/test_panda_slides.py
ant0nsc: I added that check specifically because with all-ones slides the check below is meaningless. I agree that under normal circumstances it could happen that a tile is only background. Here in this test we are using a fixed seed, and will for sure not run into that problem.

hi-ml-histopathology/testhisto/testhisto/datasets/test_panda_slides.py
ant0nsc: Just using num_workers>0 will not be enough here because we need to test if Lightning does the right thing. Testing the end-to-end training would take the PR effectively onto @mebristo's territory - she has been working on all those tests, but is blocked now. I'll chat with her.

kenza-bouzid: Fair enough.

hi-ml/src/health_ml/run_ml.py
```diff
@@ -94,8 +96,11 @@ def setup(self, azure_run_info: Optional[AzureRunInfo] = None) -> None:
                         raise ValueError(f"Invalid setup: The dataset at index {i} is None")
                     local_datasets.append(check_dataset_folder_exists(dataset))
                 self.container.local_datasets = local_datasets  # type: ignore
-        # Ensure that we use fixed seeds before initializing the PyTorch models
-        seed_everything(self.container.get_effective_random_seed())
+        # Ensure that we use fixed seeds before initializing the PyTorch models.
+        # MONAI needs a separate method to make all transforms deterministic by default
+        seed = self.container.get_effective_random_seed()
+        seed_monai_if_available(seed)
```
vale-salvatelli, line 102: for my own understanding - to make the transforms deterministic we need both to fix the determinism in MONAI and to  fix a seed in the transform?

hi-ml/src/health_ml/model_trainer.py
ant0nsc: see above, I think we want to test that with @mebristo's regression tests rather.

ant0nsc:
> Thanks Anton, the PR looks good to me. 2 overall comments:
> 
> * Adding a PR description would help keeping track of what are the randomness sources that have been fixed in this PR
> * Is the plan to address in a separate PR randomness in models that uses pre-tiled tiles? to have an integration test that check the TCGA-CRCk DeepSMILE benchmark we need also the models using pre-processed tiles to be completely deterministic

PR description updated. 
For the pre-tiled datasets, there is no randomness that I'm aware of - beyond what is already controlled by the present setting of seeds.

hi-ml/src/health_ml/run_ml.py
ant0nsc: No, seeding MONAI separately is more of a preventative measure. If someone accidentally adds randomized transforms (and forgets to seed them), the code added here will probably be enough to still make it reproducible

kenza-bouzid (approved):

vale-salvatelli (approved):

hi-ml/src/health_ml/run_ml.py
vale-salvatelli: I like the extra caution :)

PR #444, opened 2022-06-20
ENH: Redirect profiler traces to outputs folder and add debugging doc

kenza-bouzid:


docs/source/debugging.md
```diff
@@ -0,0 +1,183 @@
...
+code runs as expected locally before submitting a job to AzureML. hi-ml supports a set of debugging flags that triggers
+[Pytorch Lightning Trainer](https://pytorch-lightning.readthedocs.io/en/stable/common/trainer.html#trainer) arguments to
+help you detect any potential errors or bugs at early stage.
+
+These are available as part of the
+[TrainerParams](https://github.com/microsoft/hi-ml/blob/96b8ba60ebf84416f5c9b13e2df15ee07a13e6bb/hi-ml/src/health_ml/deep_learning_config.py#L357)
+and can be used as extra command line arguments with the [hi-ml
+runner](https://github.com/microsoft/hi-ml/blob/746c8b58c1af71f71eeaaac2a8584be1d9a5386f/hi-ml/src/health_ml/runner.py#L107).
+
+* `pl-fast-dev-run`: If set to `n`, runs the pipeline for only `n` batch(es) of train, val and test for only a single
```
ant0nsc, line 62: Can you double check, I think the separator for the arg is underscore `_` rather than dash `-`?

docs/source/debugging.md
```diff
@@ -0,0 +1,183 @@
...
+* `pl-limit-val-batches`: Limits the validation dataset to the given number of batches `n`.
+* `pl-limit-train-batches`: Limits the test dataset to the given number of batches `n`.
+
+In general, it is very useful to run the following two steps as part of the developement cycle:
+
+1. Make sure all training, validation and test loops complete properly:
+
+```shell
+conda activate HimlHisto
+python ../hi-ml/src/health_ml/runner.py --model=YourCustomContainer --crossval-count=0 --bach-size=2 --pl-fast-dev-run=4
```
ant0nsc, line 76: I think we can skip the `crossval_count=0` flag, it is now off by default

docs/source/debugging.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Debugging configs
+
+The hi-ml repository is organised in
+[Multi-root workspaces](https://code.visualstudio.com/docs/editor/workspaces#_multiroot-workspaces) to account
+for environment differences among the modules and offer flexibility to configure each module seperatly.
+
+We provide a set of example debugging configs for each of hi-ml module:
+
+* [launch.json in hi-ml](https://github.com/microsoft/hi-ml/tree/main/hi-ml/.vscode/lanch.json)
```
mebristo, line 21: typo here - lanch.json

docs/source/debugging.md
```diff
@@ -0,0 +1,183 @@
...
+        "request": "launch",
+        "program": "${file}",
+        "console": "integratedTerminal",
+        "justMyCode": false
+      },
+    ],
+}
+```
+
+### Pytorch Lightning flags for debugging and quick runs
```
mebristo, line 46: Nit: PyTorch with capital T (there are a few other references to it on this page too)

docs/source/debugging.md
```diff
@@ -0,0 +1,183 @@
...
+
+The hi-ml toolbox is built upon [Pytorch Lightning (PL)](https://www.pytorchlightning.ai/) to help you build scalable
+deep learning models for healtcare and life sciences. Refer to [Running ML Experiments with hi-ml](runner.md) for
```
mebristo, line 49: typo: healthcare

docs/source/debugging.md
```diff
@@ -0,0 +1,183 @@
...
+argument `--pl_profiler` that you can set to either
+[`simple`](https://pytorch-lightning.readthedocs.io/en/stable/advanced/profiler.html#simple-profiler),
+[`advanced`](https://pytorch-lightning.readthedocs.io/en/stable/advanced/profiler.html#simple-profiler), or
+[`pytorch`](https://pytorch-lightning.readthedocs.io/en/stable/advanced/profiler.html#pytorch-profiler).
+
+The profiler outputs will be saved in a subfolder `profiler` inside the outputs folder of the run.
+
+```shell
+conda activate HimlHisto
+python ../hi-ml/src/health_ml/runner.py --model=YourCustomContainer --crossval-count=0 --bach-size=2 --pl-limit-train-batches=4 --pl-limit-val-batches=4 --pl-limit-test-batches=4 --max_epochs=4 --pl-profiler=pytorch
```
mebristo, line 104: It's not immediate obvious to the reader which is the important flag here. Perhaps we could instead say 'add the following flag to your command' and just show the flag of interest

ant0nsc (changes requested):
Looking great! One thing that we should discuss, are we passing constructor arguments for the profiler or is it a full object? Left a comment to that effect.

docs/source/debugging.md
kenza-bouzid: You're right!

docs/source/debugging.md
kenza-bouzid: Good to know, done!

docs/source/debugging.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+### Interpret Pytorch Profiling outputs via Tensorboard
+
+[Pytorch Profiler](https://pytorch.org/tutorials/recipes/recipes/profiler_recipe.html) can effectively be interpreted via
+TensorBoard dashbord interface that is integrated in [VS
+Code](https://code.visualstudio.com/docs/datascience/pytorch-support#_tensorboard-integration) as part of the Python
+extension. Once you have the outputs of the Pytorch Profiler in `outputs/YYYY-MM-DDTHHmmssZ_YourContainerName/pytorch_profiler`, you can
```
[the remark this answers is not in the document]
kenza-bouzid: fixed

hi-ml/src/health_ml/model_trainer.py
```diff
@@ -37,6 +38,22 @@ def write_experiment_summary_file(config: Any, outputs_folder: Path) -> None:
...
+    if pl_profiler:
+        pl_profilers = {"simple": SimpleProfiler, "advanced": AdvancedProfiler, "pytorch": PyTorchProfiler}
+        if pl_profiler not in pl_profilers:
+            raise ValueError("Unsupported profiler. Please choose one of the following options: simple, advanced, "
+                             "pytorch. You can refer to https://pytorch-lightning.readthedocs.io/en/stable/advanced/"
+                             "profiler.html to learn more about each profiler. You can specify additional arguments by "
+                             "overriding the default behavior of get_trainer_arguments() in your lightning container. "
+                             "You can find an example here https://github.com/microsoft/hi-ml/tree/main/docs/source/"
+                             "debugging.md#L145")
+        profiler = pl_profilers[pl_profiler](dirpath=outputs_folder / f"{pl_profiler}_profiler", **profiler_args)
```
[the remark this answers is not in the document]
kenza-bouzid: fair point!

docs/source/debugging.md
```diff
@@ -0,0 +1,183 @@
...
+  epoch. Additionally [this flag](https://pytorch-lightning.readthedocs.io/en/stable/common/debugging.html#fast-dev-run)
+  disables all callbacks and hyperparameters serialization which makes the debugging process very quick. This must be
+  used for debugging purposes only.
+* `pl-limit-train-batches`: Limits the training dataset to the given number of batches `n`.
+* `pl-limit-val-batches`: Limits the validation dataset to the given number of batches `n`.
+* `pl-limit-train-batches`: Limits the test dataset to the given number of batches `n`.
+
+In general, it is very useful to run the following two steps as part of the developement cycle:
+
+1. Make sure all training, validation and test loops complete properly:
```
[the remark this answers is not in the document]
kenza-bouzid: I dont' think we should add it to the Makefile because these are "toy" commands (YourcustomContainer doesn't exist for instance) to show how to debug using the hi-ml runner and the flags in TrainerParams.  

I'm afraid specifying only the flags might be confusing. which script to launch? 

What if we change HimlHisto by YourEnvName?

docs/source/debugging.md
kenza-bouzid: I just wanted to make it clear that these flags should be used along with the himl runner

kenza-bouzid:
> My only concern is there are a lot of links here that could potentially become broken/ outdated.. What would be our strategy for checking these links are intact? Maybe we should have a script for this?

Do we have a script for this? Do you check that for the rest of the documentation?

mebristo:
> > My only concern is there are a lot of links here that could potentially become broken/ outdated.. What would be our strategy for checking these links are intact? Maybe we should have a script for this?
> 
> Do we have a script for this? Do you check that for the rest of the documentation?

We don't currently but it might be a good idea going forward :) not a blocker for this PR tho

docs/source/debugging.md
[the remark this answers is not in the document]
mebristo: We should definitely remove HimlHisto (to be it might be overkill to tell the user to activate their environment anyway?) but my concern is that it's not clear which are the standard runner args and which are debugging args

docs/source/debugging.md
mebristo: You could explain that with a sentence tho rather than repeating the runner command

hi-ml/src/health_ml/model_trainer.py
```diff
@@ -140,9 +140,10 @@ def create_lightning_trainer(container: LightningContainer,
         else:
             callbacks.append(more_callbacks)  # type: ignore
     callbacks.extend(container.get_callbacks())
-    profiler_args = {}
-    if "profiler" in additional_args:
-        profiler_args = additional_args.pop("profiler")
+    # Set profiler: if --pl_profiler=profiler is specified, it overrides any custom profiler defined in the container
+    custom_profiler = additional_args.pop("profiler", PassThroughProfiler())
```
ant0nsc, line 144: nitpicking - using PassThroughProfiler instead of None requires a lot of complicated checks, is that really necessary?

ant0nsc (approved):

hi-ml/src/health_ml/model_trainer.py
kenza-bouzid: Yes, this is what PL trainer uses by default. I changed get_pl_profiller also to return PassThroughProfiler as default.

https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.profiler.PassThroughProfiler.html

kenza-bouzid: We can also use None for custom_profiler but it complicates the number of checks to make to pick either the pl_profiler or the custom one. 

Or do you prefer setting defaults to None for both custom and pl_profiler and adding a check line 185, `profiler=profiler if profiler else PassTroughProfiler()` for the trainer

kenza-bouzid: Oh no, my bad if we give it None pl will set it to PassTroughProfiler. it's like when container.pl_profiler was None by default, will fix that.

PR #448, opened 2022-06-21
ENH: Improve mypy arguments

fepegar:
1. Show mypy error codes in output. That way, we can check and ignore specific errors in a readable way. For example, `# type: ignore[arg-type]` is probably better than `# type: ignore`.
1. Spell out argument name for readability


peterhessey (approved):
LGTM!

ant0nsc (approved):

PR #450, opened 2022-06-22
BUG: Fix confusion matrix plots

kenza-bouzid:
Issue: Using ConfusionMatrix from torchmetrics in a DDP context triggers devices synchronization by default, however we only plot the CM from rank 0 which leads rank 0 to hang waiting for the other devices acknowledgement.

Solution: we don't need a distributed metric to compute the CM on rank 0 as the results have already been gathered across devices -> confusion_matrix from scikit-learn is more adapted in this context.


ant0nsc (commented):
@kenza-bouzid , can you say in the description what this PR is actually fixing? If it fixes something, it would be good to have a test that shows what is being fixed.

dccastro (changes requested):

hi-ml-histopathology/src/histopathology/utils/plots_utils.py
```diff
@@ -47,23 +46,13 @@ def save_scores_histogram(results: ResultsType, figures_dir: Path) -> None:
     save_figure(fig=fig, figpath=figures_dir / "hist_scores.png")
 
 
-def save_confusion_matrix(
-    conf_matrix_metric: ConfusionMatrix, class_names: Sequence[str], figures_dir: Path, stage: ModelKey
-) -> None:
+def save_confusion_matrix(results: ResultsType, class_names: Sequence[str], figures_dir: Path) -> None:
```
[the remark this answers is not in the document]
kenza-bouzid: Wasn't sure if I should add a test for that, we already have [test_plot_normalized_confusion_matrix](https://github.com/microsoft/hi-ml/blob/85ba1928540321fdfad2a1d18a01ed6369fc78d9/hi-ml-histopathology/testhisto/testhisto/utils/test_viz_utils.py#L204) for the plotting part. This leaves only line 55 untested but it's external code...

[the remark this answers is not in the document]
dccastro: That's why I'm suggesting an integration test, just to ensure that the tested parts are mutually compatible. 🙂

hi-ml-histopathology/src/histopathology/configs/classification/BaseMIL.py
```diff
@@ -183,12 +183,12 @@ def get_pooling_layer(self) -> Tuple[nn.Module, int]:
 
     def get_test_plot_options(self) -> Set[PlotOption]:
         if self.num_top_slides > 0:
-            return {PlotOption.HISTOGRAM, PlotOption.TOP_BOTTOM_TILES}
+            return {PlotOption.HISTOGRAM, PlotOption.CONFUSION_MATRIX, PlotOption.TOP_BOTTOM_TILES}
         else:
-            return {PlotOption.HISTOGRAM}
+            return {PlotOption.HISTOGRAM, PlotOption.CONFUSION_MATRIX}
```
vale-salvatelli, line 188: Minor suggestion: the below would be easier to read and avoid duplication
`def get_test_plot_options(self) -> Set[PlotOption]:
  options = {PlotOption.HISTOGRAM, PlotOption.CONFUSION_MATRIX}
   if self.num_top_slides > 0:
         options.update({PlotOption.TOP_BOTTOM_TILES})
    return options`

hi-ml-histopathology/testhisto/testhisto/utils/test_plots_utils.py
```diff
@@ -58,9 +58,7 @@ def assert_plot_func_called_if_among_plot_options(
         },
     ],
 )
-def test_plots_handler_plots_only_desired_plot_options(
-    plot_options: Collection[PlotOption],
-) -> None:
+def test_plots_handler_plots_only_desired_plot_options(plot_options: Collection[PlotOption],) -> None:
```
vale-salvatelli, line 61: following up on @ant0nsc 's comment: this test was not failing also before because it operates on a single GPU while the issue is with DDP, right? do we have a way to test outputs in DDP?

hi-ml-histopathology/src/histopathology/utils/output_utils.py
```diff
@@ -357,7 +356,6 @@ def save_test_outputs(self, epoch_results: EpochResultsType, is_global_rank_zero
         # All DDP processes must reach this point to allow synchronising epoch results
         gathered_epoch_results = gather_results(epoch_results)
         if PlotOption.TOP_BOTTOM_TILES in self.test_plots_handler.plot_options and self.tiles_selector:
-            logging.info("Selecting tiles ...")
```
vale-salvatelli, line 360: any specific reason to delete the logging?

kenza-bouzid: Not really, Anton mentioned earlier that we have a lot of verbosity here and there. I grouped everything in a single log in plots_utils.py line 194.

hi-ml-histopathology/testhisto/testhisto/utils/test_plots_utils.py
kenza-bouzid: The test didn't fail because I am mocking "save_confusion_matrix". Also, this test makes sure that we're calling the right functions for a set of plot_options. The plot functions are tested separately in test_viz_utils.py.

The outputsHandler deals with DDP synchronization for which we already have tests. The plots handler is meant to be executed from rank=0 only, not sure if we should test it in DDP because we'll have to rewrite the logic from outputs handler in the test....

ant0nsc (approved):

Issue #464, opened 2022-06-27
Read the Docs build fails due to new PyTorch Lightning requirements 


fepegar:
Extracted from https://readthedocs.org/projects/hi-ml/builds/17289566/

```text
ERROR: Cannot install pytorch-lightning<1.7 and >=1.6.0 and pytorch-lightning==1.5.5 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested pytorch-lightning<1.7 and >=1.6.0
    The user requested pytorch-lightning==1.5.5

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

fepegar:
@ozan-oktay: "Addresses" is not a valid keyword to automatically close an issue when merging a PR. More info here: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #466, opened 2022-06-27
ENH: Remove the dependency on Lightning and Lightning Bolts

ozan-oktay:
Addresses #464 

- Removes the dependency on pytorch-lightning and bolts.
- Wraps the torchvision resnet class to return the intermediate layer activations.

### DIFF
--- a/multimodal/health_multimodal/image/__init__.py
+++ b/multimodal/health_multimodal/image/__init__.py
@@ -21,6 +21,7 @@
 
    model
    modules
+   resnet
 """
 
 from .model.model import ImageModel, ResnetType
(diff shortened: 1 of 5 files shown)


fepegar:
It looks like there's a test that is still trying to import Bolts.

fepegar (changes requested):
A test is failing because it tries to import Bolts. I can work on fixing docstrings and ensuring the new classes are shown on the API docs.

ozan-oktay:
> It looks like there's a test that is still trying to import Bolts.

Yes, I can fix this one. Thanks.

ant0nsc (approved):

PR #224, opened 2022-03-08
ENH: Enforce prefixes for PR titles

ant0nsc:
Please follow the guidelines for PRs contained [here](docs/source/contributing.md). Checklist:

- [ ] Ensure that your PR is small, and implements one change.
- [ ] Add unit tests for all functions that you introduced or modified.
- [ ] Run PyCharm's code cleanup tools on your Python files.
- [ ] Ensure that documentation renders correctly in Sphinx (run Sphinx via `make html` in the `docs folder)
- [ ] Link the correct GitHub issue for tracking.
- [ ] When merging your PR, **replace the default merge message** with a brief description of your PR,
and if needed a motivation why that change was required.

PR #474, opened 2022-06-29
ENH: Enable pre-commit CI autofixes

fepegar:
I'm going to introduce a formatting issue next, to make sure the bot fixes it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,7 +3,6 @@
 # See https://github.com/pre-commit/pre-commit-hooks/blob/master/.pre-commit-config.yaml for an example with more hooks
 
 ci:
-  autofix_prs: false
   autoupdate_commit_msg: "ENH: Autoupdate hooks [pre-commit.ci]"
   autoupdate_schedule: monthly
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,7 @@
 # See https://github.com/pre-commit/pre-commit-hooks/blob/master/.pre-commit-config.yaml for an example with more hooks
 
 ci:
+  autofix_prs: true
   autoupdate_commit_msg: "ENH: Autoupdate hooks [pre-commit.ci]"
   autoupdate_schedule: monthly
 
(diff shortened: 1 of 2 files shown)


fepegar:
Looks like it's working.

.pre-commit-config.yaml
```diff
@@ -3,7 +3,6 @@
 # See https://github.com/pre-commit/pre-commit-hooks/blob/master/.pre-commit-config.yaml for an example with more hooks
 
 ci:
-  autofix_prs: false
   autoupdate_commit_msg: "ENH: Autoupdate hooks [pre-commit.ci]"
```
vale-salvatelli, line 6: if this is the commit that happens in the PR I don't think it should have ENH in front - maybe better something like '[pre-commit.ci] pre-commit fixes'?

fepegar: It's actually the default PR title for the autoupdates, not for commits. We can change the PR prefix to `MNT` after #224.

.pre-commit-config.yaml
```diff
@@ -3,7 +3,6 @@
 # See https://github.com/pre-commit/pre-commit-hooks/blob/master/.pre-commit-config.yaml for an example with more hooks
 
 ci:
-  autofix_prs: false
```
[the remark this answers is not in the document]
fepegar: Resolved in a08172d.

.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@
...
+  autofix_prs: true
   autoupdate_commit_msg: "ENH: Autoupdate hooks [pre-commit.ci]"
   autoupdate_schedule: monthly
```
ant0nsc, line 8: do we still need the monthly update?

mebristo (approved):

.pre-commit-config.yaml
fepegar: It's like Dependabot: it looks for newer versions of the hooks monthly (it doesn't run `pre-commit` monthly).

fepegar: So, we don't _need_ it, but I think it's a good feature to enable.

PR #484, opened 2022-07-01
ENH: Upgrade Histopathology folder to Python 3.10.4

ant0nsc:


ant0nsc:
Present blockers:
- `azureml-dataprep-runtime` package does not work with Python 3.10
- Even when we comment all those out, the builds presently pull in `hi-ml-azure` from PyPi when installing `hi-ml` in local editable mode. This will try to pull `azureml-dataprep-runtime` again, which will in turn fail.

One solution for the latter would be to simply skip `hi-ml-azure` in the `pip_local` target: Pip has a `no-dependencies` switch, and we can install the dependencies without `hi-ml-azure` manually.

ant0nsc:
Closing this for now, will re-activate once the packages are available for Python 3.10

PR #497, opened 2022-07-08
ENH: Prepare codebase for image model release

fepegar:
What I have done:

- Add some natural language explanations to the demo notebook ([see rendered demo here](https://github.com/microsoft/hi-ml/blob/60625fd93d7233d9258f4f4f00a2579d0f63bb54/hi-ml-multimodal/notebooks/phrase_grounding.ipynb))
- Harmonise methods to get inference engines in notebook: `get_cxr_bert_inference` and `get_cxr_resnet_inference` (maybe there is a better name for the latter)
- Add support to download the image weights from Hugging Face by default, instead of having to use a local path
- Set default interpolation mode (for the heatmap) to `nearest` but use `bilinear` for the notebook. This 1) makes the demo look better while 2) keeping potential benchmarks consistent with the paper
- Miscellaneous refactoring to make the above possible (and nice)

To do:

- [ ] Add URL of image model weights
- [ ] Add MD5 of image model weights


hi-ml-multimodal/src/health_multimodal/image/utils.py
```diff
@@ -0,0 +1,53 @@
...
+from pathlib import Path
+
+from torchvision.datasets.utils import download_url
+
+from .. import REPO_URL
+from .model import ImageModel
+from .inference_engine import ImageInferenceEngine
+from .data.transforms import create_chest_xray_transform_for_inference
+
+IMAGE_WEIGHTS_NAME = ""  # TODO
```
Shruthi42, line 16: Why isn't this in `health_multimodal/__init__.py` along with the text model?

hi-ml-multimodal/notebooks/phrase_grounding.ipynb
```diff
@@ -12,6 +12,32 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Phrase grounding\n",
+    "\n",
+    "This notebook demonstrates the multimodal models introduced in our ECCV 2022 paper:\n",
+    "\n",
+    "> Boecking, B., Usuyama, N., Bannur, S., Castro, D., Schwaighofer, A., Hyland, S., Wetscherek, M., Naumann, T., Nori, A., Alvarez-Valle, J., Poon, H., & Oktay, O. (2022). *Making the Most of Text Semantics to Improve Biomedical Vision–Language Processing* ([preprint](https://arxiv.org/abs/2204.09817))\n",
+    "\n",
+    "Given a chest X-ray and a text prompt, the joint model grounds the phrase in the image, i.e., highlights the regions of the image that share features similar to the phrase.\n",
```
ozan-oktay, line 25: "and a radiology text phrase"

hi-ml-multimodal/src/health_multimodal/image/utils.py
fepegar: I was having some circular imports issues and wasn't able to put it where I wanted: in `image/__init__.py`. I can try moving it up the tree, but I don't think it's too important. Feel free to commit a fix!

Shruthi42 (approved):

hi-ml-multimodal/notebooks/phrase_grounding.ipynb
```diff
@@ -59,64 +98,91 @@
...
     "    plot_phrase_grounding(image_path, text_prompt)"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Inference\n",
+    "\n",
+    "We will run inference on a chest X-ray from [Radiopaedia](https://radiopaedia.org/), but any can be used."
```
ozan-oktay, line 160: but any other chest-xray image in DICOM or JPEG format can be used for research purposes.

hi-ml-multimodal/notebooks/phrase_grounding.ipynb
```diff
@@ -12,6 +12,32 @@
...
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Phrase grounding\n",
+    "\n",
+    "This notebook demonstrates the multimodal models introduced in our ECCV 2022 paper:\n",
+    "\n",
+    "> Boecking, B., Usuyama, N., Bannur, S., Castro, D., Schwaighofer, A., Hyland, S., Wetscherek, M., Naumann, T., Nori, A., Alvarez-Valle, J., Poon, H., & Oktay, O. (2022). *Making the Most of Text Semantics to Improve Biomedical Vision–Language Processing* ([preprint](https://arxiv.org/abs/2204.09817))\n",
```
ozan-oktay, line 23: It might be good to emphasise that the notebook is solely for research evaluation purposes, not intended to be medical product or clinical use.

hi-ml-multimodal/src/health_multimodal/image/utils.py
fepegar: I've moved them in f7cd899. Thanks!

fepegar:
I've reported the failing CredScan CI check:
- https://github.com/microsoft/security-devops-action/issues/24

hi-ml-multimodal/src/health_multimodal/vlp/inference_engine.py
```diff
@@ -19,23 +19,19 @@
...
 
-    def get_similarity_map_from_raw_data(self, image_path: Path, query_text: str) -> np.ndarray:
-        """
-        Return a heatmap of the similarities between each patch embedding from the image and the text embedding.
-        """
+    def get_similarity_map_from_raw_data(self,
+                                         image_path: Path,
+                                         query_text: str,
+                                         interpolation: str = "nearest") -> np.ndarray:
+        """Return a heatmap of the similarities between each patch embedding from the image and the text embedding."""
```
ozan-oktay, line 34: can we update the documentation here? What kind of image formats are supported, max_text length (e.g., 512), and where this interpolation is actually used?

hi-ml-multimodal/src/health_multimodal/image/utils.py
```diff
@@ -0,0 +1,53 @@
...
+IMAGE_WEIGHTS_URL = f"{REPO_URL}/raw/main/{IMAGE_WEIGHTS_NAME}"
+IMAGE_WEIGHTS_MD5 = ""  # TODO
+
+
+def _download_image_model_weights() -> Path:
```
ozan-oktay, line 19: `_download_biovil_image_model_weights`
Let's specify exactly which model it is -- In the docstring we can add a URL to the entry page of our models.

hi-ml-multimodal/src/health_multimodal/image/utils.py
```diff
@@ -0,0 +1,53 @@
...
+    """Download image model weights from Hugging Face."""
+    root_dir = tempfile.gettempdir()
```
ozan-oktay, line 21: Shouldn't we place them in a dedicated directory instead of downloading into different place in each execution? I think that's how other libraries handle it, no?

hi-ml-multimodal/src/health_multimodal/image/utils.py
```diff
@@ -0,0 +1,53 @@
...
+    download_url(
+        IMAGE_WEIGHTS_URL,
+        root=tempfile.gettempdir(),
+        filename=IMAGE_WEIGHTS_NAME,
+        md5=IMAGE_WEIGHTS_MD5,
+    )
+    return Path(root_dir, IMAGE_WEIGHTS_NAME)
+
+
+def get_cxr_resnet(model_type: str = "resnet50", joint_feature_size: int = 128) -> ImageModel:
```
ozan-oktay, line 31: `get_biovil_resnet`

hi-ml-multimodal/src/health_multimodal/image/utils.py
```diff
@@ -0,0 +1,53 @@
...
+    """Download weights from Hugging Face and instantiate the image model."""
+    resnet_checkpoint_path = _download_image_model_weights()
+    image_model = ImageModel(
+        img_model_type=model_type,
```
ozan-oktay, line 35: The checkpoint is only for `resnet50` so perhaps there is not much point in parameterising it here because the feature size and architecture type are fixed. If any of them are changed, there will be a run-time error.

hi-ml-multimodal/src/health_multimodal/image/utils.py
fepegar: `tempfile.gettempdir()` is deterministic. For example, it's `/tmp` on Linux. I think you're thinking of `tempfile.mkdtemp`.

Another option, slightly more intrusive and maybe not necessary here, is using a cache dir in home. For example, `~/.cache/biovil`.

hi-ml-multimodal/src/health_multimodal/image/utils.py
```diff
@@ -0,0 +1,53 @@
...
+        joint_feature_size=joint_feature_size,
+        pretrained_model_path=resnet_checkpoint_path,
+    )
+    return image_model
+
+
+def get_cxr_resnet_inference(resize: int = 512, center_crop_size: int = 480) -> ImageInferenceEngine:
```
ozan-oktay, line 42: Again `get_biovil_resnet_inference`

hi-ml-multimodal/src/health_multimodal/vlp/inference_engine.py
fepegar: Is there a maximum text length?

fepegar: I've improved the docstring in b8715c8. Let me know if there is a maximum length and I'll add that.

ozan-oktay:
@fepegar : Here is the link to the image model checkpoint:
https://huggingface.co/microsoft/BiomedVLP-CXR-BERT-specialized/resolve/main/biovil_image_resnet50_proj_size_128.pt

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -56,8 +57,11 @@ def __init__(self,
         self.train()
 
         if pretrained_model_path is not None:
-            assert isinstance(pretrained_model_path, str), f"Expected a string, got {type(pretrained_model_path)}"
-            self.load_state_dict(torch.load(pretrained_model_path))
+            if not isinstance(pretrained_model_path, (str, Path)):
+                raise TypeError(f"Expected a string or Path, got {type(pretrained_model_path)}")
+            device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
ozan-oktay, line 62: I think we shouldn't make such assumptions. The functions role is just to load the weights, it's up to the application and outer context to place it in appropriate compute device (cuda:0, cuda:1, etc.)

fepegar: The problem is that the weights were saved on GPU, so this fails in a machine without one. I agree it would be up to the application to place it in an appropriate device for inference. Shall we default to always load to `"cpu"`?

ozan-oktay: I can update the weigths, give me 10 mins. We can push a change.

hi-ml-multimodal/src/health_multimodal/vlp/inference_engine.py
```diff
@@ -92,26 +98,30 @@ def convert_similarity_to_image_size(
         # TODO:
         # verify_resize_params(val_img_transforms, resize_size, crop_size)
 
+        reshaped_similarity = similarity_map.reshape(target_shape)
+        align_corners_modes = "linear", "bilinear", "bicubic", "trilinear"
+        align_corners = False if interpolation in align_corners_modes else None
```
ozan-oktay, line 103: Why don't we use "nearest" by default and let other auxiliary plotting functions to handle interpolation? I think model inference and plotting interpolation are two different concepts and they are mixed in here by design. This would reduce the code complexity here and improve the readability

ozan-oktay (changes requested):

hi-ml-multimodal/src/health_multimodal/vlp/inference_engine.py
fepegar: Nearest is already the default. The issue is that, if bilinear interpolation is passed, the `align_corners` kwargs should be set. Otherwise, a warning will be shown.

Maybe interpolation should not happen in this method. What do you suggest?

hi-ml-multimodal/src/health_multimodal/__init__.py
```diff
@@ -2,3 +2,21 @@
 #  Copyright (c) Microsoft Corporation. All rights reserved.
 #  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 #  -------------------------------------------------------------------------------------------
+
+BIOMED_VLP_CXR_BERT_SPECIALIZED = "microsoft/BiomedVLP-CXR-BERT-specialized"
+REPO_URL = f"https://huggingface.co/{BIOMED_VLP_CXR_BERT_SPECIALIZED}"
+CXR_BERT_COMMIT_TAG = "v1.0"
```
Shruthi42, line 8: We should probably pull the same version of the text and image models - maybe we can create a new tag for the image model release?

hi-ml-multimodal/src/health_multimodal/vlp/inference_engine.py
fepegar: [Agreed to potentially modify this in the future instead of here]

ozan-oktay: I agree with you regarding the last comment. We should handle it in the next PR.

hi-ml-multimodal/src/health_multimodal/image/model/model.py
ozan-oktay: Good idea -- discussed it offline.

ozan-oktay (approved):

hi-ml-multimodal/src/health_multimodal/__init__.py
ozan-oktay: @Shruthi42 : Thanks for raising this. We created a common tag for both models.

hi-ml-multimodal/src/health_multimodal/image/model/model.py
fepegar: In https://github.com/microsoft/hi-ml/pull/497/commits/30fb7ff62a05b9de5275ff76802c7f7effc5df33:

- Use `"cpu"` as map location when loading weights
- Add `to` method to joint inference engine and `to` attribute to the image and text engines
- Automatically select the device in the notebook and call the new method
- In the inference engines, choose device from model at inference time, not at instantiation time

hi-ml-multimodal/src/health_multimodal/__init__.py
fepegar: @ozan-oktay has created `v1.1`. I'll update in a minute.

Issue #508, opened 2022-07-14
plot_normalized_confusion_matrix fails if predicted and true labels don't include all expected classes


mebristo:
Currently if the true and predicted labels in validation results don't include all possible class labels, an error is raised when trying to plot a heatmap of the confusion matrix, as we expect the number of ticks to equal the number of possible class labels. This can be solved by specifying the expected labels when the confusion matrix is created.

PR #509, opened 2022-07-14
BUG: Plot conf matrix when not all labels present

mebristo:
Closes [this issue](https://github.com/microsoft/hi-ml/issues/508)

### DIFF
--- a/hi-ml-histopathology/src/histopathology/utils/plots_utils.py
+++ b/hi-ml-histopathology/src/histopathology/utils/plots_utils.py
@@ -55,7 +55,7 @@ def save_confusion_matrix(results: ResultsType, class_names: Sequence[str], figu
     """
     true_labels = [i.item() if isinstance(i, Tensor) else i for i in results[ResultsKey.TRUE_LABEL]]
     pred_labels = [i.item() if isinstance(i, Tensor) else i for i in results[ResultsKey.PRED_LABEL]]
-    all_potential_labels = [i for i in range(len(class_names))]
+    all_potential_labels = list(range(len(class_names)))
     true_labels_diff_expected = set(true_labels).difference(set(all_potential_labels))
     pred_labels_diff_expected = set(pred_labels).difference(set(all_potential_labels))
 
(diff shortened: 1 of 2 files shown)


hi-ml-histopathology/src/histopathology/utils/plots_utils.py
```diff
@@ -51,7 +53,18 @@ def save_confusion_matrix(results: ResultsType, class_names: Sequence[str], figu
     :param class_names: List of class names.
     :param figures_dir: The path to the directory where to save the confusion matrix.
     """
-    cf_matrix_n = confusion_matrix(results[ResultsKey.TRUE_LABEL], results[ResultsKey.PRED_LABEL], normalize="pred")
+    true_labels = [i.item() if isinstance(i, Tensor) else i for i in results[ResultsKey.TRUE_LABEL]]
+    pred_labels = [i.item() if isinstance(i, Tensor) else i for i in results[ResultsKey.PRED_LABEL]]
+    all_potential_labels = [i for i in range(len(class_names))]
+    present_labels_diff_expected = set(true_labels).union(set(pred_labels)).difference(set(all_potential_labels))
```
ant0nsc, line 59: I think it would be better to split this into two checks, one for true and one for predicted labels, so that the error message can be more specific?

hi-ml-histopathology/testhisto/testhisto/utils/test_plots_utils.py
```diff
@@ -88,11 +89,43 @@ def test_plots_handler_plots_only_desired_plot_options(plot_options: Collection[
...
+
     save_confusion_matrix(results, class_names, tmp_path)
     file = Path(tmp_path) / "normalized_confusion_matrix.png"
     assert file.exists()
+
+    # check that an error is raised if true or predicted labels include indices greater than the expected number of
+    # classes
+    invalid_results_1 = {
+        ResultsKey.TRUE_LABEL: [0, 1, 0, 1, 0, 2],
+        ResultsKey.PRED_LABEL: [0, 1, 0, 0, 0, 1]
```
kenza-bouzid, line 108: I am not sure this is actually an error. The set of true label might be {0, 1, 2} but the model could have predicted false labels that are only in {0, 1}, I wouldn't consider that an error especially when we use a limited number of val/test batches that might not include all possible labels...

mebristo: The error here is that the code will try to index class_names at index 2, which is out of range

kenza-bouzid (approved):

Issue #501, opened 2022-07-12
Publish a package for hi-ml-histopathology


ant0nsc:
- [ ] Rename the histo folder to be in line with the other packages
- [ ] Set up a package publishing workflow

kenza-bouzid:
@ant0nsc what's the status on this?

ant0nsc:
This is done. hi-ml-cpath is published with each tag creation.

PR #472, opened 2022-06-28
ENH: Add CI job to publish multimodal package

fepegar:
Resolves #461.

PR #510, opened 2022-07-14
ENH: Simplify publishing workflow by using a single PyPi token

ant0nsc:
Rather than having 2 separate publishing tokens, use one with scope "all projects".

This is setting the stage for #501, and will simplify #472


peterhessey (approved):

fepegar (approved):
Awesome, thanks!

PR #518, opened 2022-07-18
ENH: Enable running inference-only with the hi-ml runner

### LINKED ISSUE #480: Add ability to load a pre-trained model for resuming training or inference
Should be able to reference an existing AML run and point to a specific checkpoint.

kenza-bouzid:
Closes #480 
* Add `run_inference_only` flag to trigger trainer.test only
* Add `src_checkpoint` to specify the azure ml run id from which to retrieve the weights checkpoint for inference or transfer learning
* Rename `ckpt_run_id` -> `ssl_ckpt_run_id` in BasEMIL container to differentiate between different checkpoint run_ids


hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -41,7 +41,7 @@ def __init__(self,
                  class_names: Optional[Sequence[str]] = None,
                  dropout_rate: Optional[float] = None,
                  verbose: bool = False,
-                 ckpt_run_id: Optional[str] = None,
+                 ssl_ckpt_run_id: Optional[str] = None,
```
mebristo, line 44: Why is this necessarily SSL ckpt? Could it not also be one of the imagenet encoders?

kenza-bouzid: Because the only weights we load from AML runs are SSL checkpoints, it's used here [encoder](https://github.com/microsoft/hi-ml/blob/d7c5ebf32f869cf56b898fc2150be6965f16f8c1/hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py#L75) to create the SSLEncoder.
It can definitely be an imagenet encoder but we don't need an aml_run_id for these.

docs/source/runner.md
```diff
@@ -210,3 +210,23 @@ class MyContainer(LightningContainer):
 To the optimizer and LR scheduler: the Lightning model returned by `create_model` should define its own
 `configure_optimizers` method, with the same signature as `LightningModule.configure_optimizers`,
 and returns a tuple containing the Optimizer and LRScheduler objects
+
+## Run inference with a pretrained model
+
+You can use the hi-ml-runner in inference mode only by switching the `--run_inference` flag on and specifying
```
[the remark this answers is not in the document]
mebristo: Same on line 222

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -153,18 +155,24 @@ class WorkflowParams(param.Parameterized):
                      doc="When comparing CSV files during regression tests, use this value as the maximum allowed "
                          "relative difference of actual and expected results. Default: 0.0 (must match exactly)")
     regression_metrics: str = param.String(default=None, doc="A list of names of metrics to compare")
+    run_inference_only: bool = param.Boolean(False, doc="If True, run inference using the specified ckpt_run_id as "
```
mebristo, line 158: What about the weights_url or local_weights_path?

hi-ml/src/health_ml/utils/checkpoint_handler.py
```diff
@@ -126,6 +126,10 @@ def get_local_checkpoints_path_or_download(self) -> Path:
             download_folder.mkdir(exist_ok=True, parents=True)
             checkpoint_path = CheckpointHandler.download_weights(url=self.container.weights_url,
                                                                  download_folder=download_folder)
+        elif self.container.ckpt_run_id:
+            downloader = get_checkpoint_downloader(self.container.ckpt_run_id, self.container.outputs_folder)
+            checkpoint_path = downloader.local_checkpoint_path
+            assert checkpoint_path.is_file(), f"Couln't download checkpoint from run {self.container.ckpt_run_id}."
         else:
             raise ValueError("Cannot download weights, neither local_weights_path or weights_url are set")
```
mebristo, line 134: This error message needs to inlcude ckpt_run_id

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -1,21 +1,37 @@
...
+from health_azure.utils import is_global_rank_zero, create_aml_run_object
+from health_azure.utils import WORKSPACE_CONFIG_JSON, check_config_json, get_workspace
+from testazure.utils_testazure import DEFAULT_WORKSPACE
+from testhiml.utils.fixed_paths_for_tests import full_test_data_path
+from testazure.utils_testazure import get_shared_config_json
 
 no_gpu = not is_gpu_available()
 
 
+def _get_workspace(tmp_path: Path) -> Workspace:
```
mebristo, line 24: I'm not sure why this is necessary? lines 30-31 work on a local machine too

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -204,19 +252,78 @@ def _expected_files_exist() -> bool:
...
-def test_run(ml_runner_with_container: MLRunner) -> None:
+@pytest.mark.parametrize("run_extra_val_epoch", [True, False])
+@pytest.mark.parametrize("run_inference_only", [True, False])
+def test_run(run_inference_only: bool, run_extra_val_epoch: bool, ml_runner_with_container: MLRunner) -> None:
     """Test that model runner gets called """
+    ml_runner_with_container.container.run_inference_only = run_inference_only
+    ml_runner_with_container.container.run_extra_val_epoch = run_extra_val_epoch
     ml_runner_with_container.setup()
     assert not ml_runner_with_container.checkpoint_handler.has_continued_training
     with patch.object(ml_runner_with_container, "checkpoint_handler"):
```
mebristo, line 263: You may find it clearer to use [patch.multiple](https://docs.python.org/3/library/unittest.mock.html#patch-multiple) here for mocking many different methods of the same object

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -204,19 +252,78 @@ def _expected_files_exist() -> bool:
...
+    with patch("health_azure.utils.get_workspace") as mock_get_workspace:
+        mock_get_workspace.return_value = _get_workspace(tmp_path)
+        runner = MLRunner(experiment_config=experiment_config, container=container)
+        runner.setup()
+        weights_before: torch.Tensor = container.model.model.weight.detach().clone()  # type: ignore
+        with patch.object(runner, "run_validation"):
+            with patch.object(runner, "run_inference"):
+                runner.run()
+        weights_after: torch.Tensor = runner.container.model.model.weight  # type: ignore
+        assert not torch.allclose(weights_before, weights_after)
```
mebristo, line 329: I'm not convinced we're asserting the right thing here, given the test name. It makes sense when `run_inference_only=True` that the only way that the weights have been updated is because `load_model_checkpoint` was called - but when `run_inference_only=False` then training occurs so the weights will update regardless of whether `load_model_checkpoint` gets called. Does it make more sense to mock `load_model_checkpoint` and assert that it has been called?

docs/source/runner.md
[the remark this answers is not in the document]
kenza-bouzid: Good catch!

hi-ml/src/health_ml/deep_learning_config.py
kenza-bouzid: Good point, will update the doc.

hi-ml/src/health_ml/utils/checkpoint_handler.py
kenza-bouzid: Sure

hi-ml/src/health_ml/utils/checkpoint_handler.py
```diff
@@ -126,6 +126,10 @@ def get_local_checkpoints_path_or_download(self) -> Path:
             download_folder.mkdir(exist_ok=True, parents=True)
             checkpoint_path = CheckpointHandler.download_weights(url=self.container.weights_url,
                                                                  download_folder=download_folder)
+        elif self.container.ckpt_run_id:
+            downloader = get_checkpoint_downloader(self.container.ckpt_run_id, self.container.outputs_folder)
+            checkpoint_path = downloader.local_checkpoint_path
+            assert checkpoint_path.is_file(), f"Couln't download checkpoint from run {self.container.ckpt_run_id}."
```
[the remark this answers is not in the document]
kenza-bouzid: Good point!

hi-ml/testhiml/testhiml/test_run_ml.py
kenza-bouzid: Definitely, but you need to set the environment variables for the service principal. Not everyone has these set especially new joiners, whereas everyone has a config.json at the repo root for other purposes.  It's just to make it easier to debug the tests without having to go through all the service principal setup

hi-ml/testhiml/testhiml/test_run_ml.py
kenza-bouzid: `load_model_checkpoint`[ is not called systematically for all containers](https://github.com/microsoft/hi-ml/blob/b66b42c7ed4c67b4b6b385b0bdf5862dcedde92d/hi-ml/src/health_ml/run_ml.py#L337). if `test_step `is not specified, we don't run inference nor load the model checkpoint unless we activate the extra validation epoch...
 
The HelloRegression Lightning Module does not override `test_step `so `load_model_checkpoint` won't be called by default. 
the container.model weights will neverthless change by the trainer but won't necessarily correspond to the "best checkpoint" (it's usually the last checkpoint by default)

Here I just want to make sure that we're using the right weights in 2 scenarios:
run_inference_only=True: all training will be skipped, and I want to make sure that inference is run with the loaded checkpoint (weights_before != weights_after) and here `load_model_checkpoint `was indeed called
run_inference_only=False: but a ckpt_run_id is provided so we want to resume training from there, however I can't really test that because it's passed to the trainer through `resume_from_checkpoint` and that would be testing external code... however I still want to make sure that the weights have changed at this point, here `load_model_checkpoint ` is not called...

Does it make sense? Happy to follow up offline

hi-ml/testhiml/testhiml/test_run_ml.py
ant0nsc: We have code like that already. For unittests, just use `DEFAULT_WORKSPACE.workspace`, this has all the smarts already built in.

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -48,6 +64,38 @@ def ml_runner_with_container() -> Generator:
         shutil.rmtree(output_dir)
 
 
+@pytest.fixture(scope="module")
```
ant0nsc, line 67: I think if you re-design your tests, you could save on some code. You could separate them out into two parts:
- First testing that all of the various ways of specifying a checkpoint give you a valid local file (effectively, testing `get_local_checkpoints_path_or_download`)
- Once you know that you can get a valid local file with all 3 options, you only need to test if the runner correctly picks up the local weights.

docs/source/runner.md
```diff
@@ -210,3 +210,23 @@ class MyContainer(LightningContainer):
 To the optimizer and LR scheduler: the Lightning model returned by `create_model` should define its own
 `configure_optimizers` method, with the same signature as `LightningModule.configure_optimizers`,
 and returns a tuple containing the Optimizer and LRScheduler objects
+
+## Run inference with a pretrained model
+
+You can use the hi-ml-runner in inference mode only by switching the `--run_inference_only` flag on and specifying
+the model weights by setting any of those arguments `--weights_url`, `--local_weights_path` or `--ckpt_run_id`.
```
ant0nsc, line 217: Not sure I wrote that elsewhere: the naming does not look very consistent (weights/weights/ckpt). Also, how common is the abbreviation `ckpt`? `checkpoint_from_run` would be longer, but reads nicer. `checkpoint_url` and `local_checkpoint`?

ant0nsc (commented):
Looking good, overall! I think your recent refactoring has also paid off. I've left a couple of comments.

docs/source/runner.md
kenza-bouzid: Sounds great, will rename them!

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -153,18 +155,24 @@ class WorkflowParams(param.Parameterized):
...
-        if self.weights_url and self.local_weights_path:
-            raise ValueError("Cannot specify more than one of local_weights_path, weights_url.")
+        if sum([bool(self.weights_url), bool(self.local_weights_path), bool(self.ckpt_run_id)]) > 1:
+            raise ValueError("Cannot specify more than one of local_weights_path, weights_url, ckpt_run_id.")
 
         if self.crossval_count > 1:
             if not (0 <= self.crossval_index < self.crossval_count):
                 raise ValueError(f"Attribute crossval_index out of bounds (crossval_count = {self.crossval_count})")
 
+        if self.run_inference_only and not (self.weights_url or self.local_weights_path or self.ckpt_run_id):
```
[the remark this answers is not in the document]
kenza-bouzid: I don't think we can merge the two checks as it's two different kind of errors? however I like your suggestion about precomputing the sum.

hi-ml/src/health_ml/configs/hello_world.py
```diff
@@ -257,3 +260,13 @@ def get_data_module(self) -> LightningDataModule:
...
+
+    def get_callbacks(self) -> List[Callback]:
+        if self.save_checkpoint:
+            return [ModelCheckpoint(dirpath=self.checkpoint_folder,
+                                    monitor="val_loss",
+                                    filename="checkpoint",
+                                    auto_insert_metric_name=False,
+                                    mode="min")]
+        else:
+            return super().get_callbacks()
```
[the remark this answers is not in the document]
kenza-bouzid: Good point!

docs/source/runner.md
```diff
@@ -210,3 +210,23 @@ class MyContainer(LightningContainer):
 To the optimizer and LR scheduler: the Lightning model returned by `create_model` should define its own
 `configure_optimizers` method, with the same signature as `LightningModule.configure_optimizers`,
 and returns a tuple containing the Optimizer and LRScheduler objects
+
+## Run inference with a pretrained model
+
+You can use the hi-ml-runner in inference mode only by switching the `--run_inference` flag on and specifying
+the model weights by setting any of those arguments `--weights_url`, `--local_weights_path` or `--ckpt_run_id`.
```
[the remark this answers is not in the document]
kenza-bouzid: renamed args!

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -673,6 +673,26 @@ def download_checkpoint_if_necessary(self) -> Path:
...
+def get_checkpoint_downloader(ckpt_run_id: str, outputs_folder: Path) -> CheckpointDownloader:
+    """Creates an instance of CheckpointDownloader for the specified run_id and downloads the corresponding checkpoint.
+
+    :param ckpt_run_id: The Azure ML run_id of the run from which to load the checkpoint.
+    :param outputs_folder: The outputs folder where to download the checkpoint.
+    :return: A downloader instance for the specified run_id.
+    """
+    from health_ml.utils.common_utils import CHECKPOINT_FOLDER, DEFAULT_AML_UPLOAD_DIR
+    from health_ml.utils.checkpoint_utils import LAST_CHECKPOINT_FILE_NAME_WITH_SUFFIX
+    downloader = CheckpointDownloader(
```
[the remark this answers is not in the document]
kenza-bouzid: That was used in cpath to load ssl checkpoints, I didn't want to duplicate it so I moved it to health_azure, looking at you comment below, It might not be the best solution.

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -673,6 +673,26 @@ def download_checkpoint_if_necessary(self) -> Path:
...
+        aml_workspace=get_workspace(),
+        run_id=ckpt_run_id,
+        checkpoint_filename=LAST_CHECKPOINT_FILE_NAME_WITH_SUFFIX,
```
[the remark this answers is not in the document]
kenza-bouzid: That is a very good point, I will make that possible!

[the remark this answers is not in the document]
mebristo: @ant0nsc I'm not sure it makes sense to provide a checkpoint path under the `run_id` flag. Does it make more sense to have a separate argument for 'checkpoint_path'? (Which actually I think we should add in any case, since we currently assume the checkpoints are uploaded to a folder called 'outputs' which might not be the case for models that weren't trained with our pipeline)

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -127,11 +129,20 @@ class WorkflowParams(param.Parameterized):
...
+                                       doc="This flag can be used for 3 different scenarios:"
+                                           "1- Resume training from a checkpoint to train longer."
+                                           "2- Run inference only by jointly using `run_inference_only` flag."
+                                           "3- Transfer learning from a pretrained model checkpoint."
+                                           "We currently support Three types of checkpoints: "
+                                           "    a. A local checkpoint folder that contains a checkpoint file."
+                                           "    b. A URL to a remote checkpoint to be downloaded."
+                                           "    c. A previous azureml run id where the checkpoint is supposed to be "
+                                           "       saved in 'outputs/checkpoints/' folder.")
+    src_checkpoint_filename: str = param.String(default="last.ckpt",
```
ant0nsc, line 142: I would add that to the Azure run ID via a `:`.
Also, we probably want to be more flexible with where the checkpoint can live. For example, if I just say `foo:last.ckpt` it will search for it in `outputs/checkpoints`. But I should also be able to say `foo:some/other/path.ckpt`

docs/source/runner.md
```diff
@@ -210,3 +210,34 @@ class MyContainer(LightningContainer):
...
+
+You can use the hi-ml-runner in inference mode only by switching the `--run_inference_only` flag on and specifying
+the model weights by setting `--src_checkpoint` argument that supports three types of checkpoints:
+
+* A local path where the checkpoint is stored `--src_checkpoint=local/path/to/my_checkpoint/model.ckpt`
+* A remote URL from where to download the weights `--src_checkpoint=https://my_checkpoint_url.com/model.ckpt`
+* An azureml run id where checkpoints are saved in `outputs/checkpoints`. For this specific use case, you can experiment
+  with different checkpoints by setting `--src_checkpoint` according to the format
+  `<MyContainer_xxx_yyy>:<optional/custom/path/to/checkpoints/><filename.ckpt>`. If no custom path is provided
+  (e.g., `--src_checkpoint=MyContainer_xxx_yyy:best.ckpt`) we assume the checkpoints to be saved in the default
```
kenza-bouzid, line 224:
```suggestion
  (e.g., `--src_checkpoint=MyContainer_xxx_yyy:best.ckpt`), we assume the checkpoints to be saved in the default
```

docs/source/runner.md
```diff
@@ -210,3 +210,34 @@ class MyContainer(LightningContainer):
...
+  checkpoints folder `outputs/checkpoints`. If no filename is provided (e.g., `--src_checkpoint=MyContainer_xxx_yyy`)
```
kenza-bouzid, line 225:
```suggestion
  checkpoints folder `outputs/checkpoints`. If no filename is provided (e.g., `--src_checkpoint=MyContainer_xxx_yyy`),
```

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -127,11 +128,23 @@ class WorkflowParams(param.Parameterized):
     This class contains all parameters that affect how the whole training and testing workflow is executed.
     """
     random_seed: int = param.Integer(42, doc="The seed to use for all random number generators.")
-    weights_url: str = param.String(default="",
-                                    doc="If provided, use this URL to download checkpoints for inference.")
-    local_weights_path: Optional[Path] = \
-        param.ClassSelector(class_=Path, allow_None=True, default=None,
-                            doc="Checkpoint paths to use for inference, when the job is running outside Azure.")
+    src_checkpoint: str = param.String(default="",
+                                       doc="This flag can be used for 3 different scenarios:"
```
kenza-bouzid, line 132:
```suggestion
                                       doc="This flag can be used in 3 different scenarios:"
```

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -127,11 +128,23 @@ class WorkflowParams(param.Parameterized):
...
+                                           "1- Resume training from a checkpoint to train longer."
+                                           "2- Run inference-only using `run_inference_only` flag jointly."
+                                           "3- Transfer learning from a pretrained model checkpoint."
+                                           "We currently support three types of checkpoints: "
+                                           "    a. A local checkpoint folder that contains a checkpoint file."
+                                           "    b. A URL to a remote checkpoint to be downloaded."
+                                           "    c. A previous azureml run id where the checkpoint is supposed to be "
+                                           "       saved ('outputs/checkpoints/' folder by default.)"
+                                           "For the 'c' case: src_checkpoint should be in the format of "
```
kenza-bouzid, line 141:
```suggestion
                                           "For the latter case 'c' : src_checkpoint should be in the format of "
```

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -127,11 +128,23 @@ class WorkflowParams(param.Parameterized):
...
+                                           "<MyContainer_xxx_yyy>:<optional/custom/path/to/checkpoints/><filename.ckpt>"
+                                           "If no custom path is provide (e.g., <MyContainer_xxx_yyy>:<filename.ckpt>)"
```
kenza-bouzid, line 143:
```suggestion
                                           "If no custom path is provided (e.g., <MyContainer_xxx_yyy>:<filename.ckpt>)"
```

hi-ml/src/health_ml/utils/checkpoint_utils.py
```diff
@@ -157,3 +158,76 @@ def cleanup_checkpoints(ckpt_folder: Path) -> None:
...
+        Extracts the checkpoint filename from the run_id if run_id is in the format
+        <MyContainer_xx>:<checkpoint_filename.ckpt>. Otherwise, uses the last epoch checkpoint filename as default.
+        """
+        run_id_split = self.run_id.split(":")
+        self.run_id = run_id_split[0]
+        return run_id_split[-1] if len(run_id_split) > 1 else LAST_CHECKPOINT_FILE_NAME_WITH_SUFFIX
+
+    def extract_remote_checkpoint_dir_from_checkpoint_filename(self) -> Path:
+        """
+        Extracts the remote checkpoint directory from the run_id if run_id is in the format
```
kenza-bouzid, line 199:
```suggestion
        Extracts the remote checkpoint directory from the checkpoint_filename if checkpoint_filename is in the format
```

hi-ml/src/health_ml/deep_learning_config.py
kenza-bouzid: Done!

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -153,18 +166,52 @@ class WorkflowParams(param.Parameterized):
...
+        except ValueError:
+            return False
+
+    @property
+    def src_checkpoint_is_local_file(self) -> bool:
+        return Path(self.src_checkpoint).exists()
 
+    @property
+    def src_checkpoint_is_aml_run_id(self) -> bool:
+        return re.match(r"[_\w-]*$", self.src_checkpoint.split(":")[0]) is not None
```
kenza-bouzid, line 190: Note: I support dashes "-" as well because when I created mock aml runs they had ids in this format `699e5a65-8f36-451d-b955-594fcaf6906e`,

ant0nsc (approved):
Great work! Left a few minor typo corrections, good to merge :-)

hi-ml/src/health_ml/deep_learning_config.py
mebristo, line 190: Should we also include the conditions that `src_checkpoint_is_local_file` and  `src_checkpoint_is_url` be False in this property? Otherwise a URL such as 'https://www.microsoft.com' could satisfy both `src_checkpoint_is_url` and `src_checkpoint_is_aml_run_id`

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -153,18 +166,52 @@ class WorkflowParams(param.Parameterized):
...
+            raise ValueError(f"Invalid src_checkpoint: {self.src_checkpoint}. Please provide a valid URL, local file "
+                             "or azureml run id.")
         if self.crossval_count > 1:
             if not (0 <= self.crossval_index < self.crossval_count):
                 raise ValueError(f"Attribute crossval_index out of bounds (crossval_count = {self.crossval_count})")
 
+        if self.run_inference_only and not self.src_checkpoint:
+            raise ValueError("Cannot run inference without a src_checkpoint. Please specify a valid src_checkpoint."
+                             "You can either use a URL, a local file or an azureml run id. For custom checkpoint paths "
+                             "within an azureml run, (other than last.ckpt), provide a src_checkpoint in the format."
```
mebristo, line 209: This same message string exists in a few different places. It might be better to store it in a variable so it is easier to update in future

hi-ml/src/health_ml/utils/checkpoint_handler.py
```diff
@@ -35,15 +36,12 @@ def __init__(self,
 
     def download_recovery_checkpoints_or_weights(self) -> None:
         """
-        Download checkpoints from a run recovery object or from a weights url. Set the checkpoints path based on the
-        run_recovery_object, weights_url or local_weights_path.
+        Download checkpoints from a run recovery object or from a given checkpoint. Set the checkpoints path based on
+        the run_recovery_object, the checkpoint_url, local_checkpoint or checkpoint from an azureml run id.
```
mebristo, line 40: I think we can remove run_recovery_object from here now?

hi-ml/src/health_ml/utils/checkpoint_handler.py
```diff
@@ -119,15 +117,19 @@ def get_local_checkpoints_path_or_download(self) -> Path:
...
-                                                                 download_folder=download_folder)
+            checkpoint_path = self.download_weights(url=self.container.src_checkpoint, download_folder=download_folder)
+        elif self.container.src_checkpoint_is_aml_run_id:
+            downloader = CheckpointDownloader(
+                run_id=self.container.src_checkpoint, download_dir=self.container.outputs_folder
+            )
+            checkpoint_path = downloader.local_checkpoint_path
         else:
-            raise ValueError("Cannot download weights, neither local_weights_path or weights_url are set")
+            raise ValueError("Cannot download weights, src_checkpoint is not set.")
```
mebristo, line 132: The error message is out of sync now - it could be the case that `src_checkpoint` is set but that it doesn't satisfy any of the conditions (is_url, is_local_file, is_aml_run_id)

hi-ml-cpath/testhisto/conftest.py
```diff
@@ -25,6 +25,9 @@
 himl_package_root = himl_root / "hi-ml" / "src"
 logging.info(f"Adding {str(himl_package_root)} to path")
 sys.path.insert(0, str(himl_package_root))
+test_himl_package_root = himl_root / "hi-ml" / "testhiml"
+logging.info(f"Adding {str(test_himl_package_root)} to path")
+sys.path.insert(0, str(test_himl_package_root))
 himl_azure_package_root = himl_root / "hi-ml-azure" / "src"
 logging.info(f"Adding {str(himl_azure_package_root)} to path")
 sys.path.insert(0, str(himl_azure_package_root))
```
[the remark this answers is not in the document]
kenza-bouzid: Done

hi-ml/src/health_ml/deep_learning_config.py
kenza-bouzid: Good point!

mebristo (approved):

PR #523, opened 2022-07-19
ENH: Move SSL code into the health_cpath namespace

ant0nsc:
Moving all files from `SSL` to `health_cpath/SSL`. Removed the `classification` folder from the configs folder hierarchy. Moved all SSL configs into the `health_cpath/SSL/configs` folder


ant0nsc:
Closing because SSL code is not yet needed in the package.

PR #524, opened 2022-07-19
ENH: Publishing the hi-ml-cpath package

ant0nsc:
Moved the SSL configurations out of the `health_cpath` folder because they reference code that is not published in the package.


hi-ml-cpath/setup.py
```diff
@@ -0,0 +1,89 @@
...
+package_name = 'hi-ml-cpath'
+(here / 'package_name.txt').write_text(package_name)
+(here / 'latest_version.txt').write_text(version)
+
+# Read run_requirements.txt to get install_requires
+install_requires = (here / 'requirements_run.txt').read_text().split("\n")
+# Remove any whitespace and blank lines
+install_requires = [line.strip() for line in install_requires if line.strip()]
+
+description = 'Microsoft Health Futures package for computational pathology data'
```
kenza-bouzid, line 65: Can we improve this description? "Computational pathology data" only might be a bit misleading: is it just data processing/analytics? It would be nice to mention that it's a machine learning toolbox for computational pathology?

hi-ml-cpath/setup.py
```diff
@@ -0,0 +1,89 @@
...
+    author="Microsoft Research Cambridge Medical Imaging Team ",
+    author_email="innereyedev@microsoft.com",
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Intended Audience :: Science/Research',
+        "Topic :: Scientific/Engineering :: Medical Science Apps.",
+        'License :: OSI Approved :: MIT License',
+        'Programming Language :: Python :: 3.7'
+    ],
+    keywords='HealthIntelligence, Computational Pathology, AzureML',
```
kenza-bouzid, line 83: Maybe add Health Futures?

kenza-bouzid (approved):
Looking good, left minor suggestions. 
I have a quick question on how the SSL code should be used now in hi-ml-cpath, can we still import from SSL in health_cpath?  Can you please clarify what the hi-ml-cpath package comprises? Can I still import a config file that has an SSL import 
from the pip installed version of health_cpath?

hi-ml-cpath/environment.yml
```diff
@@ -126,8 +126,10 @@ dependencies:
     - google-auth==1.35.0
     - google-auth-oauthlib==0.4.6
     - grpcio==1.47.0
+    - hi-ml==0.2.3
```
mebristo, line 129: If we include this pinned version of hi-ml as a requirement, will there be any problems in the build pipeline, when we initially create the python environment with this pinned package, then call pip locally inside the hi-ml module?

mebristo: (and same for hi-ml-azure)

hi-ml-cpath/requirements_run.txt
```diff
@@ -1,6 +1,7 @@
 coloredlogs==15.0.1
 cucim==22.04.00
 girder-client==3.1.14
+hi-ml>=0.2.3
```
mebristo, line 4: Is there a reason this is lower bounded here but pinned in environment.yml?

hi-ml-cpath/src/health_cpath/datasets/panda_tiles_dataset.py
```diff
@@ -116,7 +116,7 @@ def __len__(self) -> int:
         return len(self.base_dataset)
 
 
-class PandaTilesDatasetWithReturnIndex(DataClassBaseWithReturnIndex, PandaTilesDatasetReturnImageLabel):
+class PandaTilesDatasetWithReturnIndex(DatasetWithReturnIndex, PandaTilesDatasetReturnImageLabel):
     """
     Any dataset used in SSL needs to inherit from DataClassBaseWithReturnIndex as well as VisionData.
```
mebristo, line 121: Remaining reference to `DataClassBaseWithReturnIndex` here (and in `tcga_crck_tiles_dataset.py` too and possibly in other files)

hi-ml-cpath/src/health_cpath/models/encoders.py
```diff
@@ -112,6 +109,12 @@ def __init__(self, pl_checkpoint_path: Path, tile_size: int, n_channels: int = 3
         super().__init__(tile_size=tile_size, n_channels=n_channels)
 
     def _get_encoder(self) -> Tuple[torch.nn.Module, int]:
+        try:
+            from SSL.lightning_modules.ssl_classifier_module import SSLClassifier
+            from SSL.utils import create_ssl_image_classifier
+            from SSL import encoders
+        except (ImportError, ModuleNotFoundError):
+            raise ValueError("SSL not found. This class can only be used by using hi-ml from the GitHub source")
```
mebristo, line 117: On this note - there's a line in `docs/source/histopathology.md` that states 'The tools for computational pathology cannot be as a Python package, but rather directly from the Git repository.'  which I think it would make sense to update in this PR

hi-ml-cpath/environment.yml
ant0nsc: Yes, I checked that this works. If `hi-ml` is already installed, a next attempt to install `hi-ml` from local in editable mode will unistall the existing, and put the local version on top.
An alternative way would have been to exclude `hi-ml` when we build the locked environment.

hi-ml-cpath/requirements_run.txt
ant0nsc: `environment.yml` is a locked environment, similar to what we have now in IE-DL: all dependencies and secondary are pinned. To update, we use a script - see header at the top of the env file.

hi-ml-cpath/src/health_cpath/datasets/panda_tiles_dataset.py
ant0nsc: Thanks, good find, I did not check documentation, but relied on automatic refactoring...

mebristo (approved):

PR #531, opened 2022-07-21
BUG: Clear the top and bottom slides heaps after every validation epoch

mebristo:
Clear the top and bottom slides heaps after every validation epoch. Otherwise `save_test_outputs` the slides in the results do not match those stored in the top and bottom slides heaps (which are only updated in the validation step)

### DIFF
--- a/hi-ml-cpath/src/health_cpath/utils/tiles_selection_utils.py
+++ b/hi-ml-cpath/src/health_cpath/utils/tiles_selection_utils.py
@@ -120,6 +120,10 @@ class TilesSelector:
         report_cases.update({f"FN_{class_id}": [] for class_id in range(1, self.n_classes)})
         return report_cases
 
+    def _clear_cached_slides_heaps(self):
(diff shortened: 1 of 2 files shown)


kenza-bouzid (approved):
LGTM, thanks for catching and fixing this bug.

hi-ml-cpath/src/health_cpath/utils/tiles_selection_utils.py
```diff
@@ -120,6 +120,10 @@ def init_report_cases(self) -> Dict[str, List[str]]:
         report_cases.update({f"FN_{class_id}": [] for class_id in range(1, self.n_classes)})
         return report_cases
 
+    def _clear_cached_slides_heaps(self) -> None:
```
ant0nsc, line 123: Can I suggest to apply DRY? there are now 4 copies of the same code snippet, in lines 107, 108, 124, 125. How we initialize and how we clear can hence run out of sync.

ant0nsc (approved):
Looks good - it would be good though to ensure that code can't run out of sync, see my comment.

PR #536, opened 2022-07-26
ENH: Allow skipping of top/ bottom tile plotting

mebristo:
At the moment training can fail if not enough tiles are available to plot top and bottom tiles. This PR updates this behaviour so that a warning is raised instead


hi-ml-cpath/src/health_cpath/utils/viz_utils.py
```diff
@@ -108,10 +109,11 @@ def plot_attention_tiles(
     """
     tile_nodes = slide_node.top_tiles if top else slide_node.bottom_tiles
     num_rows = int(ceil(len(tile_nodes) / num_columns))
-    assert (
-        num_rows > 0
-    ), "The number of selected top and bottom tiles is too low. Try debugging with a higher num_top_tiles and/or a "
-    "higher number of batches."
+    if num_rows == 0:
+        logging.warning(
+            "The number of selected top and bottom tiles is too low. Try debugging with a higher "
```
kenza-bouzid, line 114: Can you also say that plotting will be skipped? just so that it's clear for the user

hi-ml-cpath/src/health_cpath/utils/plots_utils.py
```diff
@@ -87,11 +87,15 @@ def save_top_and_bottom_tiles(
     top_tiles_fig = plot_attention_tiles(
         slide_node=slide_node, case=case, top=True, num_columns=num_columns, figsize=figsize
     )
+    if top_tiles_fig is None:
```
kenza-bouzid, line 90: I would suggest moving this check to save_figure function? to avoid duplicating code? 
make fig optional and only save when it's not None?

kenza-bouzid (approved):
LGTM, approved with minor suggestions.

ant0nsc (approved):

PR #538, opened 2022-07-26
ENH: Make label_column an instance attribute

kenza-bouzid:
Change LABEL_COLUMN and N_CLASSES to be instance attributes instead of class attributes to offer more flexibility to change labels for a same data cohort and make split labels global in the repo.


hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
```diff
@@ -65,11 +62,15 @@ def __init__(self,
         :param validate_columns: Whether to call `validate_columns()` at the end of `__init__()`.
         `validate_columns()` checks that the loaded data frame for the dataset contains the expected column names
         for this class
+        :param label_column: CSV column name for tile label.
```
harshita-s, line 65: Could you add the default value for `label_column`

hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
```diff
@@ -177,11 +170,15 @@ def __init__(self,
...
```
harshita-s, line 173: Same as above, default value is missing

harshita-s (approved):

hi-ml-cpath/testhisto/testhisto/models/test_deepmil.py
```diff
@@ -231,7 +231,7 @@ def _mock_get_encoder(  # type: ignore
                 TilesDataset.SLIDE_ID_COLUMN: [str(slide_idx)] * bag_size,
                 TilesDataset.TILE_ID_COLUMN: [f"{slide_idx}-{tile_idx}" for tile_idx in range(bag_size)],
                 TilesDataset.IMAGE_COLUMN: rand(bag_size, *input_dim),
-                TilesDataset.LABEL_COLUMN: bag_label.expand(bag_size),
+                LABEL_COLUMN: bag_label.expand(bag_size),
```
dccastro, line 234: The need for this change highlights that we should eventually consider switching the pipeline to use standardised `TileKey`s.

dccastro (approved):

hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
```diff
@@ -15,44 +15,41 @@
...
-    N_CLASSES: int = 1  # binary classification by default
-
     def __init__(self,
                  root: Union[str, Path],
                  dataset_csv: Optional[Union[str, Path]] = None,
                  dataset_df: Optional[pd.DataFrame] = None,
                  train: Optional[bool] = None,
-                 validate_columns: bool = True) -> None:
+                 validate_columns: bool = True,
+                 label_column: str = "label",
```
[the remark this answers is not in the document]
kenza-bouzid: good point, will do!

hi-ml-cpath/testhisto/testhisto/models/test_deepmil.py
kenza-bouzid: Definitely, we should keep it in mind for future refinements of the pipeline.

hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
```diff
@@ -15,44 +15,41 @@
 from health_cpath.utils.naming import SlideKey
 
 
+TRAIN_SPLIT_LABEL = "train"  # Value used to indicate the training split in `SPLIT_COLUMN`
+VAL_SPLIT_LABEL = "val"  # Value used to indicate the validation split in `SPLIT_COLUMN`
+TEST_SPLIT_LABEL = "test"  # Value used to indicate the test split in `SPLIT_COLUMN`
```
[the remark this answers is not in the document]
kenza-bouzid: I moved it to avoid duplications in the code base.

Especially for testing where we were redefining these labels in MockTilesDataset which makes it hard to keep the mock dataset in sync with the base code.

So far, all datasets use "train"/"val"/"test" which is the standard afaik. These global variables should be used as default behaviour.

Maybe renaming them to `DEFAULT_TRAIN/TEST/VAL_SPLIT_LABEL` will make it clearer? 

I can also revert this change if you think that it reduces splitting flexibility and that it will affect your work on the test split.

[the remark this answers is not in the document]
dccastro: Makes sense. Renaming to `DEFAULT_*` sounds like a good compromise.

PR #543, opened 2022-07-28
ENH: Add hubconf to load models without installing

fepegar:
This allows to run

```python
import torch
github = 'microsoft/hi-ml:add-hubconf'
model = torch.hub.load(github, 'biovil_resnet', pretrained=True)
image = torch.rand(1, 3, 480, 480)
output = model(image)
output.projected_patch_embeddings.shape  # prints torch.Size([1, 128, 15, 15])
```

It will work **even if `hi-ml-multimodal` is not installed**. Only `torch` and `torchvision` need to be installed. For that to be possible, I've had to do some not-so-nice refactoring, like declaring the same variable (Hugging Face constants) in two modules.

If this PR is merged, the snippet above won't need `:add-hubconf` and will use `main` by default.

### DIFF
--- a/hi-ml-multimodal/src/health_multimodal/__init__.py
+++ b/hi-ml-multimodal/src/health_multimodal/__init__.py
@@ -2,3 +2,5 @@
 #  Copyright (c) Microsoft Corporation. All rights reserved.
 #  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 #  -------------------------------------------------------------------------------------------
+
+__version__ = "0.1.0"
--- a/hi-ml-multimodal/src/health_multimodal/__init__.py
+++ b/hi-ml-multimodal/src/health_multimodal/__init__.py
@@ -2,21 +2,3 @@
 #  Copyright (c) Microsoft Corporation. All rights reserved.
 #  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 #  -------------------------------------------------------------------------------------------
-
-BIOMED_VLP_CXR_BERT_SPECIALIZED = "microsoft/BiomedVLP-CXR-BERT-specialized"
-REPO_URL = f"https://huggingface.co/{BIOMED_VLP_CXR_BERT_SPECIALIZED}"
-CXR_BERT_COMMIT_TAG = "v1.1"
-
-BIOVIL_IMAGE_WEIGHTS_NAME = "biovil_image_resnet50_proj_size_128.pt"
-BIOVIL_IMAGE_WEIGHTS_URL = f"{REPO_URL}/resolve/{CXR_BERT_COMMIT_TAG}/{BIOVIL_IMAGE_WEIGHTS_NAME}"
-BIOVIL_IMAGE_WEIGHTS_MD5 = "02ce6ee460f72efd599295f440dbb453"
-
-
-__all__ = [
-    "BIOMED_VLP_CXR_BERT_SPECIALIZED",
-    "REPO_URL",
-    "CXR_BERT_COMMIT_TAG",
-    "BIOVIL_IMAGE_WEIGHTS_NAME",
-    "BIOVIL_IMAGE_WEIGHTS_URL",
-    "BIOVIL_IMAGE_WEIGHTS_MD5",
-]
--- a/hi-ml-multimodal/src/health_multimodal/image/__init__.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/__init__.py
@@ -37,6 +37,7 @@
 from .model import ImageModel
 from .model import ResnetType
 from .inference_engine import ImageInferenceEngine
+from .utils import get_biovil_resnet
 from .utils import get_biovil_resnet_inference
 
 
@@ -44,5 +45,6 @@ __all__ = [
     "ImageModel",
     "ResnetType",
     "ImageInferenceEngine",
+    "get_biovil_resnet",
     "get_biovil_resnet_inference",
 ]
--- a/hi-ml-multimodal/src/health_multimodal/image/__init__.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/__init__.py
@@ -36,8 +36,8 @@
 
 from .model import ImageModel
 from .model import ResnetType
+from .model import get_biovil_resnet
 from .inference_engine import ImageInferenceEngine
-from .utils import get_biovil_resnet
 from .utils import get_biovil_resnet_inference
 
 
--- a/hi-ml-multimodal/src/health_multimodal/image/model/model.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/model/model.py
@@ -3,7 +3,10 @@
 #  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 #  -------------------------------------------------------------------------------------------
 
+from __future__ import annotations
+
 import enum
+import tempfile
 from pathlib import Path
 from dataclasses import dataclass
 from typing import Any, Optional, Tuple, Union, Sequence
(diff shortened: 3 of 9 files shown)


fepegar:
You might be interested, @benbo.

hubconf.py
```diff
@@ -0,0 +1,21 @@
...
+repo_dir = Path(__file__).parent
+multimodal_src_dir = repo_dir / "hi-ml-multimodal" / "src" / "health_multimodal" / "image"
+sys.path.append(str(multimodal_src_dir))
+
+from model import ImageModel
+from model import get_biovil_resnet as _biovil_resnet
+# autopep8: on
+
+
+def biovil_resnet(pretrained: bool = False) -> ImageModel:
```
ozan-oktay, line 15: Could we please add a unit-test comparing the models initialised with (I) our old resnet code and (II) torch-hub code, running them on some dummy torch rand (3, 448, 448) array and comparing their outputs?

ozan-oktay (changes requested):

hubconf.py
ozan-oktay: Not only the shapes, but also the data content.

hi-ml-multimodal/src/health_multimodal/text/utils.py
```diff
@@ -6,14 +6,15 @@
...
-from .. import BIOMED_VLP_CXR_BERT_SPECIALIZED
-from .. import CXR_BERT_COMMIT_TAG
-
 from .inference_engine import TextInferenceEngine
 from .model import CXRBertModel
 from .model import CXRBertTokenizer
 
 
+BIOMED_VLP_CXR_BERT_SPECIALIZED = "microsoft/BiomedVLP-CXR-BERT-specialized"
+CXR_BERT_COMMIT_TAG = "v1.1"
```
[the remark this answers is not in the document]
fepegar: Good point, I'll try that.

hi-ml-multimodal/src/health_multimodal/image/model/resnet.py
```diff
@@ -8,7 +8,10 @@
 import torch
 
 from torchvision.models.resnet import model_urls, ResNet, BasicBlock, Bottleneck
-from torchvision.models.utils import load_state_dict_from_url
+try:
```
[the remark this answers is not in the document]
fepegar: The module has been removed in recent versions. I'll add a comment and verify that behaviour is unchanged.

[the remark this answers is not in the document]
fepegar: I've just checked. It's actually importing the same code, they just moved it (and broke the API!). I've also updated this import and verified that it works for old (ours) and new (latest) versions of PyTorch (432ea3c).

hubconf.py
fepegar: If we add this test now, it will need to reference the branch of this PR (https://github.com/microsoft/hi-ml/pull/543#issue-1320507174) and will fail once it's merged. I suggest the following instead:

1. Merge this PR
2. Open an issue about that test
3. Open a PR with the new test
4. Merge the new PR and close the issue

What do you think?

hi-ml-multimodal/src/health_multimodal/text/utils.py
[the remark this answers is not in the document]
fepegar: Fixed in 9f02edd.

hi-ml-multimodal/src/health_multimodal/__init__.py
```diff
@@ -2,21 +2,3 @@
 #  Copyright (c) Microsoft Corporation. All rights reserved.
 #  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 #  -------------------------------------------------------------------------------------------
-
-BIOMED_VLP_CXR_BERT_SPECIALIZED = "microsoft/BiomedVLP-CXR-BERT-specialized"
```
Shruthi42, line 6: We're moving functions/constants around a lot, and these are part of our public API. Are the changes here important enough to give up on backwards compatibility, why are we choosing to go this route?

We should also be more cautious in deciding what we make part of the public API in the future. https://peps.python.org/pep-0008/#public-and-internal-interfaces

hi-ml-multimodal/src/health_multimodal/image/__init__.py
```diff
@@ -37,12 +37,14 @@
 from .model import ImageModel
 from .model import ResnetType
 from .inference_engine import ImageInferenceEngine
+from .utils import get_biovil_resnet
```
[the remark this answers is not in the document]
fepegar: Good catch. I wonder why tests are not failing.

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -3,19 +3,60 @@
...
+from torchvision.datasets.utils import download_url
 
 from .resnet import resnet18, resnet50
 from .modules import MLP, MultiTaskModel
 
 TypeImageEncoder = Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]
+MODEL_TYPE = "resnet50"
+JOINT_FEATURE_SIZE = 128
+
+BIOMED_VLP_CXR_BERT_SPECIALIZED = "microsoft/BiomedVLP-CXR-BERT-specialized"
```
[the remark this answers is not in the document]
fepegar: I know! But this is the only workaround I found to overcome the issues I mentioned in the description of this PR, i.e., the PyTorch Hub method should work without having to install the package. Feel free to try placing this somewhere else, I know it's not very pretty at the moment.

hi-ml-multimodal/src/health_multimodal/__init__.py
fepegar: I think this variable doesn't need to be public. I agree we could be more cautious about this. I don't have much experience with making these differences, happy to hear suggestions.

As I explained in https://github.com/microsoft/hi-ml/pull/543#discussion_r934416369 and the PR description, some refactoring was needed so the PyTorch Hub method works.

fepegar: Thinking a bit more about this. I think to understand your comment, we'd need to define "our public API". Some of the variables that are being moved are "public" in the sense that they're used by other modules within the library. That doesn't mean we're expecting potential users to use them, I think. The package is not even released, and it doesn't even follow semantic versioning (see #545). Before version 1.0.0, the API should be considered unstable anyway. In summary, maybe we don't need to worry too much about these issues, at least for now.

hi-ml-multimodal/src/health_multimodal/image/__init__.py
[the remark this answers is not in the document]
Shruthi42: It's imported from `model` in `utils`, this is probably why.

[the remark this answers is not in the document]
fepegar: It doesn't fail because that function is imported in `utils`. But it's not needed here, so I'll remove this import.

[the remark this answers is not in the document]
fepegar: Fixed in 0a2075c.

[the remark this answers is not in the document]
fepegar: Oops, I missed your comment https://github.com/microsoft/hi-ml/pull/543#discussion_r934450819

hi-ml-multimodal/src/health_multimodal/__init__.py
Shruthi42: Sounds good to me for this PR. We should think about the public API soon, so we can keep it in mind as we add to the package.

hubconf.py
ozan-oktay: If this is the new and suggested way of loading biovil resnet, then we should it test and make sure that the forward pass outputs are matched. The worst-case scenario would be the following: The PR is merged, someone pulls the models from torch-hub or checks out to this commit and runs experiments with the wrong model. I think we probably don't want this corner case.

Something like the following, it may be syntactically incorrect but gives the basic idea.
```
import torch
github = 'microsoft/hi-ml:add-hubconf'
model1 = torch.hub.load(github, 'biovil_resnet', pretrained=True)
model2 = "<your_old_way_of_loading_it>
dummy_input = torch.rand(...)
assert torch.all(model1(dummy_input), model2(dummy_input), tol=...)
```

hi-ml-multimodal/src/health_multimodal/image/model/model.py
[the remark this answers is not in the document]
fepegar: I also agree but haven't been able to come up with a better solution so far.

hubconf.py
fepegar: Context for @Shruthi42:

I will add the suggested test pointing to this branch. If the CI passes, I'll remove the name of the branch, which will make the CI fail again, but will be ready to merge. I hope that makes sense.

fepegar:
Now that all tests have passed, including the recently added one, I'll remove the branch from the `github` string. Tests should fail, but we'll be safe to merge.

hi-ml-multimodal/test_multimodal/image/model/test_model.py
```diff
@@ -132,3 +136,23 @@ def test_reload_resnet_with_dilation() -> None:
...
+    image = torch.rand(1, 3, 480, 480)
+
+    github = 'microsoft/hi-ml'
+    model_hub = torch.hub.load(github, 'biovil_resnet', pretrained=True)
+    model_himl = get_biovil_resnet()
+
+    output_hub: ImageModelOutput = model_hub(image)
+    output_himl: ImageModelOutput = model_himl(image)
+
+    for field_hub, field_himl in zip(fields(output_hub), fields(output_himl)):
```
Shruthi42, line 153: Quick comment - couldn't you be missing some fields here when you zip them together? Why not just check the ones from the himl model?

ozan-oktay (approved):

hi-ml-multimodal/test_multimodal/image/model/test_model.py
fepegar: Addressed in a14767f.

hi-ml-multimodal/test_multimodal/image/model/test_model.py
```diff
@@ -132,3 +136,23 @@ def test_reload_resnet_with_dilation() -> None:
...
+    image = torch.rand(1, 3, 480, 480)
+
+    github = 'microsoft/hi-ml'
+    model_hub = torch.hub.load(github, 'biovil_resnet', pretrained=True)
+    model_himl = get_biovil_resnet()
+
+    output_hub: ImageModelOutput = model_hub(image)
+    output_himl: ImageModelOutput = model_himl(image)
+
+    for field_himl in zip(fields(output_himl)):
```
Shruthi42, line 153:
```suggestion
    for field_himl in fields(output_himl):
```

Otherwise this will return each field enclosed in a tuple.

Shruthi42 (approved):

hi-ml-multimodal/test_multimodal/image/model/test_model.py
fepegar: Oops, good catch!

PR #552, opened 2022-08-04
FIX: Hotfix HTTP error downloading model with hub

fepegar:
Specifying the branch when downloading a model using PyTorch Hub shouldn't be necessary, but an HTTP error was being raised anyway. There seem to have been issues with this in the past:

- https://github.com/pytorch/pytorch/pull/62139
- https://github.com/pytorch/pytorch/issues/61755
- https://github.com/n2cholas/jax-resnet/issues/4
- https://github.com/pytorch/pytorch/pull/61761

Explicitly specifying `main` seems to help, so I'm adding it to the test, and it should be included in any future documentation, just in case.

### DIFF
--- a/hi-ml-multimodal/test_multimodal/image/model/test_model.py
+++ b/hi-ml-multimodal/test_multimodal/image/model/test_model.py
@@ -143,7 +143,7 @@ def test_hubconf() -> None:
     """Test that instantiating the image model using the PyTorch Hub is consistent with older methods."""
     image = torch.rand(1, 3, 480, 480)
 
-    github = 'microsoft/hi-ml'
+    github = 'microsoft/hi-ml:main'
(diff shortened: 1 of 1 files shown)


Shruthi42 (approved):

Shruthi42:
Approved. Do we know how torch hub recommends versioning models? Do they recommend always pointing to main or to a specific release, for example?

fepegar:
Thanks. No recommendations, really. Just possibilities. This is all they have:

> **github** - a string with format “repo_owner/repo_name[:ref]” with an optional ref (tag or branch). If ref is not specified, the default branch is assumed to be main if it exists, and otherwise master. Example: ‘pytorch/vision:0.10’

In future docs, we can ask mention that tags should be specified for specific versions. But this is not trivial in a multi-package repo...

PR #554, opened 2022-08-04
ENH: Enable flexible weights initialization from src_checkpoint

kenza-bouzid:
Add flags `pretrain_encoder pretrainer_pooling pretrain_classifier` to init submodules weights from a pretrained checkpoint in `src_checkpoint`

### DIFF
--- a/hi-ml-cpath/src/health_cpath/models/deepmil.py
+++ b/hi-ml-cpath/src/health_cpath/models/deepmil.py
@@ -17,7 +17,7 @@ from health_cpath.models.encoders import IdentityEncoder
 from health_cpath.utils.deepmil_utils import EncoderParams, PoolingParams, set_module_gradients_enabled
 
 from health_cpath.datasets.base_dataset import TilesDataset
-from health_cpath.utils.naming import MetricsKey, ResultsKey, SlideKey, ModelKey, TileKey
+from health_cpath.utils.naming import DeepMILSubModules, MetricsKey, ResultsKey, SlideKey, ModelKey, TileKey
 from health_cpath.utils.output_utils import (BatchResultsType, DeepMILOutputsHandler, EpochResultsType,
                                              validate_class_names)
 
@@ -105,7 +105,9 @@ class BaseDeepMILModule(LightningModule):
         self.test_metrics = self.get_metrics()
 
     @staticmethod
-    def copy_weights(current_submodule: nn.Module, pretrained_submodule: nn.Module, submodule_name: str) -> None:
+    def copy_weights(
+        current_submodule: nn.Module, pretrained_submodule: nn.Module, submodule_name: DeepMILSubModules
+    ) -> None:
         """Copy weights from pretrained submodule to current submodule.
 
         :param current_submodule: Submodule to copy weights to.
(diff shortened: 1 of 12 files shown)


hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -301,17 +302,3 @@ def test_model_weights_when_resume_training() -> None:
             mock_create_trainer.assert_called_once()
             recovery_checkpoint = mock_create_trainer.call_args[1]["resume_from_checkpoint"]
             assert recovery_checkpoint == runner.checkpoint_handler.trained_weights_path
-
-
-def test_runner_end_to_end() -> None:
```
kenza-bouzid, line 306: This test doesn't do anything, I merged by accident in #518

hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -94,14 +98,44 @@ def __init__(self,
...
+        :param current_submodule: Submodule to copy weights to.
+        :param pretrained_submodule: Submodule to copy weights from.
+        :param submodule_name: Name of the submodule.
+        """
+
+        for param, pretrained_param in zip(current_submodule.parameters(), pretrained_submodule.parameters()):
+            try:
+                param.data.copy_(pretrained_param.data)
+            except Exception as e:
+                logging.warning(f"Failed to copy weights for {submodule_name} because of the following exception: {e}"
```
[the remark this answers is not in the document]
maxilse: IMO, this should be blocking.

maxilse (changes requested):

hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -94,14 +97,43 @@ def __init__(self,
...
+        :param current_submodule: Submodule to copy weights to.
+        :param pretrained_submodule: Submodule to copy weights from.
+        :param submodule_name: Name of the submodule.
+        """
+
+        for param, pretrained_param in zip(current_submodule.parameters(), pretrained_submodule.parameters()):
+            try:
+                param.data.copy_(pretrained_param.data)
+            except Exception as e:
+                raise ValueError(f"Failed to copy weights for {submodule_name} because of the following exception: {e}")
```
vale-salvatelli, line 120: I think we should test this method - with a couple of different submodules. I see a test below to test weight options but that is not testing the values, only the number of calls. Things I can think we want to assert: values are copied correctly, no mismatches in the value types. 
We could also check here that the number of parameters is the same and raise an error if not (I can see the possibility we try by mistake to initialize a model with weights coming from a checkpoint trained with a different version of the model)

vale-salvatelli (changes requested):

hi-ml-cpath/src/health_cpath/models/deepmil.py
kenza-bouzid: Agree, will add more tests.

maxilse (approved):

hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -94,14 +97,43 @@ def __init__(self,
...
+            pretrained_model = self.load_from_checkpoint(checkpoint_path=str(pretrained_checkpoint_path))
+
+            if self.encoder_params.pretrained_encoder:
+                self.copy_weights(self.encoder, pretrained_model.encoder, "encoder")
+
+            if self.pooling_params.pretrained_pooling:
+                self.copy_weights(self.aggregation_fn, pretrained_model.aggregation_fn, "pooling")
+
+            if self.pretrained_classifier and pretrained_model.n_classes == self.n_classes:
+                self.copy_weights(self.classifier_fn, pretrained_model.classifier_fn, "classifier")
```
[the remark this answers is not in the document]
kenza-bouzid: Done!

hi-ml-cpath/testhisto/testhisto/models/test_deepmil.py
```diff
@@ -566,3 +573,119 @@ def test_init_weights_options(pretrained_encoder: bool, pretrained_pooling: bool
...
+def _get_base_deepmil_module(
+    pretrained_encoder: bool = True,
+    pretrained_pooling: bool = True,
+    pretrained_classifier: bool = True,
+    n_classes: int = 3,
+    num_layers: int = 2,
+    num_heads: int = 1,
+    hidden_dim: int = 8,
+) -> BaseDeepMILModule:
+    module = BaseDeepMILModule(
```
vale-salvatelli, line 587: Wonder if it would be better to test this with TilesDeepMILModule - just aware it contains more params of the base class and it's what we're using now for experiments

kenza-bouzid: It doesn't matter much because all we're testing is defined in the Base Class but sure, we can change it to TilesDeepMILModule

vale-salvatelli (approved):

vale-salvatelli:
Excellent work Kenza! Approved with a minor suggestion

PR #555, opened 2022-08-04
BUG: Fix output saving for distributed models

mebristo:
Saving outputs was previously failing (under different error messages depending) when runs were distributed across multiple devices on a machine. This PR fixes that, and also updates the smoke tests to run on multiple devices to catch any such error in future


hi-ml-cpath/src/health_cpath/utils/output_utils.py
```diff
@@ -205,7 +206,12 @@ def should_save_validation_outputs(self, metrics_dict: Mapping[MetricsKey, Metri
         :return: Whether this is the best validation epoch so far.
         """
         # The metric needs to be computed on all ranks to allow synchronisation
-        metric_value = float(metrics_dict[self.primary_val_metric].compute())
+        metric = metrics_dict[self.primary_val_metric]
+        if not metric._update_called:
+            return False
+        metric_value = float(metric.compute())
+        if isinstance(metric, Accuracy):
```
dccastro, line 213: Would you mind adding a comment explaining why we need to special-case for `Accuracy`?

hi-ml-cpath/src/health_cpath/utils/output_utils.py
```diff
@@ -205,7 +206,12 @@ def should_save_validation_outputs(self, metrics_dict: Mapping[MetricsKey, Metri
         :return: Whether this is the best validation epoch so far.
         """
         # The metric needs to be computed on all ranks to allow synchronisation
-        metric_value = float(metrics_dict[self.primary_val_metric].compute())
+        metric = metrics_dict[self.primary_val_metric]
+        if not metric._update_called:
```
dccastro, line 210: What's the reasoning here?

hi-ml-cpath/Makefile
```diff
@@ -99,34 +99,41 @@ define CRCKSIMCLR_ARGS
 --model=SSL.CRCK_SimCLR
 endef
 
+define REGRESSION_TEST_ARGS
+--cluster dedicated-nc24s-v2 --regression_test_csv_tolerance=0.5
+endef
+
+define PANDA_REGRESSION_METRICS
+--regression_metrics='test/accuracy,test/macro_accuracy,test/weighted_accuracy,test/auroc,test/ISUP 0,test/ISUP 1,\
+test/ISUP 2,test/ISUP 3,test/ISUP4, test/ISUP5,test/loss_epoch'
```
dccastro, line 108: Typos: `ISUP 4`, `ISUP 5`?

dccastro (approved):

hi-ml-cpath/src/health_cpath/utils/output_utils.py
mebristo: Without it we see this warning: `UserWarning: The ``compute`` method of metric Accuracy was called before the ``update`` method which may lead to errors, as metric states have not yet been updated.` 
If the metric hasn't been updated then it is erroneous so we don't want to save it

hi-ml-cpath/Makefile
mebristo: Good catch!

hi-ml-cpath/src/health_cpath/utils/output_utils.py
mebristo: I'm a little confused here since on `validation_epoch_end` we call `log_metrics(ModelKey.VAL)` which iteratively calls `metric.log()` on every metric in the validation dictionary. This implicitly calls `.compute()` and then `.reset()` on each metric. So by the point this part of the code is reached - all validation metrics should have been reset and therefore there is nothing to compute...

mebristo: I've added a warning to catch any case like this

vale-salvatelli (approved):

vale-salvatelli:
This PR looks good to me but it doesn't seem to address the issue I am seeing - that one seems due to an error in removing the previous validation folder, if needed - am I missing something? I can open a different PR if the issue is not covered here

mebristo:
> This PR looks good to me but it doesn't seem to address the issue I am seeing - that one seems due to an error in removing the previous validation folder, if needed - am I missing something? I can open a different PR if the issue is not covered here

This line will fix the error you are seeing: 
`shutil.rmtree(source, ignore_errors=True)`

PR #559, opened 2022-08-08
ENH: Add code to create PANDAS tiles montage

fepegar:
This is a cleaner version of

- https://github.com/microsoft/hi-ml/pull/327

Here are @dccastro's comments on it, with which I agree:

> Thanks for implementing this functionality, @fepegar! Could we perhaps move the thumbnail and montage stuff to a separate module as standalone functions, so we can easily reuse across datasets? Also not convinced it makes sense for them to be part of the dataset class itself.

No docs, apologies. This was very experimental. But hopefully the code is quite self-descriptive. I suspect Copilot would do a good job if prompted to complete docstrings.

@harshita-s, feel free to ask if something is not clear, and I'll try to remember why I did what I did. I've suggested some reviewers, but you might want to change them.

Here's a video by Matt Parker I randomly discovered recently and wished I had watched before implementing this: [Why don't Jigsaw Puzzles have the correct number of pieces?](https://www.youtube.com/watch?v=vXWvptwoCl8)

### DIFF
--- a/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
+++ b/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
@@ -9,10 +9,13 @@ from typing import Any, Dict, Optional, Tuple, Union
 import numpy as np
 import pandas as pd
 import torch
-from sklearn.utils.class_weight import compute_class_weight
 from torch.utils.data import Dataset
+import torchvision
+from sklearn.utils.class_weight import compute_class_weight
+from monai.data.image_reader import WSIReader
 
 from health_cpath.utils.naming import SlideKey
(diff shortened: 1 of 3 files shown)


harshita-s:
> Main comments:
> 
> * I would not add this functionality to the panda_dataset class. I understand the dependency from the masks makesit difficult to generalize across datasets but this to me adds too much to PandaDataset. You could subclass to keep this functionality separate.
> * it needs some tests to be merged in main
> 
> Not sure if the aim at this point is to merge this branch or @harshita-s will create a different version. Happy to chat more and having a deeper look if the first.

I am using a version in cpath. Will open a PR soon.

PR #562, opened 2022-08-08
ENH: Enable custom SSL encoder checkpoint

mebristo:
Add command line arg to enable custom SSL encoder checkpoint

This also allows us to add an additional smoke test for TcgaCrckSslMil which requires a checkpoint for the encoder

### DIFF
--- a/hi-ml-cpath/Makefile
+++ b/hi-ml-cpath/Makefile
@@ -77,6 +77,8 @@ pytest:
 pytest_coverage:
 	pytest --cov=health_cpath --cov SSL --cov-branch --cov-report=html --cov-report=xml --cov-report=term-missing --cov-config=.coveragerc
 
+SSL_CKPT_RUN_ID_CRCK := CRCK_SimCLR_1655731022_85790606
+
 # Run regression tests and compare performance
 define BASE_CPATH_RUNNER_COMMAND
 cd ../ ; \
@@ -92,7 +94,7 @@ define DEEPSMILEPANDATILES_ARGS
 endef
 
 define TCGACRCKSSLMIL_ARGS
---model=health_cpath.TcgaCrckSSLMIL --ssl_checkpoint_run_id=CRCK_SimCLR_1655731022_85790606
+--model=health_cpath.TcgaCrckSSLMIL --ssl_checkpoint_run_id=${SSL_CKPT_RUN_ID_CRCK}
 endef
 
 define CRCKSIMCLR_ARGS
--- a/hi-ml-cpath/Makefile
+++ b/hi-ml-cpath/Makefile
@@ -92,7 +92,7 @@ define DEEPSMILEPANDATILES_ARGS
 endef
 
 define TCGACRCKSSLMIL_ARGS
---model=health_cpath.TcgaCrckSSLMIL
+--model=health_cpath.TcgaCrckSSLMIL --ssl_checkpoint=CRCK_SimCLR_1655731022_85790606
 endef
 
 define CRCKSIMCLR_ARGS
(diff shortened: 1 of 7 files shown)


hi-ml-cpath/testhisto/testhisto/mocks/container.py
```diff
@@ -38,6 +38,7 @@ def __init__(self, tmp_path: Path, **kwargs: Any) -> None:
             # declared in TrainerParams:
             max_epochs=2,
             crossval_count=1,
+            ssl_checkpoint_run_id="",
```
kenza-bouzid, line 41: Why do we need to reset it to the default value here?

kenza-bouzid (approved):
LGTM, left a quick question for the mock container

hi-ml-cpath/testhisto/testhisto/mocks/container.py
mebristo: Good catch - removed

hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
```diff
@@ -83,7 +85,7 @@ def cache_dir(self) -> Path:
         return Path(f"/tmp/himl_cache/{self.__class__.__name__}-{self.encoder_type}/")
 
     def setup(self) -> None:
-        self.ssl_ckpt_run_id = ""
+        self.ssl_ckpt_run_id = self.ssl_checkpoint_run_id
```
dccastro, line 88: I feel this duplication could be confusing. How about consolidating it into a single parameter/attribute?

hi-ml-cpath/Makefile
```diff
@@ -92,7 +92,7 @@ define DEEPSMILEPANDATILES_ARGS
 endef
 
 define TCGACRCKSSLMIL_ARGS
---model=health_cpath.TcgaCrckSSLMIL
+--model=health_cpath.TcgaCrckSSLMIL --ssl_checkpoint=CRCK_SimCLR_1655731022_85790606
```
dccastro, line 95: Shouldn't this argument match the parameter name (`ssl_checkpoint_run_id`)?

dccastro (changes requested):

hi-ml-cpath/Makefile
vale-salvatelli: if we have some golden checkpoints what about we save them in a naming class and give them more telling names? and pass here the attribute of the class?

hi-ml-cpath/src/health_cpath/configs/classification/DeepSMILEPanda.py
```diff
@@ -70,7 +70,8 @@ def __init__(self, **kwargs: Any) -> None:
 
     def setup(self) -> None:
         BaseMILTiles.setup(self)
-        self.ssl_ckpt_run_id = innereye_ssl_checkpoint_binary
+        # If no SSL checkpoint is provided, use the default one
+        self.ssl_ckpt_run_id = self.ssl_ckpt_run_id or innereye_ssl_checkpoint_binary
```
vale-salvatelli, line 74: along the line of my comment above `innereye_ssl_checkpoint_binary` could be `innereye_panda_ssl_checkpoint_binary`

mebristo: I was trying to avoid having anything hardcoded since these run ids are workspace dependent, and we are using 2 so it would be a pain to change this variable in order to change workspace. IMO it is easier to specify it from the command line.
 To make it easier for us to remember which run id to specify on the command line, we could create variables in the Makefile instead?

dccastro (approved):

hi-ml-cpath/src/health_cpath/configs/classification/DeepSMILEPanda.py
vale-salvatelli: variables in the makefile sounds good to me

PR #565, opened 2022-08-09
ENH: Clean up logs

mebristo:
A couple of fixes to remove some of the noise from logs from running the runner

### DIFF
--- a/hi-ml/src/health_ml/run_ml.py
+++ b/hi-ml/src/health_ml/run_ml.py
@@ -253,8 +253,9 @@ class MLRunner:
             assert self.trainer, "Trainer should be initialized before training. Call self.init_training() first."
             self.trainer.fit(self.container.model, datamodule=self.data_module)
 
-        assert self.trainer.logger is not None
-        self.trainer.logger.finalize('success')
+        for logger in self.trainer.loggers:
+            assert logger is not None
+            logger.finalize('success')
 
     def run_validation(self) -> None:
         """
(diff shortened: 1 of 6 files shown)


hi-ml/src/health_ml/run_ml.py
```diff
@@ -253,8 +253,9 @@ def run_training(self) -> None:
             assert self.trainer, "Trainer should be initialized before training. Call self.init_training() first."
             self.trainer.fit(self.container.model, datamodule=self.data_module)
 
-        assert self.trainer.logger is not None
-        self.trainer.logger.finalize('success')
```
kenza-bouzid, line 257: is logger deprecated now?

mebristo: It will be in 1.8. This change is to prevent it from logging a warning here (about the deprecation)

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -146,12 +146,12 @@ def test_run_training() -> None:
...
             mock_trainer.fit = Mock()
             mock_close_logger = Mock()
-            mock_trainer.logger = MagicMock(close=mock_close_logger)
+            mock_trainer.loggers = [MagicMock(close=mock_close_logger)]
 
             runner.run_training()
 
             mock_trainer.fit.assert_called_once()
-            mock_trainer.logger.finalize.assert_called_once()
+            mock_trainer.loggers[0].finalize.assert_called_once()
```
kenza-bouzid, line 154: logger property return first logger why can't we use it anymore?
image

mebristo: We could, I just think this is more explicit and therefore clearer

harshita-s (approved):

kenza-bouzid (approved):

PR #568, opened 2022-08-11
FIX: Fix best checkpoint name 

mebristo:
Best metric checkpoint should reflect the primary metric of choice and whether the goal was to maximise or minimise

### DIFF
--- a/hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
+++ b/hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
@@ -75,7 +75,8 @@ class BaseMIL(LightningContainer, EncoderParams, PoolingParams):
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         self.run_extra_val_epoch = True  # Enable running an additional validation step to save tiles/slides thumbnails
-        self.best_checkpoint_filename = "checkpoint_max_val_auroc"
+        metric_optim = "max" if self.maximise_primary_metric else "min"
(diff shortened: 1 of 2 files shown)


hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
```diff
@@ -77,7 +77,8 @@ class BaseMIL(LightningContainer, EncoderParams, PoolingParams):
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         self.run_extra_val_epoch = True  # Enable running an additional validation step to save tiles/slides thumbnails
-        self.best_checkpoint_filename = "checkpoint_max_val_auroc"
+        metric_optim = "max" if self.maximise_primary_metric else "min"
+        self.best_checkpoint_filename = f"checkpoint_{metric_optim}_val_{self.primary_val_metric.value}"
         self.best_checkpoint_filename_with_suffix = self.best_checkpoint_filename + ".ckpt"
```
harshita-s, line 82: shouldn't the equivalent of "checkpoint_max_val_auroc" be `checkpoint_{metric_optim}_val_{self.primary_val_metric.name}`? I guess `self.primary_val_metric.value` will return the value of the metric?

mebristo: `self.primary_val_metric.name` would be  e.g. "ACC" / "ACC_MACRO" whereas `self.primary_val_metric.value` would be "accuracy"/"macro_accuracy" which I think is more readable

harshita-s (approved):

kenza-bouzid (approved):
LGTM!

PR #569, opened 2022-08-11
BUG: Set dataframe index column only when necessary

kenza-bouzid:
Index datasets dataframe by tile_id or slide_id only when necessary (e.g. the current index is not already to tile_id or slide_id  columns)


dccastro (approved):
Looks good! Only suggestion is to make the PR title more explicit, e.g. "Set dataframe index column only when necessary"

kenza-bouzid:
Index datasets dataframe by tile_id or slide_id only when necessary (e.g. the current index is not already set to tile_id or slide_id columns)

PR #572, opened 2022-08-12
ENH: Parameterize tiles intensity scaling

kenza-bouzid:
Add parameter to control tiles intensity scaling to be able to run augmentations on uint8 data to save some memory and speed up computations.

### DIFF
--- a/hi-ml-cpath/src/health_cpath/models/transforms.py
+++ b/hi-ml-cpath/src/health_cpath/models/transforms.py
@@ -30,8 +30,10 @@ def load_image_as_tensor(image_path: PathOrString, scale_intensity: bool = True)
     """Load an image as a tensor from the given path
 
     :param image_path: path to the image
-    :param scale_intensity: if True, scale the intensity to [0, 1]
+    :param scale_intensity: if True, scale the intensity to [0, 1]. Otherwise, uses `to_tensor` from torchvision
+        which scales the image pixel intensities tp [0, 1] by default.
     """
+
     pil_image = load_pil_image(image_path)
     if scale_intensity:
         return to_tensor(pil_image)
@@ -46,8 +48,10 @@ def load_image_stack_as_tensor(image_paths: Sequence[PathOrString],
 
     :param image_paths: paths to the images
     :param progress: if True, show a progress bar
-    :param scale_intensity: if True, scale the intensity to [0, 1]
+    :param scale_intensity: if True, scale the intensity to [0, 1].  Otherwise, uses `to_tensor` from torchvision
+        which scales the image pixel intensities tp [0, 1] by default.
     """
+
     loading_generator = (load_image_as_tensor(path) for path in image_paths)
     if progress:
         from tqdm import tqdm
@@ -117,8 +121,10 @@ class LoadTilesBatchd(MapTransform):
         :param allow_missing_keys: If `False` (default), raises an exception when an input
         dictionary is missing any of the specified keys.
         :param progress: Whether to display a tqdm progress bar.
-        :param scale_intensity: If True, scale the intensity to [0, 1]
+        :param scale_intensity: If True, scale the intensity to [0, 1]. Otherwise, uses `to_tensor` from torchvision
+        which scales the image pixel intensities tp [0, 1] by default.
         """
+
         super().__init__(keys, allow_missing_keys)
         self.progress = progress
         self.scale_intensity = scale_intensity
--- a/hi-ml-cpath/src/health_cpath/models/transforms.py
+++ b/hi-ml-cpath/src/health_cpath/models/transforms.py
@@ -115,7 +115,7 @@ class LoadTilesBatchd(MapTransform):
 
     # Cannot reuse MONAI readers because they support stacking only images with no channels
     def __init__(self, keys: KeysCollection, allow_missing_keys: bool = False,
-                 progress: bool = False, scale_intensity: bool = False) -> None:
+                 progress: bool = False, scale_intensity: bool = True) -> None:
         """
         :param keys: Key(s) for the image path(s) in the input dictionary.
         :param allow_missing_keys: If `False` (default), raises an exception when an input
--- a/hi-ml-cpath/src/health_cpath/models/transforms.py
+++ b/hi-ml-cpath/src/health_cpath/models/transforms.py
@@ -26,15 +26,28 @@ def load_pil_image(image_path: PathOrString) -> PIL.Image.Image:
     return image
 
 
-def load_image_as_tensor(image_path: PathOrString) -> torch.Tensor:
-    """Load an image as a tensor from the given path"""
+def load_image_as_tensor(image_path: PathOrString, scale_intensity: bool = True) -> torch.Tensor:
+    """Load an image as a tensor from the given path
+
+    :param image_path: path to the image
+    :param scale_intensity: if True, scale the intensity to [0, 1]
+    """
     pil_image = load_pil_image(image_path)
-    return to_tensor(pil_image)
+    if scale_intensity:
+        return to_tensor(pil_image)
+    else:
+        return torch.from_numpy(pil_image.transpose((2, 0, 1))).contiguous()
 
 
 def load_image_stack_as_tensor(image_paths: Sequence[PathOrString],
-                               progress: bool = False) -> torch.Tensor:
-    """Load a batch of images of the same size as a tensor from the given paths"""
+                               progress: bool = False,
+                               scale_intensity: bool = True) -> torch.Tensor:
+    """Load a batch of images of the same size as a tensor from the given paths
+
+    :param image_paths: paths to the images
+    :param progress: if True, show a progress bar
+    :param scale_intensity: if True, scale the intensity to [0, 1]
+    """
     loading_generator = (load_image_as_tensor(path) for path in image_paths)
     if progress:
         from tqdm import tqdm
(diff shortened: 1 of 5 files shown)


hi-ml-cpath/testhisto/testhisto/mocks/tiles_generator.py
```diff
@@ -114,7 +114,7 @@ def generate_mock_histo_data(self) -> None:
                 raise NotImplementedError
 
             tile_filename = self.dest_data_path / row[PandaTilesDataset.IMAGE_COLUMN]
-            save_image(tile.float(), tile_filename)
```
kenza-bouzid, line 117: save_image does some funky rescaling (assuming that the image is in [0, 1] scale )which was resulting in just white tiles
image

hi-ml-cpath/testhisto/testhisto/models/test_deepmil.py
```diff
@@ -133,27 +133,6 @@ def _test_lightningmodule(
             assert torch.all(score <= 1)
 
 
-@pytest.fixture(scope="session")
-def mock_panda_tiles_root_dir(
```
kenza-bouzid, line 137: moved it to conftest so that I can use it for other tests than test_deepmil

hi-ml-cpath/src/health_cpath/models/transforms.py
```diff
@@ -26,16 +26,33 @@ def load_pil_image(image_path: PathOrString) -> PIL.Image.Image:
...
 
 
-def load_image_as_tensor(image_path: PathOrString) -> torch.Tensor:
-    """Load an image as a tensor from the given path"""
+def load_image_as_tensor(image_path: PathOrString, scale_intensity: bool = True) -> torch.Tensor:
+    """Load an image as a tensor from the given path
+
+    :param image_path: path to the image
+    :param scale_intensity: if True, scale the intensity to [0, 1]. Otherwise, uses `to_tensor` from torchvision
+        which scales the image pixel intensities tp [0, 1] by default.
```
mebristo, line 34: Is this docstring the right way round? To me it looks like if scale_intensity is True, uses to_tensor, otherwise we return a transposed copy of the tensor.

Also there's nothing in the docs explaining why we transpose it/ permute the indices here?

kenza-bouzid: Good catch, my bad. I got confused in my own docs haha.

We need to permute because pil_image is in channel last format and we need channel first.
I will add a clarification comment

hi-ml-cpath/testhisto/testhisto/models/test_transforms.py
```diff
@@ -89,6 +90,28 @@ def test_load_tiles_batch() -> None:
...
+    index = 0
+
+    # Test that the transform returns images in [0, 255] range
+    bagged_batch = bagged_dataset[index]
+    manually_loaded_batch = load_batch_transform(bagged_batch)
+
+    pixels_dtype = torch.uint8 if not scale_intensity else torch.float32
+    max_val = 255 if not scale_intensity else 1.
+
+    assert manually_loaded_batch[image_key][0].dtype == pixels_dtype
```
mebristo, line 110: Would it be fairer to use a random integer as the index here instead of 0?

kenza-bouzid: Good point, will do!

hi-ml-cpath/src/health_cpath/models/transforms.py
```diff
@@ -26,16 +26,33 @@ def load_pil_image(image_path: PathOrString) -> PIL.Image.Image:
...
+    """Load an image as a tensor from the given path
+
+    :param image_path: path to the image
+    :param scale_intensity: if True, use `to_tensor` from torchvision which scales the image pixel intensities to
+    [0, 1] by default. Otherwise, transpose the image to [C, H, W] format and return it as a torch tensor.
+    """
+
+    pil_image = load_pil_image(image_path)  # pil_image is in channels last format [H, W, C]
+    if scale_intensity:
+        return to_tensor(pil_image)  # to_tensor scales the image pixel intensities tp [0, 1] by default
```
ant0nsc, line 39:
```suggestion
        return to_tensor(pil_image)  # to_tensor scales the image pixel intensities to [0, 1] by default
```

hi-ml-cpath/src/health_cpath/models/transforms.py
```diff
@@ -26,16 +26,33 @@ def load_pil_image(image_path: PathOrString) -> PIL.Image.Image:
...
 
-def load_image_as_tensor(image_path: PathOrString) -> torch.Tensor:
-    """Load an image as a tensor from the given path"""
-    pil_image = load_pil_image(image_path)
-    return to_tensor(pil_image)
+def load_image_as_tensor(image_path: PathOrString, scale_intensity: bool = True) -> torch.Tensor:
+    """Load an image as a tensor from the given path
+
+    :param image_path: path to the image
+    :param scale_intensity: if True, use `to_tensor` from torchvision which scales the image pixel intensities to
```
ant0nsc, line 33: In the function description here and in line 44ff, it would be good to emphasize that the returned tensor is _always_ in CHW format. At present, it looks as if this might be the case only for `scale=False`.

hi-ml-cpath/src/health_cpath/models/transforms.py
```diff
@@ -98,20 +115,26 @@ class LoadTilesBatchd(MapTransform):
...
     # Cannot reuse MONAI readers because they support stacking only images with no channels
     def __init__(self, keys: KeysCollection, allow_missing_keys: bool = False,
-                 progress: bool = False) -> None:
+                 progress: bool = False, scale_intensity: bool = True) -> None:
         """
         :param keys: Key(s) for the image path(s) in the input dictionary.
         :param allow_missing_keys: If `False` (default), raises an exception when an input
         dictionary is missing any of the specified keys.
         :param progress: Whether to display a tqdm progress bar.
+        :param scale_intensity: if True, use `to_tensor` from torchvision which scales the image pixel intensities to
```
ant0nsc, line 124: see above for comment about CHW

hi-ml-cpath/testhisto/conftest.py
```diff
@@ -73,3 +75,24 @@ def tmp_path_to_pathmnist_dataset(tmp_path_factory: pytest.TempPathFactory) -> G
     download_azure_dataset(tmp_dir, dataset_id=MockHistoDataType.PATHMNIST.value)
     yield tmp_dir
     shutil.rmtree(tmp_dir)
+
+
+@pytest.fixture(scope="session")
```
ant0nsc, line 80: How much time does that cost? Are session scoped fixtures created under all circumstances, even if they are not used (for example, because we only run a single test that does not use the fixture)?

hi-ml-cpath/testhisto/testhisto/models/test_transforms.py
```diff
@@ -89,6 +90,31 @@ def test_load_tiles_batch() -> None:
...
+
+    pixels_dtype = torch.uint8 if not scale_intensity else torch.float32
+    max_val = 255 if not scale_intensity else 1.
+
+    bag_size = manually_loaded_batch[image_key][index].shape[0] - 1
+    index = min(np.random.randint(0, max_bag_size), bag_size)
+
+    assert manually_loaded_batch[image_key][index].dtype == pixels_dtype
+    assert manually_loaded_batch[image_key][index].min() >= 0
+    assert manually_loaded_batch[image_key][index].max() <= max_val
```
ant0nsc, line 115: For the case where `max_val==255`, you should also check that the maximum is >1.

hi-ml-cpath/testhisto/testhisto/models/test_transforms.py
```diff
@@ -89,6 +90,31 @@ def test_load_tiles_batch() -> None:
...
+
+    # Test that the transform returns images in [0, 255] range
+    bagged_batch = bagged_dataset[index]
+    manually_loaded_batch = load_batch_transform(bagged_batch)
+
+    pixels_dtype = torch.uint8 if not scale_intensity else torch.float32
+    max_val = 255 if not scale_intensity else 1.
+
+    bag_size = manually_loaded_batch[image_key][index].shape[0] - 1
+    index = min(np.random.randint(0, max_bag_size), bag_size)
```
ant0nsc, line 111: I would just work with a fixed index (either check all entries in the bag, or just a single one - but if single, don't randomize. Randomization adds obvious problems with reproducibility of test failures.)

hi-ml-cpath/testhisto/testhisto/models/test_transforms.py
```diff
@@ -89,6 +90,31 @@ def test_load_tiles_batch() -> None:
...
+
```
ant0nsc, line 112: maybe add a check that the tile is not degenerate and has >1 distinct values?

ant0nsc (changes requested):
Looking good. Some minor suggestions, and a question around test setup speed.

hi-ml-cpath/src/health_cpath/models/transforms.py
kenza-bouzid: Good point, will do!

hi-ml-cpath/testhisto/conftest.py
kenza-bouzid: Not much, 
pytest takes 3m 43s and gpu tests 5m 54s 
vs 3m 39s / 5m 20s (before this change) 
Session scoped fixtures are not executed unless supplied as an argument to a test_* function
https://stackoverflow.com/questions/26899001/pytest-how-do-i-use-global-session-wide-fixtures

I think you might be referring to `autouse=True` ? 
https://docs.pytest.org/en/6.2.x/fixture.html#autouse-fixtures-fixtures-you-don-t-have-to-request

hi-ml-cpath/testhisto/testhisto/models/test_transforms.py
kenza-bouzid: Good point!

hi-ml-cpath/testhisto/testhisto/models/test_transforms.py
kenza-bouzid: Done!

hi-ml-cpath/testhisto/conftest.py
ant0nsc: I was more thinking of a case where a fixture is defined, but not actually used. Would it still execute something? I just verified that this is not the case.

ant0nsc (approved):

PR #574, opened 2022-08-16
ENH: Add specific ImageNet encoders for transferability study

kenza-bouzid:
Add specific ImageNet encoders for transferability study: Resnet18/50 with and without preproc

### DIFF
--- a/hi-ml-cpath/src/health_cpath/models/encoders.py
+++ b/hi-ml-cpath/src/health_cpath/models/encoders.py
@@ -68,6 +68,7 @@ class ImageNetEncoder(TileEncoder):
         returns a classifier pretrained on ImageNet, such as the ones from `torchvision.models.*`.
         :param tile_size: Tile width/height, in pixels.
         :param n_channels: Number of channels in the tile (default=3).
+        :param imagenet_preprocessing: Whether to apply ImageNet preprocessing to the input.
         """
         self.create_feature_extractor_fn = feature_extraction_model
         self.imagenet_preprocessing = imagenet_preprocessing
--- a/hi-ml-cpath/src/health_cpath/models/encoders.py
+++ b/hi-ml-cpath/src/health_cpath/models/encoders.py
@@ -10,7 +10,7 @@ import numpy as np
 import torch
 from pl_bolts.models.self_supervised import SimCLR
 from torch import Tensor as T, nn
-from torchvision.models import resnet18
+from torchvision.models import resnet18, resnet50
 from torchvision.transforms import Compose
 
 from health_cpath.utils.layer_utils import (get_imagenet_preprocessing,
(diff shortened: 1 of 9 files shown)


hi-ml-cpath/src/health_cpath/models/encoders.py
```diff
@@ -62,18 +62,21 @@ class ImageNetEncoder(TileEncoder):
...
 
     def __init__(self, feature_extraction_model: Callable[..., nn.Module],
-                 tile_size: int, n_channels: int = 3) -> None:
+                 tile_size: int, n_channels: int = 3, imagenet_preprocessing: bool = True) -> None:
         """
         :param feature_extraction_model: A function accepting a `pretrained` keyword argument that
         returns a classifier pretrained on ImageNet, such as the ones from `torchvision.models.*`.
         :param tile_size: Tile width/height, in pixels.
         :param n_channels: Number of channels in the tile (default=3).
         """
```
harshita-s, line 71: docstring needs to be updated

hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
```diff
@@ -38,6 +38,26 @@ def set_module_gradients_enabled(model: nn.Module, tuning_flag: bool) -> None:
         params.requires_grad = tuning_flag
 
 
+class Resnet18(ImageNetEncoder):
```
harshita-s, line 41: Tests need to be updated as there are a few codecov warnings

ant0nsc (approved):
Looking good so far. As suggested by Harshita, please add tests

hi-ml-cpath/src/health_cpath/models/encoders.py
kenza-bouzid: Done

harshita-s (approved):

hi-ml-cpath/src/health_cpath/models/encoders.py
```diff
@@ -62,18 +62,22 @@ class ImageNetEncoder(TileEncoder):
...
         self.create_feature_extractor_fn = feature_extraction_model
+        self.apply_imagenet_preprocessing = apply_imagenet_preprocessing
         super().__init__(tile_size=tile_size, n_channels=n_channels)
 
     def _get_preprocessing(self) -> Callable:
-        return get_imagenet_preprocessing()
+        if self.apply_imagenet_preprocessing:
+            return get_imagenet_preprocessing()
+        return super()._get_preprocessing()
```
vale-salvatelli, line 80: this risks to lead to errors - the _get_preprocessing of the super class is ignored if the flag is true. I know the method is basically empty at the moment but we could change with a composition or with and an if/else?

hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
```diff
@@ -38,6 +38,26 @@ def set_module_gradients_enabled(model: nn.Module, tuning_flag: bool) -> None:
...
+
+class Resnet50(ImageNetEncoder):
+    def __init__(self, tile_size: int, n_channels: int = 3) -> None:
+        super().__init__(resnet50, tile_size, n_channels, apply_imagenet_preprocessing=True)
+
+
+class Resnet50_NoPreproc(ImageNetEncoder):
+    def __init__(self, tile_size: int, n_channels: int = 3) -> None:
+        super().__init__(resnet50, tile_size, n_channels, apply_imagenet_preprocessing=False)
+
```
vale-salvatelli, line 59: Are these classes used in any config test? if we want tests to reflect the latest status of the codebase they should

kenza-bouzid: No, I didn't want to make a lot of changes (I will need to update the tests and quite a lot of config files).
I just added these new classes so that we can use them in cpath. I know that it's not ideal cause it introduces some code duplication, but we can maybe address that in a separate PR with a cleaner refactoring. I will add a ticket to the Board

PR #576, opened 2022-08-17
ENH: Switch to new AzureML runtime

ant0nsc:


hi-ml-azure/run_pytest.py
```diff
@@ -180,3 +181,4 @@ def pytest_after_submission_hook(azure_run: Run) -> None:
                 after_submission=pytest_after_submission_hook,
             )
     run_pytest(folder_to_test=config.folder, pytest_mark=config.mark, coverage_module=config.coverage_module)
+    time.sleep(10)  # Give the AzureML job time to finish uploading the pytest result file.
```
kenza-bouzid, line 184: Can you explain why do we need to wait now?
is 10 sec enough?

kenza-bouzid (approved):

hi-ml-azure/run_pytest.py
ant0nsc: I really don't know why that has become a problem now. It worked for me now, may need to adjust in the future.

peterhessey (approved):

PR #577, opened 2022-08-18
BUG: Fix SettingWithCopyWarning and parametrize metadata validation

kenza-bouzid:
Use df.assign to fix SettingWithCopyWarning  warning and make metadata validation optional to be able to read only subsets of dataset csv

### DIFF
--- a/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
+++ b/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
@@ -229,7 +229,8 @@ class SlidesDataset(Dataset):
         )
         columns = mandatory_columns.union(optional_columns)
         # SLIDE_ID_COLUMN is used for indexing and is not in df.columns anymore
-        columns_not_found = columns - set(self.dataset_df.columns) - {None} - {self.SLIDE_ID_COLUMN}
+        # None might be in columns if SPLITS_COLUMN is None
+        columns_not_found = columns - set(self.dataset_df.columns) - {None, self.SLIDE_ID_COLUMN}
         if len(columns_not_found) > 0:
             raise ValueError(f"Expected columns '{columns_not_found}' not found in the dataframe")
 
--- a/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
+++ b/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
@@ -160,6 +160,7 @@ class SlidesDataset(Dataset):
                  dataset_df: Optional[pd.DataFrame] = None,
                  train: Optional[bool] = None,
                  validate_columns: bool = True,
+                 validate_metadata: bool = True,
                  label_column: str = DEFAULT_LABEL_COLUMN,
                  n_classes: int = 1,
                  dataframe_kwargs: Dict[str, Any] = {}) -> None:
@@ -175,6 +176,8 @@ class SlidesDataset(Dataset):
         :param validate_columns: Whether to call `validate_columns()` at the end of `__init__()`.
         `validate_columns()` checks that the loaded data frame for the dataset contains the expected column names
         for this class
+        :param validate_metadata: Whether to include metadata columns validation in `validate_columns()`. In some cases
+        when only a subset of metadata columns is loaded via custom `dataframe_kwargs`, this can be set to `False` to avoid unnecessary checks.
         :param label_column: CSV column name for tile label. Default is `DEFAULT_LABEL_COLUMN="label"`.
         :param n_classes: Number of classes indexed in `label_column`. Default is 1 for binary classification.
         :param dataframe_kwargs: Keyword arguments to pass to `pd.read_csv()` when loading the dataset CSV.
@@ -185,6 +188,7 @@ class SlidesDataset(Dataset):
         self.root_dir = Path(root)
         self.label_column = label_column
         self.n_classes = n_classes
+        self.validate_metadata = validate_metadata
 
         if dataset_df is not None:
             self.dataset_csv = None
@@ -210,6 +214,8 @@ class SlidesDataset(Dataset):
         call `validate_columns()` after creating derived columns, for example.
         """
         columns = [self.IMAGE_COLUMN, self.label_column, self.MASK_COLUMN, self.SPLIT_COLUMN]
+        if self.validate_metadata:
+            columns.extend(list(self.METADATA_COLUMNS))
         columns_not_found = []
         for column in columns:
             if column is not None and column not in self.dataset_df.columns:
--- a/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
+++ b/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
@@ -12,7 +12,7 @@ import torch
 from sklearn.utils.class_weight import compute_class_weight
 from torch.utils.data import Dataset
 
-from health_cpath.utils.naming import SlideKey
+from health_cpath.utils.naming import SlideKey, TileKey
 
 
 DEFAULT_TRAIN_SPLIT_LABEL = "train"  # Value used to indicate the training split in `SPLIT_COLUMN`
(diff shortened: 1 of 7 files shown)


hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
```diff
@@ -209,8 +214,9 @@ def validate_columns(self) -> None:
         If the constructor is overloaded in a subclass, you can pass `validate_columns=False` and
         call `validate_columns()` after creating derived columns, for example.
         """
-        columns = [self.IMAGE_COLUMN, self.label_column, self.MASK_COLUMN,
-                   self.SPLIT_COLUMN] + list(self.METADATA_COLUMNS)
+        columns = [self.IMAGE_COLUMN, self.label_column, self.MASK_COLUMN, self.SPLIT_COLUMN]
+        if self.validate_metadata:
+            columns.extend(list(self.METADATA_COLUMNS))
```
mebristo, line 219: Rather than having an extra flag, could we not just do:
```
if dataframe_kwargs is not empty:
    udpdate`self.METADATA_COLUMNS` to include only the ones that are in dataframe_kwargs
```
at some point in the pipeline?

hi-ml-cpath/src/health_cpath/datasets/panda_tiles_dataset.py
```diff
@@ -76,9 +76,9 @@ def __init__(self,
         # Copy columns "left" --> "tile_x" and "top" --> "tile_y"
         # to be consistent with TilesDataset `TILE_X_COLUMN` and `TILE_Y_COLUMN`
         if TileKey.TILE_LEFT in self.dataset_df.columns:
-            self.dataset_df[TilesDataset.TILE_X_COLUMN] = self.dataset_df[TileKey.TILE_LEFT]
+            self.dataset_df = self.dataset_df.assign(tile_x=self.dataset_df[TileKey.TILE_LEFT])
```
mebristo, line 79: Should this not be `TilesDataset.TILE_X_COLUMN`?

hi-ml-cpath/src/health_cpath/datasets/panda_tiles_dataset.py
```diff
@@ -76,9 +76,9 @@ def __init__(self,
...
         if TileKey.TILE_TOP in self.dataset_df.columns:
-            self.dataset_df[TilesDataset.TILE_Y_COLUMN] = self.dataset_df[TileKey.TILE_TOP]
+            self.dataset_df = self.dataset_df.assign(tile_y=self.dataset_df[TileKey.TILE_TOP])
```
mebristo, line 81: Similarly, TilesDataset.TILE_Y_COLUMN?

hi-ml-cpath/testhisto/testhisto/datasets/test_panda_tiles.py
```diff
@@ -0,0 +1,40 @@
...
+    base_df_columns = set(base_df.columns)
+
+    assert coordinates_columns_v0.issubset(dataset_columns)   # v0 columns are always present
+
+    if version == 0:
+        assert base_df_columns == dataset_columns
+        assert not coordinates_columns_v1.issubset(dataset_columns)
+    elif version == 1:
+        assert dataset_columns == base_df_columns.union(coordinates_columns_v0)
+        assert not coordinates_columns_v0.issubset(base_df_columns)
```
mebristo, line 40: It's tricky to follow what checking here, especially since I'm not sure what version 0/ 1 means. Could you add a comment that is something like the following:
'if version 0, ensure that only the coordinate columns are in the dataset columns. If version 1, check that columns have been added to the original dataframe columns during dataset creation. ' (if I have understood that correctly)

Also - we only check that coordinate_columns_v1 are _not_ in version 0 dataframe. Would we expect that they _are_ in version 1 dataframe and if so, should we assert that too?

hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
kenza-bouzid: but dataframe_kwargs is not only specific to usecols, it's meant to be used to override all [pd.read_csv](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html) arguments

hi-ml-cpath/testhisto/testhisto/mocks/tiles_generator.py
```diff
@@ -10,18 +10,22 @@
...
 
 
 class MockPandaTilesGenerator(MockHistoDataGenerator):
     """Generator class to create mock tiles dataset on the fly. The tiles are positioned randomly in a wsi grid."""
 
-    def __init__(self, img_size: int = 224, **kwargs: Any) -> None:
+    def __init__(self, img_size: int = 224, version: int = 0, **kwargs: Any) -> None:
         """
         :param img_size: The whole slide image resolution, defaults to 224.
+        :param version: The version of the tiles dataset, defaults to 0. This is used to support both the old and new
```
mebristo, line 23: Why do we still support the old tiling scheme?

hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
kenza-bouzid: I can maybe check if usecols is in dataframe_kwargs there and maybe validate only the columns defined there?

hi-ml-cpath/src/health_cpath/datasets/panda_dataset.py
```diff
@@ -42,9 +42,10 @@ def __init__(self,
                  dataset_csv: Optional[Union[str, Path]] = None,
                  dataset_df: Optional[pd.DataFrame] = None,
                  label_column: str = "isup_grade",
-                 n_classes: int = 6) -> None:
+                 n_classes: int = 6,
+                 **kwargs: Any) -> None:
```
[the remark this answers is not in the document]
kenza-bouzid: It's mainly because we were adding extra params in the base class recently and forgot to propagate them to child classes, using **kwargs makes it easier to extend base classes args. Or should I add all params in base class?  (Note we should keep in mind updating subclasses in the future)

hi-ml-cpath/src/health_cpath/datasets/panda_tiles_dataset.py
kenza-bouzid: It didn't work :( , assign expects a param name not a string/enum :(

hi-ml-cpath/src/health_cpath/datasets/panda_tiles_dataset.py
kenza-bouzid: Same as above

hi-ml-cpath/testhisto/testhisto/mocks/tiles_generator.py
kenza-bouzid: Not sure tbh, I think it was mainly to be able to work with old datasets that were created with the old tiling script
Also in deepmil we support both 
image

hi-ml-cpath/testhisto/testhisto/datasets/test_panda_tiles.py
kenza-bouzid: version 0 and 1 correspond to the old/new tiling. 
Coordinates are present in both versions they are just named differently, v0 we have tile_x and tile_y, v1 we have left and top which is meant to be x and y of v0.

In v1 we just copy them over to x and y because some part of the code base still uses x and y I think

hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
mebristo: Sorry yes that's what I mean

hi-ml-cpath/src/health_cpath/datasets/panda_tiles_dataset.py
mebristo: In that case you should be able to do `**{TilesDataset.TILE_X_COLUMN.value: self.dataset_df[TileKey.TILE_LEFT]})`?

hi-ml-cpath/src/health_cpath/datasets/panda_dataset.py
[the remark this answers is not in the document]
mebristo: I guess it depends how many params we are talking about adding...can discuss offline

hi-ml-cpath/src/health_cpath/datasets/panda_tiles_dataset.py
kenza-bouzid: Good point, will try it!

hi-ml-cpath/testhisto/testhisto/datasets/test_panda_tiles.py
ant0nsc: +1 on adding more comments. It is extremely helpful to have comments in tests, I'd happily read a whole story.

ant0nsc (approved):
Melissa raised a couple of good points, nothing to add beyond those.

mebristo (approved):

PR #578, opened 2022-08-18
ENH: Add/reorganise metrics and correctly normalise confusion matrix

harshita-s:
In this PR:
- Average precision metric is added to multi-class (n_classes > 1) case.
- AUROC is modified such that `num_classes=None` for binary case, as prescribed in Pytorch documentation https://torchmetrics.readthedocs.io/en/stable/classification/auroc.html.
- Parameter `num_classes` is not explicitly given in Specificity, similar to the other binary metrics.
- Hardcoded `threshold=0.5` is removed from binary metrics (since this is default value).
- Confusion matrix is normalized on true values.
- Metrics are re-organized for readability.

### DIFF
--- a/hi-ml-cpath/src/health_cpath/models/deepmil.py
+++ b/hi-ml-cpath/src/health_cpath/models/deepmil.py
@@ -203,19 +203,18 @@ class BaseDeepMILModule(LightningModule):
                 MetricsKey.ACC_MACRO: Accuracy(num_classes=self.n_classes, average='macro'),
                 MetricsKey.ACC_WEIGHTED: Accuracy(num_classes=self.n_classes, average='weighted')})
         else:
-            threshold = 0.5
             return nn.ModuleDict({
-                MetricsKey.ACC: Accuracy(threshold=threshold),
+                MetricsKey.ACC: Accuracy(),
                 MetricsKey.AUROC: AUROC(num_classes=None),
                 # Average precision is a measure of area under the PR curve
                 MetricsKey.AVERAGE_PRECISION: AveragePrecision(),
-                MetricsKey.COHENKAPPA: CohenKappa(num_classes=2, weights='quadratic', threshold=threshold),
-                MetricsKey.CONF_MATRIX: ConfusionMatrix(num_classes=2, threshold=threshold),
+                MetricsKey.COHENKAPPA: CohenKappa(num_classes=2, weights='quadratic'),
+                MetricsKey.CONF_MATRIX: ConfusionMatrix(num_classes=2),
                 # Metrics below are computed for binary case only
-                MetricsKey.F1: F1(threshold=threshold),
-                MetricsKey.PRECISION: Precision(threshold=threshold),
-                MetricsKey.RECALL: Recall(threshold=threshold),
-                MetricsKey.SPECIFICITY: Specificity(threshold=threshold)})
+                MetricsKey.F1: F1(),
+                MetricsKey.PRECISION: Precision(),
+                MetricsKey.RECALL: Recall(),
+                MetricsKey.SPECIFICITY: Specificity()})
 
     def log_metrics(self, stage: str) -> None:
         valid_stages = [stage for stage in ModelKey]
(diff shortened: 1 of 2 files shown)


hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -190,29 +190,32 @@ def get_bag_label(labels: Tensor) -> Tensor:
...
-                                  MetricsKey.F1: F1(threshold=threshold),
-                                  MetricsKey.CONF_MATRIX: ConfusionMatrix(num_classes=2, threshold=threshold),
-                                  # Average precision is a measure of area under the PR curve
-                                  # https://sanchom.wordpress.com/tag/average-precision/
-                                  MetricsKey.AVERAGE_PRECISION: AveragePrecision(),
-                                  MetricsKey.COHENKAPPA: CohenKappa(num_classes=2, weights='quadratic',
-                                                                    threshold=threshold),
-                                  MetricsKey.SPECIFICITY: Specificity(num_classes=self.n_classes, threshold=threshold)})
+            return nn.ModuleDict({
+                MetricsKey.ACC: Accuracy(threshold=threshold),
```
vale-salvatelli, line 208: I am still confused about what would happen in n_classes = 2 vs n_classes = 1. I think our pipeline should give equivalent results because in both cases we would be working with a binary label while here if n_classes=1 we take a threshold of 0.5, while if n_classes=2 we pass num_classes=2

harshita-s: I am investigating this aspect.

kenza-bouzid: Note that threshold is always 0.5 by default in all metrics, not sure if we should override it for binary case given that it will be set to 0.5 by default in multi class as well. (Sorry for jumping in I was just curious)

harshita-s: Yes, I am aware of the default values. I am not sure why these were hardcoded in the first instance, perhaps @vale-salvatelli could clarify?

harshita-s: I have now removed the hardcoded threshold for binary case

hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -190,29 +190,31 @@ def get_bag_label(labels: Tensor) -> Tensor:
...
-                                  # https://sanchom.wordpress.com/tag/average-precision/
-                                  MetricsKey.AVERAGE_PRECISION: AveragePrecision(),
-                                  MetricsKey.COHENKAPPA: CohenKappa(num_classes=2, weights='quadratic',
-                                                                    threshold=threshold),
-                                  MetricsKey.SPECIFICITY: Specificity(num_classes=self.n_classes, threshold=threshold)})
+            return nn.ModuleDict({
+                MetricsKey.ACC: Accuracy(),
+                MetricsKey.AUROC: AUROC(num_classes=None),
+                # Average precision is a measure of area under the PR curve
+                MetricsKey.AVERAGE_PRECISION: AveragePrecision(),
```
vale-salvatelli, line 210: why we don't pass self.n_classes in this case?

vale-salvatelli (approved):
I think this PR doesn't solve the inconsistency between num_classes 1 and num_classes 2 but it contains some fixes we should merge soon - OK to merge for me. 

I would probably change the title to say it also fixes the confusion matrix

hi-ml-cpath/src/health_cpath/models/deepmil.py
harshita-s: I followed the docs https://torchmetrics.readthedocs.io/en/stable/classification/accuracy.html to determine the parameters for each case.

harshita-s:
> I think this PR doesn't solve the inconsistency between num_classes 1 and num_classes 2 but it contains some fixes we should merge soon - OK to merge for me.
> 
> I would probably change the title to say it also fixes the confusion matrix

The class discrepency will be handled in a different PR or summarized on OneNote.

PR #587, opened 2022-08-31
ENH: Enable stain normalization for multiple images

harshita-s:
Enable stain normalization to work with multiple images (e.g., when a bag of tiles is given as input).

### DIFF
--- a/hi-ml/src/health_ml/utils/data_augmentations.py
+++ b/hi-ml/src/health_ml/utils/data_augmentations.py
@@ -131,7 +131,14 @@ class StainNormalization(object):
         return nimg
 
     def __call__(self, img: torch.Tensor) -> torch.Tensor:
-        return self.stain_normalize(img, self.reference_mean, self.reference_std)
+        # if the input is a bag of images, stain normalization needs to run on each image separately
+        if img.shape[0] > 1:
+            for i in range(img.shape[0]):
+                img_tile = img[i]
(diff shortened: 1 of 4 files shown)


hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -131,7 +131,14 @@ def stain_normalize(img: torch.Tensor, reference_mean: np.ndarray, reference_std
...
     def __call__(self, img: torch.Tensor) -> torch.Tensor:
-        return self.stain_normalize(img, self.reference_mean, self.reference_std)
+        # if the input is a bag of images, stain normalization needs to run on each image separately
+        if img.shape[0] > 1:
+            for i in range(img.shape[0]):
+                img_tile = img[i]
+                img[i] = self.stain_normalize(img_tile.unsqueeze(0), self.reference_mean, self.reference_std)
+            return img
+        else:
+            return self.stain_normalize(img, self.reference_mean, self.reference_std)
```
vale-salvatelli, line 141: Can we extend the tests to cover this case?

vale-salvatelli (changes requested):
Can we extend the tests to cover this case?

harshita-s:
Extended the test to cover the case
> Can we extend the tests to cover this case?

vale-salvatelli (approved):

PR #588, opened 2022-09-02
FIX: Update env var settings for multi node

mebristo:
Changes in the way that Azure ML set the environment variable `ENV_AZ_BATCHAI_MPI_MASTER_NODE` mean that the logic for setting environment variables for multi-node jobs needs updating

### DIFF
--- a/hi-ml-azure/src/health_azure/utils.py
+++ b/hi-ml-azure/src/health_azure/utils.py
@@ -1161,14 +1161,16 @@ def set_environment_variables_for_multi_node() -> None:
     Sets the environment variables that PyTorch Lightning needs for multi-node training.
     """
     if ENV_AZ_BATCH_MASTER_NODE in os.environ:
-        print("Found ENV_AZ_BATCH_MASTER_NODE")
+        logging.debug("Found ENV_AZ_BATCH_MASTER_NODE")
         # For AML BATCHAI
         split_master_node_addr = os.environ[ENV_AZ_BATCH_MASTER_NODE].split(":")
-        if len(split_master_node_addr) > 1:
+        if len(split_master_node_addr) == 2:
             master_addr, port = split_master_node_addr
             os.environ[ENV_MASTER_PORT] = port
-        else:
+        elif len(split_master_node_addr) == 1:
             master_addr = split_master_node_addr[0]
+        else:
+            raise ValueError(f"Format not recognized: {os.environ[ENV_AZ_BATCH_MASTER_NODE]}")
         os.environ[ENV_MASTER_ADDR] = master_addr
     elif ENV_AZ_BATCHAI_MPI_MASTER_NODE in os.environ and os.environ.get(ENV_AZ_BATCHAI_MPI_MASTER_NODE) != "localhost":
         # For AML BATCHAI
--- a/hi-ml-azure/src/health_azure/utils.py
+++ b/hi-ml-azure/src/health_azure/utils.py
@@ -1161,23 +1161,31 @@ def set_environment_variables_for_multi_node() -> None:
     Sets the environment variables that PyTorch Lightning needs for multi-node training.
     """
     if ENV_AZ_BATCH_MASTER_NODE in os.environ:
-        logging.debug("Found ENV_AZ_BATCH_MASTER_NODE")
+        master_node = os.environ[ENV_AZ_BATCH_MASTER_NODE]
+        logging.debug(
+            f"Found AZ_BATCH_MASTER_NODE: {master_node} in environment variables")
         # For AML BATCHAI
-        split_master_node_addr = os.environ[ENV_AZ_BATCH_MASTER_NODE].split(":")
+        split_master_node_addr = master_node.split(":")
         if len(split_master_node_addr) == 2:
             master_addr, port = split_master_node_addr
             os.environ[ENV_MASTER_PORT] = port
         elif len(split_master_node_addr) == 1:
             master_addr = split_master_node_addr[0]
         else:
-            raise ValueError(f"Format not recognized: {os.environ[ENV_AZ_BATCH_MASTER_NODE]}")
+            raise ValueError(f"Format not recognized: {master_node}")
         os.environ[ENV_MASTER_ADDR] = master_addr
     elif ENV_AZ_BATCHAI_MPI_MASTER_NODE in os.environ and os.environ.get(ENV_AZ_BATCHAI_MPI_MASTER_NODE) != "localhost":
+        mpi_master_node = os.environ[ENV_AZ_BATCHAI_MPI_MASTER_NODE]
+        logging.debug(
+            f"Found AZ_BATCHAI_MPI_MASTER_NODE: {mpi_master_node} in environment variables")
         # For AML BATCHAI
-        os.environ[ENV_MASTER_ADDR] = os.environ[ENV_AZ_BATCHAI_MPI_MASTER_NODE]
+        os.environ[ENV_MASTER_ADDR] = mpi_master_node
     elif ENV_MASTER_IP in os.environ:
+        master_ip = os.environ[ENV_MASTER_IP]
+        logging.debug(
+            f"Found MASTER_IP: {master_ip} in environment variables")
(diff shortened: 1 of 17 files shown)


hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1158,7 +1160,17 @@ def set_environment_variables_for_multi_node() -> None:
     """
     Sets the environment variables that PyTorch Lightning needs for multi-node training.
     """
-    if ENV_AZ_BATCHAI_MPI_MASTER_NODE in os.environ:
+    if ENV_AZ_BATCH_MASTER_NODE in os.environ:
+        print("Found ENV_AZ_BATCH_MASTER_NODE")
```
ant0nsc, line 1164:
```suggestion
        logging.debug("Found ENV_AZ_BATCH_MASTER_NODE")
```

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1158,7 +1160,17 @@ def set_environment_variables_for_multi_node() -> None:
...
+        # For AML BATCHAI
+        split_master_node_addr = os.environ[ENV_AZ_BATCH_MASTER_NODE].split(":")
+        if len(split_master_node_addr) > 1:
+            master_addr, port = split_master_node_addr
```
ant0nsc, line 1168:
```suggestion
        if len(split_master_node_addr) == 2:
            master_addr, port = split_master_node_addr
```

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1158,7 +1160,17 @@ def set_environment_variables_for_multi_node() -> None:
...
+            os.environ[ENV_MASTER_PORT] = port
+        else:
+            master_addr = split_master_node_addr[0]
```
ant0nsc, line 1171:
```suggestion
        elif len(split_master_node_addr) == 1:
            master_addr = split_master_node_addr[0]
        else:
            raise ValueError(f"Format not recognized: {os.environ[ENV_AZ_BATCH_MASTER_NODE]}")
```

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1169,12 +1181,13 @@ def set_environment_variables_for_multi_node() -> None:
...
 
     if ENV_MASTER_PORT not in os.environ:
-        os.environ[ENV_MASTER_PORT] = "6105"
+        os.environ[ENV_MASTER_PORT] = str(MASTER_PORT_DEFAULT)
 
     if ENV_OMPI_COMM_WORLD_RANK in os.environ:
         os.environ[ENV_NODE_RANK] = os.environ[ENV_OMPI_COMM_WORLD_RANK]  # node rank is the world_rank from mpi run
     env_vars = ", ".join(f"{var} = {os.environ[var]}" for var in [ENV_MASTER_ADDR, ENV_MASTER_PORT, ENV_NODE_RANK])
-    print(f"Distributed training: {env_vars}")
+    print(env_vars)
```
ant0nsc, line 1189: Remove the print statement?

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -1158,7 +1160,17 @@ def set_environment_variables_for_multi_node() -> None:
     """
     Sets the environment variables that PyTorch Lightning needs for multi-node training.
     """
-    if ENV_AZ_BATCHAI_MPI_MASTER_NODE in os.environ:
+    if ENV_AZ_BATCH_MASTER_NODE in os.environ:
```
ant0nsc, line 1163: for easier diagnostics, should we print all environment variables via logging.debug?

hi-ml-azure/testazure/testazure/test_azure_util.py
```diff
@@ -811,35 +812,123 @@ def test_set_environment_variables_for_multi_node(
...
-                util.ENV_MASTER_ADDR: "else",
-            },
-            clear=True):
-        util.set_environment_variables_for_multi_node()
-    out = capsys.readouterr().out
-    assert "Distributed training: MASTER_ADDR = here, MASTER_PORT = there, NODE_RANK = everywhere" in out
+    # If all of ENV_MASTER_IP, AZ_BATCHAI_MPI_MASTER_NODE and AZ_BATCH_MASTER_NODE are set, the latter should
+    # take precedence. NODE_RANK should get updated to the value of ENV_OMPI_COMM_WORLD_RANK
+    mock_az_batch_master_port = str(randint(1000, MASTER_PORT_DEFAULT - 1))  # Avoid setting to the default value
+    mock_node_rank = str(randint(0, 10))
```
ant0nsc, line 825: I am not a fan of random values in tests if it is not absolutely necessary. Tests should be reproducible as much as possible.

fepegar (changes requested):
LGTM, but some tests are failing.

fepegar:
> LGTM, but some tests are failing.

Also, @Shruthi42 is probably a much more appropriate reviewer than I am for this PR.

hi-ml-azure/testazure/testazure/test_azure_util.py
```diff
@@ -811,35 +812,123 @@ def test_set_environment_variables_for_multi_node(
...
+
+    mock_az_batch_master_addr = "1234.0.0.0"
+    mock_az_batch_master_node = f"{mock_az_batch_master_addr}:{mock_az_batch_master_port}"
+
+    mock_az_batch_mpi_master_addr = "5678.9.9.9"
+    mock_az_batch_mpi_master_node = f"{mock_az_batch_mpi_master_addr}"
```
ant0nsc, line 831: Nitpick: I'm having a hard time reading the tests because many variables have the same prefix - causing the same confusion that we had yesterday when analyzing the problem. How about simplifying the names to be something like "master", "master_and_port", "port"?

ant0nsc (changes requested):
Looking good! I added suggestions for corner cases.
Also, you had another safety check in the runner, to only call the function when we are really doing multi-node?
Should we add a multi-node training job to the hi-ml test suite in a follow-up PR (happy to work on that myself)?

mebristo:
> > LGTM, but some tests are failing.
> 
> Also, @Shruthi42 is probably a much more appropriate reviewer than I am for this PR.

Good to know, thanks!

mebristo:
> Looking good! I added suggestions for corner cases. Also, you had another safety check in the runner, to only call the function when we are really doing multi-node? Should we add a multi-node training job to the hi-ml test suite in a follow-up PR (happy to work on that myself)?

Good idea to add the multi-node training job in a follow-up PR

hi-ml-azure/src/health_azure/utils.py
ant0nsc: I still see the print statement in line 1202, even though we print the same info to logging in the line below. Is that by accident or design?

hi-ml-azure/testazure/testazure/test_azure_util.py
```diff
@@ -821,114 +819,161 @@ def test_set_environment_variables_for_multi_node(
 
     # If all of ENV_MASTER_IP, AZ_BATCHAI_MPI_MASTER_NODE and AZ_BATCH_MASTER_NODE are set, the latter should
     # take precedence. NODE_RANK should get updated to the value of ENV_OMPI_COMM_WORLD_RANK
-    mock_az_batch_master_port = str(randint(1000, MASTER_PORT_DEFAULT - 1))  # Avoid setting to the default value
-    mock_node_rank = str(randint(0, 10))
+    port_mock = str(MASTER_PORT_DEFAULT)  # Avoid setting to the default value
```
ant0nsc, line 822: Good point about not setting to the default value, but the code is no longer doing that:
```suggestion
    port_mock = str(MASTER_PORT_DEFAULT-1)  # Avoid setting to the default value
```

ant0nsc (approved):
Looking good! I'm still confused why codecov is complaining about lots of uncovered lines, your test looks very comprehensive?

fepegar (approved):

PR #589, opened 2022-09-02
ENH: Upgrade hi-ml-cpath to Python 3.9

ant0nsc:
All code changes in this PR in the SSL codebase are necessary because Python 3.9 does not allow to update an `Enum` - but the SSL codebase does precisely that. I've switched the enum to a class with constants. Dataset names are now plain strings, which makes extending to new datasets a lot cleaner.


mebristo:
Why not straight to version 3.10?

hi-ml-cpath/primary_deps.yml
```diff
@@ -6,7 +6,7 @@ channels:
 dependencies:
   - cudatoolkit=11.3.1
   - pip=20.1.1
-  - python=3.7.3
+  - python=3.9.10
```
fepegar, line 9: Isn't 3.9.13 the latest patch?

hi-ml-cpath/src/SSL/data/cxr_datasets.py
```diff
@@ -191,7 +191,7 @@ def _prepare_dataset(self) -> None:
         mapping = {0: 0, 3: 0, 1: 1, 2: 1}
         # For monitoring purpose with use binary classification CV03vsCV12
         self.dataset_dataframe["final_label"] = self.dataset_dataframe.final_label.apply(lambda x: mapping[x])
-        self.indices = np.arange(len(self.dataset_dataframe))
+        self.indices = np.arange(len(self.dataset_dataframe)).tolist()
```
fepegar, line 194: Why are these changes needed? Is it because of the new NumPy version?

hi-ml-cpath/src/SSL/encoders.py
```diff
@@ -91,4 +91,4 @@ def get_encoder_output_dim(
     with torch.no_grad():
         representations = pl_module(x)
 
-    return representations.shape[1]
+    return representations.shape[1]  # type: ignore
```
fepegar, line 94: Unrelated to this PR, but this is something I've been doing recently for TorchIO which I think is nicer than ignoring all errors: https://mypy.readthedocs.io/en/stable/error_codes.html#silencing-errors-based-on-error-codes

fepegar (approved):
Exciting! Thanks, Anton.

ant0nsc:
> Why not straight to version 3.10?

@mebristo , because some AzureML packages are not yet supported by 3.10

hi-ml-cpath/src/SSL/data/cxr_datasets.py
ant0nsc: `mypy` finds different errors with the new Python versions. We declared indices as being a list, but `mypy` so far did not complain about assigning an `ndarray` to that list. Now it does.

hi-ml-cpath/primary_deps.yml
ant0nsc: Good point. fixing.

hi-ml-cpath/src/SSL/configs/PANDA_SimCLRContainer.py
```diff
@@ -2,19 +2,17 @@
...
 from SSL.utils import SSLTrainingType
 from health_azure.utils import is_running_in_azure_ml
 from health_cpath.datasets.panda_tiles_dataset import PandaTilesDatasetWithReturnIndex
 from health_cpath.datasets.default_paths import PANDA_5X_TILES_DATASET_ID
 from SSL.configs.HistoSimCLRContainer import HistoSSLContainer
 
 
-class SSLDatasetNameHiml(SSLDatasetName, Enum):  # type: ignore
-    PANDA = "PandaTilesDataset"
+SSL_Dataset_PANDA = "PandaTilesDataset"
```
mebristo, line 15: Should we name this `SSL_Dataset_PANDA_Tiles` for clarity?

mebristo (approved):

PR #590, opened 2022-09-02
ENH: Dropout in transformer layers in DeepMIL

harshita-s:
In this PR:
- `transformer_dropout` parameter is added to `TransformerPooling` and `TransformerPoolingBenchmark` pooling layers.
- Tests are updated with the `transformer_dropout` parameter.

### DIFF
--- a/hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
+++ b/hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
@@ -132,7 +132,7 @@ class PoolingParams(param.Parameterized):
         "layer. Else (False), initialize the pooling layer randomly.",
     )
     transformer_dropout: float = param.Number(
-        default=0.0,
+        default=0.1,
         doc="If transformer pooling is chosen, this defines the dropout of the tranformer encoder layers.",
     )
 
--- a/hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
+++ b/hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
@@ -146,27 +146,31 @@ class PoolingParams(param.Parameterized):
         """
         pooling_layer: nn.Module
         if self.pool_type == AttentionLayer.__name__:
-            pooling_layer = AttentionLayer(num_encoding, self.pool_hidden_dim, self.pool_out_dim)
+            pooling_layer = AttentionLayer(input_dims=num_encoding,
+                                           hidden_dims=self.pool_hidden_dim,
+                                           attention_dims=self.pool_out_dim)
         elif self.pool_type == GatedAttentionLayer.__name__:
-            pooling_layer = GatedAttentionLayer(num_encoding, self.pool_hidden_dim, self.pool_out_dim)
+            pooling_layer = GatedAttentionLayer(input_dims=num_encoding,
+                                                hidden_dims=self.pool_hidden_dim,
+                                                attention_dims=self.pool_out_dim)
         elif self.pool_type == MeanPoolingLayer.__name__:
             pooling_layer = MeanPoolingLayer()
         elif self.pool_type == MaxPoolingLayer.__name__:
             pooling_layer = MaxPoolingLayer()
         elif self.pool_type == TransformerPooling.__name__:
             pooling_layer = TransformerPooling(
-                self.num_transformer_pool_layers,
-                self.num_transformer_pool_heads,
-                num_encoding,
-                self.transformer_dropout)
+                num_layers=self.num_transformer_pool_layers,
+                num_heads=self.num_transformer_pool_heads,
+                dim_representation=num_encoding,
+                transformer_dropout=self.transformer_dropout)
             self.pool_out_dim = 1  # currently this is hardcoded in forward of the TransformerPooling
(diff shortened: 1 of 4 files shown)


hi-ml/testhiml/testhiml/test_attentionlayers.py
```diff
@@ -23,6 +23,8 @@ def _test_attention_layer(attentionlayer: nn.Module, dim_in: int, dim_att: int,
         pass
     elif isinstance(attentionlayer, MaxPoolingLayer):
         pass
+    elif isinstance(attentionlayer, TransformerPoolingBenchmark):
```
dccastro, line 26: [Minor] To avoid having too many cases here, we can instead use a single `isinstance()` with a tuple of all the types we want to check at once:
```python
if not isinstance(attentionlayer, (Class1, Class2, ...)):
    pooled_features = attn_weights @ features.flatten(start_dim=1)
    assert allclose(pooled_features, output_features)
```

hi-ml/testhiml/testhiml/test_attentionlayers.py
```diff
@@ -59,8 +61,18 @@ def test_max_pooling(dim_in: int, batch_size: int,) -> None:
 @pytest.mark.parametrize("num_heads", [1, 2])
 @pytest.mark.parametrize("dim_in", [4, 8])   # dim_in % num_heads must be 0
 @pytest.mark.parametrize("batch_size", [1, 7])
-def test_transformer_pooling(num_layers: int, num_heads: int, dim_in: int, batch_size: int) -> None:
+@pytest.mark.parametrize("transformer_dropout", [0.0, 0.3])
```
dccastro, line 64: We now have 64 combinations of inputs for this test. Do you think it might make sense to test the dropout options in a separate test function, keeping this one focused on compatibility of tensor shapes?

hi-ml/testhiml/testhiml/test_attentionlayers.py
```diff
@@ -59,8 +61,18 @@ def test_max_pooling(dim_in: int, batch_size: int,) -> None:
...
+@pytest.mark.parametrize("dim_hid", [1, 4])
+def test_transformer_pooling(num_layers: int, num_heads: int, dim_in: int,
+                             batch_size: int, dim_hid: int, transformer_dropout: float) -> None:
     transformer_pooling = TransformerPooling(num_layers=num_layers,
                                              num_heads=num_heads,
-                                             dim_representation=dim_in).eval()
+                                             dim_representation=dim_in,
+                                             transformer_dropout=transformer_dropout).eval()
     _test_attention_layer(transformer_pooling, dim_in=dim_in, dim_att=1, batch_size=batch_size)
+    transformer_pooling_benchmark = TransformerPoolingBenchmark(num_layers=num_layers,
```
dccastro, line 73: I'd suggest testing this component in a separate test function or test case, so we can catch errors separately for both. Also, they have different input arguments.

dccastro (approved):

hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
```diff
@@ -150,13 +155,18 @@ def get_pooling_layer(self, num_encoding: int) -> Tuple[nn.Module, int]:
...
             self.pool_out_dim = 1  # currently this is hardcoded in forward of the TransformerPooling
         elif self.pool_type == TransformerPoolingBenchmark.__name__:
             pooling_layer = TransformerPoolingBenchmark(
-                self.num_transformer_pool_layers, self.num_transformer_pool_heads, num_encoding, self.pool_hidden_dim
-            )
+                self.num_transformer_pool_layers,
+                self.num_transformer_pool_heads,
+                num_encoding,
+                self.pool_hidden_dim,
+                self.transformer_dropout)
```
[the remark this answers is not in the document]
harshita-s: Sure! Now passing keyword arguments

hi-ml/testhiml/testhiml/test_attentionlayers.py
ant0nsc: +1 on avoiding combinatorial explosion

hi-ml/testhiml/testhiml/test_attentionlayers.py
harshita-s: Thanks, now separated the tests

hi-ml/testhiml/testhiml/test_attentionlayers.py
harshita-s: We currently have one test function for each type of pooling layer, so I don't think we should have another function only `transformer_dropout` which is being used in 2/6 layers. To prevent too many combinations, I have fixed `transformer_dropout` parameter to one value for now. I think we could test it without parametrizing. What do you think?

hi-ml/testhiml/testhiml/test_attentionlayers.py
harshita-s: Thanks, done

hi-ml/testhiml/testhiml/test_attentionlayers.py
dccastro: It should be fine to hard-code the dropout rate, as we don't expect any substantial behavioural changes.

PR #595, opened 2022-09-06
ENH: Add shadow workflow for hi-ml-pr

mebristo:
Currently, status checks don't run on the PR pipelines because they involve path filtering. This PR follows the suggestion in [github documentation](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks) to add a second workflow that runs on the inverse of the filtered paths, to enable status checks to run


mebristo:
Not really ready for review, but can't see whether merging would be blocked until the PR is marked as ready

ant0nsc:
There's one other thing we should probably do in this PR: Give all jobs a unique name. We have, for example, 3 jobs that are called flake8, one for each project. When adding `flake8` as a necessary check, which `flake8` are we getting? So could rename them the `flake8-hi-ml`, `flake8-cpath`, `flake8-multimodal`. Or `hi-ml-flake8`, `cpath-flake8`, ... - Not sure which of the two is better, but unique names for sure.

.github/workflows/cpath-pr-shadow.yml
```diff
@@ -0,0 +1,53 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - run: 'echo "No build required" '
+
+  smoke_test_slidespandaimagenetmil:
+    runs-on: ubuntu-20.04
+    steps:
+      - run: 'echo "No build required" '
+
+  smoke_test_tilespandaimagenetmil:
```
kenza-bouzid, line 35: is there a reason why we only selected these smoke tests as part of the shadow workflow? I believe this PR has been open for a while, we added other smoke tests recently, do we need smoke tests in the shadow workflow anyways?

kenza-bouzid: Also may I suggest renaming smoke tests? can we come up with a standard naming pattern for all? some have aml at the end, some have hyphens in between words some don't... we haven't agreed on a naming pattern before, would be great to add it in the docs somewhere to follow the same for future smoke tests.

ant0nsc:
Just discussed with @peterhessey that we could simplify the shadow workflows by only running the "final" step of each PR workflow, like `publish testpypi` for `hi-ml`. This way, we don't have to replicate (and maintain) the full complex graph of each PR, only need the final step that depends on the success of all previous ones.

.github/workflows/cpath-pr-shadow.yml
peterhessey: @kenza-bouzid As per Anton's latest comment these dependencies aren't needed anymore but yep I think it was just out of date with the latest smoke tests! 

Agreed on the naming for the smoke tests, I've opened [another issue](https://github.com/microsoft/hi-ml/issues/719) so this can be dealt with in a separate PR.

ant0nsc (approved):

.github/workflows/hi-ml-pr.yml
```diff
@@ -134,7 +134,7 @@ jobs:
 
   build-python:
```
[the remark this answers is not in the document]
kenza-bouzid: pyright

kenza-bouzid (approved):

PR #599, opened 2022-09-12
BUG: Fix top bottom tiles selection mismatch.

kenza-bouzid:
For small validation sets and poor performance models, we sometimes end up with some slides falsely belonging to TP/FN cases. 
We need to make sure thay top_heaps contain only true cases and bottom ones, only false cases.
Check predicted vs true label before pushing into slides_heaps to make sure we only push true predictions in top_slides_heaps and false ones in bottom_slides_heaps

### DIFF
--- a/hi-ml-cpath/src/health_cpath/utils/tiles_selection_utils.py
+++ b/hi-ml-cpath/src/health_cpath/utils/tiles_selection_utils.py
@@ -180,30 +180,33 @@ class TilesSelector:
                 pred_prob_score = results[ResultsKey.CLASS_PROBS][:, pred_label][i].item()
                 tiles = batch[SlideKey.IMAGE][i]
                 attn_scores = results[ResultsKey.BAG_ATTN][i].squeeze(0)
-                self._update_label_slides(
-                    class_slides_heap=self.top_slides_heaps[gt_label],
-                    tiles=tiles,
-                    attn_scores=attn_scores,
-                    slide_node=SlideNode(
-                        slide_id=slide_ids[i],
-                        gt_prob_score=gt_prob_score,
-                        pred_prob_score=pred_prob_score,
-                        true_label=gt_label,
-                        pred_label=pred_label,
-                    ),
-                )
-                self._update_label_slides(
-                    class_slides_heap=self.bottom_slides_heaps[gt_label],
-                    tiles=tiles,
-                    attn_scores=attn_scores,
-                    slide_node=SlideNode(
-                        slide_id=slide_ids[i],
-                        gt_prob_score=-gt_prob_score,  # negative score for bottom slides to reverse order in max heap
-                        pred_prob_score=pred_prob_score,
-                        true_label=gt_label,
-                        pred_label=pred_label,
-                    ),
-                )
+                if pred_label == gt_label:
+                    self._update_label_slides(
+                        class_slides_heap=self.top_slides_heaps[gt_label],
+                        tiles=tiles,
+                        attn_scores=attn_scores,
+                        slide_node=SlideNode(
+                            slide_id=slide_ids[i],
+                            gt_prob_score=gt_prob_score,
+                            pred_prob_score=pred_prob_score,
+                            true_label=gt_label,
+                            pred_label=pred_label,
+                        ),
+                    )
+                elif pred_label != gt_label:
+                    self._update_label_slides(
+                        class_slides_heap=self.bottom_slides_heaps[gt_label],
+                        tiles=tiles,
+                        attn_scores=attn_scores,
+                        slide_node=SlideNode(
+                            slide_id=slide_ids[i],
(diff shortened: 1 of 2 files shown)


hi-ml-cpath/src/health_cpath/utils/tiles_selection_utils.py
```diff
@@ -110,23 +110,10 @@ def __init__(self, n_classes: int, num_slides: int = 10, num_tiles: int = 12) ->
         self.num_tiles = num_tiles
         self.top_slides_heaps: SlideDict = self._initialise_slide_heaps()
         self.bottom_slides_heaps: SlideDict = self._initialise_slide_heaps()
-        self.report_cases_slide_ids = self.init_report_cases()
 
     def _initialise_slide_heaps(self) -> SlideDict:
         return {class_id: [] for class_id in range(self.n_classes)}
 
-    def init_report_cases(self) -> Dict[str, List[str]]:
```
kenza-bouzid, line 118: This is deprecated and not used anywhere

ant0nsc (approved):

hi-ml-cpath/testhisto/testhisto/utils/test_tiles_selection_utils.py
```diff
@@ -42,12 +42,14 @@ def _create_mock_results(n_samples: int, n_tiles: int = 3, n_classes: int = 2, d
     :return: A dictioanry containing randomly generated mock results.
     """
     diff_n_tiles = [n_tiles + i for i in range(n_samples)]
+    probs = torch.rand((n_samples, n_classes), device=device)
+    probs = probs / probs.sum(dim=1, keepdim=True)
```
vale-salvatelli, line 46: for my own understanding, why this normalization is now needed?

hi-ml-cpath/testhisto/testhisto/utils/test_tiles_selection_utils.py
```diff
@@ -167,13 +160,21 @@ def test_aggregate_shallow_slide_nodes(n_classes: int, rank: int = 0, world_size
...
+            )
+            selected_bottom_slides_ids = [slide_node.slide_id for slide_node in shallow_bottom_slides_heaps[label]]
             expected_bottom_slides_ids = _get_expected_slides_by_probability(results, num_top_slides, label, top=False)
-            assert expected_bottom_slides_ids == [
-                slide_node.slide_id for slide_node in shallow_bottom_slides_heaps[label]
-            ]
+            assert expected_bottom_slides_ids == selected_bottom_slides_ids
+
+            # Make sure that the top and bottom slides are disjoint.
+            assert not set(selected_top_slides_ids).intersection(set(selected_bottom_slides_ids))
```
vale-salvatelli, line 177: I don't think the current test reproduces the error you fixed with this PR, because the mock data distribution is as before. Do you think we could add an edge case test that emulates the situation small validation set (with respect to the number of slides asked) and poor performance to avoid getting again similar bugs in the future? for example an idea could be to:

-  parametrize the tests to have num_top_slides=90% of tot_slides, 
- parametrize the test to have an edge case where the predicted probability is always around 0.5

Also are there cases where we should raise errors? e.g. number of required top tiles >= total number of slides. Can we add assertions for these cases?

vale-salvatelli (changes requested):
The fix looks good, but I'd like to make sure the test suite will avoid us bugs of this kind in the future

hi-ml-cpath/testhisto/testhisto/utils/test_tiles_selection_utils.py
kenza-bouzid: also note that we had a bug as True labels were sampled ` torch.randint(2, size=(n_samples,), device=device),` with 2 as upper bound so we never really tested the multi class scenario.

hi-ml-cpath/testhisto/testhisto/utils/test_tiles_selection_utils.py
kenza-bouzid: Very good point, we can parametrize the num_top_slides as you suggested. 

About the edge case: the tests already cover it implicitly we sample 20 random slides some of which have probabilities around 0.5 and lower with n_classes>2. I had to fix one edge case where the probabilities were [0.3, 0.3, 0.4] and 0.4 were among lower probabibilities in the batch but corresponded to the predicted label so it shouldn't be selected in bottom slides.
this is why I added the `_selection_condition` to the expected slides to make sure slides with highest/lowest probs correspond to predicted labels.

vale-salvatelli: I see you parametrized the number of slides - good enough, let's get this merged

vale-salvatelli (approved):

Issue #604, opened 2022-09-14
Add support to compute embeddings and similarities


fepegar:
These functionalities are currently not trivial/well exposed for a new user. The following need to be modular and it needs to be crystal clear how to:

- Compute embeddings for an image
- Compute embeddings for a string
- Compute a similarity matrix for a number of embeddings

@ozan-oktay @Shruthi42

ozan-oktay:
Closing the issue since it's addressed with #606

PR #606, opened 2022-09-14
ENH: Enable Global Similarity Comparison Between Text and Image Inputs

ozan-oktay:
#604 
Proposed changes:
- Image inference engine returns global image embeddings if user needs to compute image and sentence level alignment.
- Similarly, VLP engine can use this method to return a single cosine similarity score as in CLIP method.


hi-ml-multimodal/src/health_multimodal/image/inference_engine.py
```diff
@@ -56,7 +57,7 @@ def load_and_transform_input_image(self, image_path: Path, transform: Callable)
 
     @torch.no_grad()
     def get_patch_embeddings_from_image(self, image_path: Path) -> Tuple[torch.Tensor, TypeShape2D]:
```
Shruthi42, line 59: It would be nice to keep the two names similar (`get_projected_patch_embeddings...`) if the change is minimal.

Shruthi42 (approved):

fepegar (commented):
Thanks, @ozan-oktay. It would be best to make the methods slightly more modular: one function to load, one function to transform, one function to compute embeddings. This will maximise flexibility and simplify testing. I'm happy to work on that if you like.

ozan-oktay:
> Thanks, @ozan-oktay. It would be best to make the methods slightly more modular: one function to load, one function to transform, one function to compute embeddings. This will maximise flexibility and simplify testing. I'm happy to work on that if you like.

The PR is not introducing any new loading or transformation functions -- It's just using the existing class methods to return a global embedding similar to patch embedding computation, and adds a separate method in VLP inference engine to return a similarity score. Please feel free to create a separate issue for the idea you have in mind, and if there is something particularly applicable within this context, please let me know. Thanks.

hi-ml-multimodal/src/health_multimodal/image/inference_engine.py
```diff
@@ -67,3 +68,19 @@ def get_patch_embeddings_from_image(self, image_path: Path) -> Tuple[torch.Tenso
         assert projected_img_emb.shape[0] == 1
 
         return projected_img_emb[0], img_shape
+
+    @torch.no_grad()
+    def get_projected_global_embedding_from_image(self, image_path: Path) -> torch.Tensor:
+        """Compute global image embedding in the joint latent space.
+
+        :param image_path: Path to the image to compute embeddings for.
+        :return: Torch tensor containing l2-normalised global image embedding [joint_feature_dim,]
```
fepegar, line 77: Maybe it can be made a bit more explicit what `[joint_feature_dim,]` means here.

hi-ml-multimodal/src/health_multimodal/image/inference_engine.py
```diff
@@ -67,3 +68,19 @@ def get_patch_embeddings_from_image(self, image_path: Path) -> Tuple[torch.Tenso
         assert projected_img_emb.shape[0] == 1
 
         return projected_img_emb[0], img_shape
+
+    @torch.no_grad()
+    def get_projected_global_embedding_from_image(self, image_path: Path) -> torch.Tensor:
```
fepegar, line 73: I would remove `_from_image` here, as it can be inferred from the argument name and the docstring. Although I know it's common in our codebase, so feel free to ignore this comment.

hi-ml-multimodal/src/health_multimodal/vlp/inference_engine.py
```diff
@@ -27,6 +27,24 @@ def __init__(self,
         self.image_inference_engine = image_inference_engine
         self.text_inference_engine = text_inference_engine
 
+    def get_similarity_score_from_raw_data(self,
```
fepegar, line 30: I'm not sure what we mean by `from_raw_data` or why it's needed, but I guess it's ok if you want to keep it consistent with the other method.

hi-ml-multimodal/src/health_multimodal/vlp/inference_engine.py
```diff
@@ -27,6 +27,24 @@ def __init__(self,
...
+                                           image_path: Path,
+                                           query_text: str) -> float:
```
fepegar, line 32: I suggest supporting a sequence of strings as well, or even enforcing it. I think `get_embeddings_from_prompt` already supports that, so I think not much more needs to be changed here, except the call to `item`.

hi-ml-multimodal/test_multimodal/vlp/test_vlp_inference_engine.py
```diff
@@ -66,3 +66,8 @@ def test_vlp_inference(height: int, query_text: str) -> None:
 
         similarity_map = img_txt_inference._get_similarity_map_from_embeddings(image_embedding, text_embedding)
         assert similarity_map.shape == expected_image_embedding_size
+
+        # Test global similarity score
+        sim_score = img_txt_inference.get_similarity_score_from_raw_data(image_path=image_path, query_text=query_text)
+        assert isinstance(sim_score, float)
+        assert sim_score >= -1 and sim_score <= 1
```
fepegar, line 73:
```suggestion
        assert -1 <= sim_score <= 1
```

fepegar (changes requested):

hi-ml-multimodal/src/health_multimodal/vlp/inference_engine.py
ozan-oktay: Updated.

fepegar: We need a test with multiple prompts, I think you need to change the call to item etc.

hi-ml-multimodal/src/health_multimodal/image/inference_engine.py
ozan-oktay: updated.

hi-ml-multimodal/src/health_multimodal/vlp/inference_engine.py
ozan-oktay: It's already added.

fepegar (approved):

hi-ml-multimodal/src/health_multimodal/image/inference_engine.py
ozan-oktay: updated.

PR #622, opened 2022-10-05
BUG: Fix double root folder for heatmaps

kenza-bouzid:
Add kwargs to avoid confusion and remove double heatmap_folder


dccastro (approved):
Thanks for fixing! Another optional suggestion to further avoid such confusion is to have a dedicated method that just formats the filename, which then gets appended to the root folder when needed.

PR #627, opened 2022-10-06
ENH: Incorporate AML SDK v2

mebristo:
Incorporate the use of AML SDK v2 into the codebase alongisde the existing SDK v1. The plan would be to eventually replace all v1 functionality with v2 commands, which will entail a lot of refactoring, and there may still be certain v1 functions that we currently use that there is no v2 equivalent for, so an incremental approach will be necessary.


hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -341,9 +356,9 @@ def test_create_run_configuration(
     mock_env_name = "Mock Env"
     mock_environment_get.return_value = mock_env_name
     mock_workspace.compute_targets = {existing_compute_target: mock_compute_cluster}
-    aml_input_dataset = MagicMock()
+    aml_input_dataset = create_autospec(DatasetConsumptionConfig)
```
ant0nsc, line 359: Learned something new, `create_autospec` is handy!

hi-ml-cpath/environment.yml
```diff
@@ -84,11 +84,14 @@ dependencies:
     - azure-mgmt-keyvault==10.0.0
     - azure-mgmt-resource==21.1.0
     - azure-mgmt-storage==20.0.0
+    - azure-storage-blob==12.10.0
```
ant0nsc, line 87: Rather than modifying this directly, can you create a fresh locked environment by chaning `primary_deps.yml` and running the lock script? See comment at the start of the file.

hi-ml-multimodal/requirements_build.txt
```diff
@@ -1,3 +1,4 @@
+colorama==0.4.4
```
ant0nsc, line 1: what's this about?

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -297,6 +297,8 @@ def get_different_seeds_hyperdrive_config(self) -> HyperDriveConfig:
 
 
 class DatasetParams(param.Parameterized):
+    input_datasets: List[str] = param.List(default=[], class_=str, doc="Do not use. For AML SDK v2 only")
```
ant0nsc, line 300: this attribute must have the same name as literal strings that you use elsewhere, is there a test for that?

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -297,6 +297,8 @@ def get_different_seeds_hyperdrive_config(self) -> HyperDriveConfig:
...
+    datastore_name: str = param.String(default="", doc="Datastore to look for data in")
```
ant0nsc, line 301: this might be used on the commandline, considering shortening to `datastore`?

ant0nsc: having said that, the more generic sthing would be to to extend azuredatasets to allow datasets to look like `datastore:datasetname` (after all, they might live on differnt datastores)

hi-ml/src/health_ml/utils/logging.py
```diff
@@ -333,6 +317,107 @@ def _print(self, message: str) -> None:
             sys.stdout.flush()
 
 
+class MLFlowLogger(LightningLoggerBase):
+    HYPERPARAMS_NAME = "hyperparams"
+
+    def _start_run(self, mlflow_run_id: Optional[str] = None) -> None:
```
ant0nsc, line 323: this class would fit better in `lightning_loggers.py`

hi-ml/src/health_ml/utils/logging.py
```diff
@@ -333,6 +317,107 @@ def _print(self, message: str) -> None:
             sys.stdout.flush()
 
 
+class MLFlowLogger(LightningLoggerBase):
```
ant0nsc, line 320: PL has a logger alrady: https://pytorch-lightning.readthedocs.io/en/stable/extensions/generated/pytorch_lightning.loggers.MLFlowLogger.html

hi-ml-azure/run_requirements.txt
```diff
@@ -1,12 +1,16 @@
...
 conda-merge>=0.1.5
+mlflow>=1.29.0
 pandas>=1.3.4
 param>=1.12
 protobuf<4.0
 pysocks>=1.5.8
 ruamel.yaml>=0.16.12
 tensorboard>=2.6.0
-typing-extensions>=4.3.0
+typing-extensions>=4.3.0#
```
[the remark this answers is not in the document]
mebristo: Whoops! Fixed

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -138,10 +288,12 @@ def to_input_dataset_local(self, workspace: Optional[Workspace]) -> Tuple[Path,
         if workspace is None:
             raise ValueError(f"Unable to make dataset '{self.name} available for a local run because no AzureML "
                              "workspace has been provided. Provide a workspace, or set a folder for local execution.")
-
-        azureml_dataset = get_or_create_dataset(workspace=workspace,
-                                                dataset_name=self.name,
-                                                datastore_name=self.datastore)
+        azureml_dataset: FileDataset = get_or_create_dataset(workspace=workspace,
```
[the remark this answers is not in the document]
mebristo: Ah this was an attempt to stop pyright from complaining,  have removed it

hi-ml-multimodal/requirements_build.txt
Shruthi42: I have the same question - to add some context, hi-ml-multimodal does not use any of the other packages in this repo, so it should be unaffected by this PR.

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -381,10 +473,44 @@ def _str_to_path(s: Optional[PathOrString]) -> Optional[Path]:
     return s
 
 
+def create_v2_inputs(workspace_client: MLClient, input_datasets: List[DatasetConfig]) -> Dict[str, Input]:
+    inputs: Dict[str, Input] = {}
+    for input_dataset in input_datasets:
+        version = input_dataset.version or 1
+        data_asset: Data = workspace_client.data.get(input_dataset.name, version=str(version))
```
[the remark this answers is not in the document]
mebristo: Here input_dataset is an insance of DatasetConfig which specifies version to be an int, but the method mlclient.data.get expects a string

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -399,6 +525,7 @@ def submit_to_azure_if_needed(  # type: ignore
         ignored_folders: Optional[List[PathOrString]] = None,
         default_datastore: str = "",
         input_datasets: Optional[List[StrOrDatasetConfig]] = None,
+        v2_input_datasets: Optional[Dict[str, Input]] = None,
```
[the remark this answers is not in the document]
mebristo: Good catch, this arg should have been deleted. We define this within the function now

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -558,14 +691,27 @@ def submit_to_azure_if_needed(  # type: ignore
...
+            run = submit_run_v2(workspace=workspace,
+                                v2_input_datasets=v2_input_datasets,
+                                v2_output_datasets=v2_output_datasets,
+                                experiment_name=effective_experiment_name,
+                                script_run_config=config_to_submit,
+                                tags=tags,
+                                wait_for_completion=wait_for_completion,
+                                wait_for_completion_show_output=wait_for_completion_show_output)
+
+    if after_submission is not None and isinstance(run, Run):
```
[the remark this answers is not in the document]
mebristo: This will require the new functionality for interacting with Jobs rather than Runs so will be in the future PR

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -2022,3 +2033,123 @@ def check_is_any_of(message: str, actual: Optional[str], valid: Iterable[Optiona
...
+            client_id=service_principal_id,
+            client_secret=service_principal_password)
+        cs_cred.get_token("https://management.azure.com/.default")
+        return cs_cred
+    else:
+        if not is_running_in_azure_ml() or is_running_on_azure_agent():
+            def_cred = DefaultAzureCredential()
+            def_cred.get_token("https://management.azure.com/.default")
+            return def_cred
+        if not is_running_in_azure_ml() or is_running_on_azure_agent():
```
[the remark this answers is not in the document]
mebristo: Whoops, good catch

hi-ml-cpath/environment.yml
mebristo: This was a workaround as I kept getting OutOfMemory errors when running the script but have since changed $TMP_DIR so should be able to try this again

hi-ml-azure/run_pytest.py
```diff
@@ -68,6 +68,8 @@ class RunPytestConfig(param.Parameterized):
         default="",
         doc="A folder name that should be added to sys.path. The folder name should be relative to repository root."
     )
+    strictly_aml_v1: bool = param.Boolean(default=False, doc="If True, use AzureML v1 SDK. If False (default), use "
```
kenza-bouzid, line 71: [Minor] Seems like it makes more sense to have default as True from the rest of the code.

kenza-bouzid (changes requested):
Looking good, I still have to go through the tests but here's a first round of comments.

hi-ml-azure/src/health_azure/datasets.py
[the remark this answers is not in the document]
mebristo: No that was also a leftover part of the attempt to silence pyright (it always passed because we don't currently have any v2 data assets)

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -167,14 +322,17 @@ def to_input_dataset(self,
         :param dataset_index: Suffix for using datasets as named inputs, the dataset will be marked INPUT_{index}
         """
         status = f"In AzureML, dataset {self.name} (index {dataset_index}) will be "
-        azureml_dataset = get_or_create_dataset(workspace=workspace,
-                                                dataset_name=self.name,
-                                                datastore_name=self.datastore)
-        named_input = azureml_dataset.as_named_input(_input_dataset_key(index=dataset_index))
+        azureml_dataset: FileDataset = get_or_create_dataset(workspace=workspace,
```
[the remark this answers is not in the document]
mebristo: Same as above

hi-ml-azure/src/health_azure/himl.py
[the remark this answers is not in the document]
mebristo: I'm hesitant to change the format of the function names because it can be a pain when you're searching for a function if there's another with a matching name + suffix...

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -460,6 +476,7 @@ def test_create_run_configuration_correct_env(mock_create_environment: MagicMock
             with pytest.raises(Exception) as e:
                 himl.create_run_configuration(mock_workspace,
                                               "dummy_compute_cluster",
+                                              True,
```
kenza-bouzid, line 479: can you please add the kwargs here? I presume it's strictly_aml_v1?

hi-ml/src/health_ml/deep_learning_config.py
mebristo: I think searching through all datastores may be outside the scope of this PR. For now I have shortened the arg to datastore

hi-ml/src/health_ml/utils/logging.py
```diff
@@ -333,6 +317,107 @@ def _print(self, message: str) -> None:
...
+                    new_key = key + f"_{num_related_keys}"
+                    new_params[new_key] = val
+                else:
+                    new_params[key] = val
+            params_final = new_params
+
+        if len(params_final) > 0:
+            mlflow.log_params(params_final)
+
+    def experiment(self) -> Optional[str]:
```
kenza-bouzid, line 388: [Minor] make experiment, name and version properties, for easy access?

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -350,22 +348,7 @@ def test_log_on_vm(log_from_vm: bool) -> None:
     assert runner.trainer.loggers is not None
     assert len(runner.trainer.loggers) > 1
     logger = runner.trainer.loggers[1]
-    assert isinstance(logger, AzureMLLogger)
-    if log_from_vm:
```
kenza-bouzid, line 354: what happened to all these checks?

hi-ml/src/health_ml/utils/logging.py
mebristo: It does but it doesn't set the experiment or start/ retrieve a run for us, and it has a different signature to what we want

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -118,7 +264,11 @@ def __init__(self,
             raise ValueError("Can't mount or download a dataset to the current working directory.")
         self.local_folder = Path(local_folder) if local_folder else None
 
-    def to_input_dataset_local(self, workspace: Optional[Workspace]) -> Tuple[Path, Optional[MountContext]]:
+    def to_input_dataset_local(self,
+                               strictly_aml_v1: bool,
+                               workspace: Workspace = None,
+                               workspace_client: Optional[MLClient] = None,
```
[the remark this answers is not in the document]
mebristo: Hopefully the renaming should have made this clearer - the Workspace object and MLClient object are different things. The AML SDK (v2) has it's own definition of a Workspace class (in `azure.ai.ml.entities`)

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -157,8 +309,11 @@ def to_input_dataset_local(self, workspace: Optional[Workspace]) -> Tuple[Path,
         return result
 
     def to_input_dataset(self,
+                         dataset_index: int,
                          workspace: Workspace,
-                         dataset_index: int) -> DatasetConsumptionConfig:
+                         strictly_aml_v1: bool,
+                         workspace_client: Optional[MLClient] = None,
```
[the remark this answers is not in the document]
mebristo: As mentioned above, they have different purposes

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -24,45 +30,185 @@ def get_datastore(workspace: Workspace, datastore_name: str) -> Datastore:
...
+    # trouble (for example, folders foo and foo_bar exist, and I'm trying to create a dataset from "foo")
+    azureml_dataset = Dataset.File.from_files(path=(datastore, dataset_name + "/"))
+    logging.info("Registering the dataset for future use.")
+    azureml_dataset.register(workspace, name=dataset_name)
+    return azureml_dataset
+
+
+def _get_or_create_v1_dataset(datastore_name: str, dataset_name: str, workspace: Workspace) -> Dataset:
+    """
+    Attempt to retrieve a v1 Dataset object and return that, otherwise atempts to create and register
```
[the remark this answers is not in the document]
mebristo: Good catch!

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -24,45 +30,185 @@ def get_datastore(workspace: Workspace, datastore_name: str) -> Datastore:
...
+    a v1 Dataset and return that.
+
+    :param datastore_name: The name of the Datastore to either retrieve or create and register the Dataset in.
+    :param dataset_name: The name of the Dataset to be retrieved or registered.
+    :param workspace: An Azure ML Workspace object.
+    :return: An Azure ML Dataset object with the provided dataset name, in the provided datastore.
+    """
+    try:
+        azureml_dataset = _retrieve_v1_dataset(dataset_name, workspace)
+        assert azureml_dataset is not None
```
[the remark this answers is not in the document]
mebristo: This is just for mypy, because the return type of  `_retrieve_v1_dataset` is Optional, but the return type of `_get_or_create_v1_dataset` should be a dataset (because if one can't be found it should be created). Note that we don't have the same issue for the v2 version of the function because there if the data can't be found it raises an Exception rather than returning None

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -24,45 +30,185 @@ def get_datastore(workspace: Workspace, datastore_name: str) -> Datastore:
...
+    except Exception:
+        azureml_dataset = _create_v1_dataset(datastore_name, dataset_name, workspace)
+    return azureml_dataset
+
+
+def _retrieve_v2_dataset(dataset_name: str, ml_client: MLClient) -> Data:
+    """
+    Attempt to retrieve a v2 Data Asset using a provided Azure ML Workspace connection. If
```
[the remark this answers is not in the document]
mebristo: Yes exactly

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -24,45 +30,185 @@ def get_datastore(workspace: Workspace, datastore_name: str) -> Datastore:
...
+    """
+    Create or update a v2 Data asset in the specified Datastore
+
+    :param datastore_name: The name of the datastore in which to create or update the Data asset.
+    :param dataset_name: The name of the dataset to be created.
+    :param ml_client: An Azure MLClient object for interacting with Azure resources.
+    :raises ValueError: If no datastore name is provided to define where to create the data
+    :return: The created or updated Data asset.
+    """
+    if not datastore_name:
```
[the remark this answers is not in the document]
mebristo: Correct. And yes good point we should check the dataset name too

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -24,45 +30,185 @@ def get_datastore(workspace: Workspace, datastore_name: str) -> Datastore:
...
+
+    :param datastore_name: The name of hte datastore in which to look for, or create and register, the dataset.
+    :param dataset_name: The name of the dataset to find or create.
+    :param workspace: An AML Workspace object for interacting with AML v1 datastores.
+    :param strictly_aml_v1: If True, use Azure ML SDK v1 to attempt to find or create and reigster the dataset.
+        Otherwise, attempt to use Azure ML SDK v2.
+    :param ml_client: An MLClient object for interacting with AML v2 datastores.
     """
     if not dataset_name:
         raise ValueError("No dataset name provided.")
```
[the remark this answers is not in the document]
mebristo: I'll move the datastore_name check here so that they are together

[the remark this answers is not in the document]
mebristo: Ahh no actually I can't because it's not a problem at this point if datastore_name is empty - we can still take the default

hi-ml-azure/src/health_azure/datasets.py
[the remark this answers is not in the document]
kenza-bouzid: Can we then check if it's None and raise an exception? It seems like it can cause some issues otherwise. It is now caught by the assertion, but maybe worth raising an exception rather than an assertion Error if not found?

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -334,12 +494,11 @@ def setup_local_datasets(dataset_configs: List[DatasetConfig],
     :return: Pair of: list of optional paths to the input datasets, list of mountcontexts, one for each mounted dataset.
     """
     workspace = find_workspace_for_local_datasets(aml_workspace, workspace_config_path, dataset_configs)
-
     mounted_input_datasets: List[Optional[Path]] = []
     mount_contexts: List[MountContext] = []
 
     for d in dataset_configs:
```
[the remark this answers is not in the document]
mebristo: Sure

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -317,12 +329,91 @@ def create_script_run(snapshot_root_directory: Optional[Path] = None,
...
+                  tags: Optional[Dict[str, str]] = None,
+                  wait_for_completion: bool = False,
+                  wait_for_completion_show_output: bool = False,
+                  workspace_config_path: Optional[PathOrString] = None,
+                  ml_client: Optional[MLClient] = None) -> Job:
+    if ml_client is None:
+        if workspace is not None:
+            ml_client = get_ml_client(
+                subscription_id=workspace.subscription_id,
+                resource_group=workspace.
```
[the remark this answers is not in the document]
kenza-bouzid: Oh sorry, it's related to Anton's comment about keeping `strictly_aml_v1 `as last positional args, might be useful for `ml_client `as well

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -381,10 +472,44 @@ def _str_to_path(s: Optional[PathOrString]) -> Optional[Path]:
...
+            mode=InputOutputModes.MOUNT,
+        )
+    return inputs
+
+
+def create_v2_outputs(output_datasets: List[DatasetConfig]) -> Dict[str, Output]:
+    outputs = {}
+    for output_dataset in output_datasets:
+        v1_datastore_path = f"azureml://datastores/{output_dataset.datastore}/paths/{output_dataset.name}"
+        # Note that there are alternative formats that the input path can take, such as:
```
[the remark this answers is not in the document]
mebristo: Here's the link: https://learn.microsoft.com/en-us/azure/machine-learning/how-to-read-write-data-v2?tabs=cli#supported-paths 
I'm hesistant to include it in the codebase since then we will have to maintain it (check the link still exists). Is there another way I can improve the description?

hi-ml-cpath/src/health_cpath/datamodules/base_module.py
```diff
@@ -98,13 +98,22 @@ def get_splits(self) -> Tuple[_SlidesOrTilesDataset, _SlidesOrTilesDataset, _Sli
         raise NotImplementedError
 
     def train_dataloader(self) -> DataLoader:
-        return self._get_dataloader(self.train_dataset, shuffle=True, stage=ModelKey.TRAIN, **self.dataloader_kwargs)
+        return self._get_dataloader(self.train_dataset,  # type: ignore
```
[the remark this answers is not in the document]
mebristo: ` "HistoDataModule[SlidesDataset]" has no attribute "_get_dataloader"; maybe "test_dataloader", "train_dataloader", or "val_dataloader"? `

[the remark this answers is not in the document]
mebristo: I don't know why this error has just started appearing now 😕

hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -260,15 +260,15 @@ def _shared_step(self, batch: Dict, batch_idx: int, stage: str) -> BatchResultsT
 
     def training_step(self, batch: Dict, batch_idx: int) -> Tensor:  # type: ignore
         train_result = self._shared_step(batch, batch_idx, ModelKey.TRAIN)
-        self.log('train/loss', train_result[ResultsKey.LOSS], on_epoch=True, on_step=True, logger=True,
+        self.log('train/loss', train_result[ResultsKey.LOSS], on_epoch=True, on_step=False, logger=True,
```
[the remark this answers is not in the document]
mebristo: Whoops I shouldn't have included this - was seeing if it made a difference to the fact we see 2 steps per epoch!

hi-ml-cpath/testSSL/testSSL/test_ssl_containers.py
```diff
@@ -109,7 +109,7 @@ def _compare_stored_metrics(runner: Runner, expected_metrics: Dict[str, float],
 
 
 @pytest.mark.flaky(reruns=3)
-def test_ssl_container_cifar10_resnet_simclr() -> None:
+def test_ssl_container_cifar10_resnet_simclr(tmp_path: Path) -> None:
```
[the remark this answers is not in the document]
mebristo: Good catch - removed

hi-ml/src/health_ml/utils/logging.py
```diff
@@ -333,6 +317,107 @@ def _print(self, message: str) -> None:
...
+    @rank_zero_only
+    def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) -> None:
+        is_epoch_metric = "epoch" in metrics
+        for key, value in metrics.items():
+            # Log all epoch-level metrics without the step information
+            # All step-level metrics with step
+            mlflow.log_metric(key, value, step=None if is_epoch_metric else step)
+
+    @rank_zero_only
+    def log_hyperparams(self, params: Union[argparse.Namespace, Dict[str, Any]]) -> None:
```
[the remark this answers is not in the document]
mebristo: I copied this from AzureMLLogger so I don't know the exact use case in mind here. Logically I don't think you would call `log_hyperparams` with params = None but maybe there is an edge case where it can occur and break logging?

hi-ml/testhiml/testhiml/test_run_ml.py
mebristo: There is no more native AzureMLLogger, and we don't have any of these properties in the new MLFlow logger so this functionality will need to be recreated for the new logger (in a separate PR)

hi-ml/src/health_ml/deep_learning_config.py
mebristo: It's the same name as this command line arg that gets passed into the Azure ML (v2) command: 
` --input_datasets=${{inputs.input_datasets}}`

There isn't currently an explicit test for this - I guess what we'd want to do is submit an AML v2 command (passing this arg in) and then within the Job, check that the LightningContainer has the expected value of input_datasets? (and the same as above for output_datasets too).

However we do implicitly test this via the smoke tests since they now use the v2 sdk and they would fail to find the data if input_datasets wasn't set...So maybe we can leave the explicit tests for a future PR?

hi-ml-azure/src/health_azure/datasets.py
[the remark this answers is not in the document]
mebristo: Just realised this is wrong, it shouldn't be wrapped in a `try/except`. If `_retrieve_v1_dataset` returns None, then we should then call `_create_v1_dataset`. Have updated it now

hi-ml-azure/src/health_azure/himl.py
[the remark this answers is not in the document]
mebristo: Ahh I see. Have a look at my comment above - I'm not sure if ml_client should be a the end or rather near the workspace argument

hi-ml-azure/run_pytest.py
mebristo: I think we should be using the SDK v2 now where possible. Are there specific places where it contradicts this? (If you're judging by the number of tests where I've set `strictly_aml_v1=True` it's because I didn't want to change the behaviour of the tests)

kenza-bouzid: Yes, exactly! I have a comment somewhere in the tests that says that I finally got why you want to set it as False by default.

hi-ml-azure/src/health_azure/datasets.py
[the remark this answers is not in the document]
mebristo: Sorry  was wrong above - it does throw an exception. So I have added the try/catch back in

hi-ml/src/health_ml/utils/logging.py
ant0nsc: Can't we then at least use the bits that match (inheritance) and modify?

ant0nsc: ... saying all this without having had a very close look at the differences.

ant0nsc (approved):

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -2035,121 +2034,183 @@ def check_is_any_of(message: str, actual: Optional[str], valid: Iterable[Optiona
...
+    """
+    Validate credential by attempting to get token. If authentication has been successful, get_token
+    will succeed. Otherwise an exception will be raised
+
+    :param credential: The credential object to validate.
+    """
+    credential.get_token("https://management.azure.com/.default")
+
+
+def _get_legitimate_service_principal_credential(tenant_id: str, service_principal_id: str,
```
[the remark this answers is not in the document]
mebristo: As opposed to a credential that would fail the validation stage

hi-ml/src/health_ml/utils/logging.py
mebristo: It doesn't have any that I can see

mebristo: I can look into this more in a future PR

hi-ml-cpath/src/SSL/lightning_modules/ssl_online_evaluator.py
```diff
@@ -97,7 +97,7 @@ def on_pretrain_routine_start(self, trainer: pl.Trainer, pl_module: pl.Lightning
         self.evaluator.to(pl_module.device)
         if hasattr(trainer, "accelerator_connector"):
             # This works with Lightning 1.3.8
```
kenza-bouzid, line 99: [Out of scope] Do you know why we're still testing older versions of PL?

kenza-bouzid (approved):
LGTM! (happy to look into the v2 tests once you add them)

ant0nsc (approved):

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -381,11 +483,59 @@ def _str_to_path(s: Optional[PathOrString]) -> Optional[Path]:
...
+            type=AssetTypes.URI_FOLDER,
+            path=data_path,
+            mode=InputOutputModes.MOUNT,
+        )
+    return inputs
+
+
+def create_v2_outputs(output_datasets: List[DatasetConfig]) -> Dict[str, Output]:
+    """
+    Create a dictionary of Azure ML v2 Output objects, required for passing output data in to an AML job
```
kenza-bouzid, line 514:
```suggestion
    Create a dictionary of Azure ML v2 Output objects, required for passing output data into an AML job
```

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -584,26 +766,35 @@ def _write_run_recovery_file(run: Run) -> None:
...
+    workspace: Workspace,
+    strictly_aml_v1: bool
+) -> Tuple[Dict[str, DatasetConsumptionConfig], Dict[str, OutputFileDatasetConfig]]:
     """
     Convert the cleaned input and output datasets into dictionaries of DatasetConsumptionConfigs for use in AzureML.
 
     :param cleaned_input_datasets: The list of input DatasetConfigs
     :param cleaned_output_datasets: The list of output DatasetConfigs
     :param workspace: The AzureML workspace
+    :param ml_client: An optional Azure MLClient object for interacting with Azure resources.
```
kenza-bouzid, line 780: wrong param? strictly_aml_v1 needs to be added instead

hi-ml-azure/src/health_azure/mlflow_utils.py
```diff
@@ -0,0 +1,48 @@
...
+#  ------------------------------------------------------------------------------------------
+"""
+Utility functions for interacting with mlflow runs
+"""
+from typing import Any, Dict, List
+from mlflow.client import MlflowClient
+from mlflow.entities import Run, Metric
+
+
+def get_mlflow_run(mlflow_client: MlflowClient, mlflow_run_id: str) -> Run:
```
kenza-bouzid, line 13: do we need a function for this? can't we jus use mlflow_client.get_run(mlflow_run_id) directly?
same for get_metric_from_mlflow_run and get_last_metrics_from_mlflow_run 
++reduces codecov complains

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -2021,3 +2108,205 @@ def check_is_any_of(message: str, actual: Optional[str], valid: Iterable[Optiona
...
+
+def fetch_job(ml_client: MLClient, run_id: str) -> Job:
+    """
+    Retrive a job with a given run_id from an MLClient
+
+    :param ml_client: An MLClient object.
+    :param run_id: The id of the run to retrieve.
+    :return: An Azure ML (v2) Job object.
+    """
+    job = ml_client.jobs.get(run_id)
```
kenza-bouzid, line 2311: same comment, do we need an extra func for this?

kenza-bouzid (commented):
Looking good, only minor comments/suggestions to reduce codecov complains and improve test coverage. (I still have to check the tests)

hi-ml-azure/src/health_azure/himl.py
mebristo: Here we are passing the data to the job, so pass 'in to' is correct (https://idioms.thefreedictionary.com/pass+into)

hi-ml-azure/src/health_azure/himl.py
mebristo: True! Fixed

hi-ml-azure/src/health_azure/mlflow_utils.py
mebristo: These will go on to replace our functions for downloading runs from AML so yes, we do need these

hi-ml-azure/src/health_azure/utils.py
mebristo: I think it's important to have an example of how to do this, since it's not clearly explained in the Azure ML docu

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -297,6 +297,10 @@ def get_different_seeds_hyperdrive_config(self) -> HyperDriveConfig:
 
 
 class DatasetParams(param.Parameterized):
+    INPUT_0: List[str] = param.List(default=[], class_=str, doc="Do not use. For AML SDK v2 only")
```
kenza-bouzid, line 300: Didn't we say that we don't need these anymore?

hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -268,7 +268,7 @@ def training_step(self, batch: Dict, batch_idx: int) -> Tensor:  # type: ignore
 
     def validation_step(self, batch: Dict, batch_idx: int) -> BatchResultsType:  # type: ignore
         val_result = self._shared_step(batch, batch_idx, ModelKey.VAL)
-        self.log('val/loss', val_result[ResultsKey.LOSS], on_epoch=True, on_step=True, logger=True,
+        self.log('val/loss', val_result[ResultsKey.LOSS], on_epoch=False, on_step=True, logger=True,
```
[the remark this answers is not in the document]
mebristo: Whoops, I think I undid a previous commit. Fixed

hi-ml-cpath/testSSL/testSSL/test_encoders.py
```diff
@@ -1,36 +0,0 @@
-#  ------------------------------------------------------------------------------------------
-#  Copyright (c) Microsoft Corporation. All rights reserved.
-#  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
-#  ------------------------------------------------------------------------------------------
-
```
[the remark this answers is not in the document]
mebristo: Yes I renamed the file to test_ssl_encoders.py because there's already a file called test_encoders.py in hi-ml-cpath and pyright  was complaining

hi-ml/src/health_ml/deep_learning_config.py
mebristo: Good point - this has been removed and replaced with argument filtering at the start of the script

hi-ml/src/health_ml/model_trainer.py
```diff
@@ -91,9 +93,24 @@ def create_lightning_trainer(container: LightningContainer,
...
-    loggers = [tensorboard_logger, azureml_logger]
+    loggers: List[Any] = [tensorboard_logger]
+
+    if is_running_in_azure_ml():
+        mlflow_run_id = os.environ.get("MLFLOW_RUN_ID", None)
+        mlflow_logger = MlflowLogger(
+            run_id=mlflow_run_id
+        )
+    else:
+        mlflow_run_dir = container.outputs_folder / "mlruns"
```
ant0nsc, line 104: so we have lost the ability to log to AzureML from a VM? (OK to punt that to a next PR)

hi-ml/src/health_ml/utils/lightning_loggers.py
```diff
@@ -109,3 +112,33 @@ def to_metrics_dicts(self, prefix_filter: str = "") -> Dict[int, DictStrFloat]:
         :return: A dictionary mapping from epoch number to metric name to metric value.
         """
         return {epoch: self.extract_by_prefix(epoch, prefix_filter) for epoch in self.epochs}
+
+
+class MlflowLogger(MLFlowLogger):
```
ant0nsc, line 117: I find it confusing to have two classes that only differ by casing.

ant0nsc (approved):

hi-ml/src/health_ml/utils/lightning_loggers.py
mebristo: Yes I wasn't sure what to call this...any ideas? HimlMlflowLogger?

hi-ml/src/health_ml/model_trainer.py
mebristo: Yes I havent quite figured that out with mlfow.  It will be something like 
```
elif azureml_run_for_logging:
    tracking_uri = workspace.get_mlflow_tracking_uri()
    mlflow_logger = MlflowLogger(run_id=azureml_run_for_logging, tracking_uri=tracking_uri)
```
but that doesn't work so something is missing

mebristo: (Might need an `mlflow.start_run()`)

PR #629, opened 2022-10-11
ENH: Track validation loss and add entropy value

kenza-bouzid:
Add entropy values to detect ambiguity and track validation loss during training

### DIFF
--- a/hi-ml-cpath/src/health_cpath/models/deepmil.py
+++ b/hi-ml-cpath/src/health_cpath/models/deepmil.py
@@ -100,8 +100,7 @@ class BaseDeepMILModule(LightningModule):
         self.aggregation_fn, self.num_pooling = pooling_params.get_pooling_layer(self.encoder.num_encoding)
         self.classifier_fn = self.get_classifier()
         self.activation_fn = self.get_activation()
-        self.loss_fn = self.get_loss()
-        self.loss_fn_no_reduction = self.get_loss(reduction="none")
+        self.loss_fn = self.get_loss(reduction="none")
 
         # Metrics Objects
         self.train_metrics = self.get_metrics()
@@ -297,6 +296,7 @@ class BaseDeepMILModule(LightningModule):
 
     def _shared_step(self, batch: Dict, batch_idx: int, stage: str) -> BatchResultsType:
 
+        results = dict()
         bag_logits, bag_labels, bag_attn_list = self.compute_bag_labels_logits_and_attn_maps(batch)
 
         if self.n_classes > 1:
@@ -312,8 +312,9 @@ class BaseDeepMILModule(LightningModule):
             predicted_labels = round(predicted_probs).int()
             probs_perclass = Tensor([[1.0 - predicted_probs[i][0].item(), predicted_probs[i][0].item()]
                                      for i in range(len(predicted_probs))])
-
-        loss = loss.view(-1, 1)
+        if stage in [ModelKey.TRAIN, ModelKey.VAL]:
+            results[ResultsKey.LOSS_PER_SAMPLE] = loss.detach().cpu().numpy()
+        loss = loss.mean().view(-1, 1)
         predicted_labels = predicted_labels.view(-1, 1)
         batch_size = predicted_labels.shape[0]
 
@@ -322,7 +323,6 @@ class BaseDeepMILModule(LightningModule):
 
         bag_labels = bag_labels.view(-1, 1)
 
-        results = dict()
         for metric_object in self.get_metrics_dict(stage).values():
             metric_object.update(predicted_probs, bag_labels.view(batch_size,).int())
         results.update({ResultsKey.LOSS: loss,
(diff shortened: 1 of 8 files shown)


hi-ml-cpath/src/health_cpath/utils/callbacks.py
```diff
@@ -384,69 +425,119 @@ def plot_loss_heatmap_for_slides_of_epoch(
...
         if self.should_cache_loss_values(trainer.current_epoch):
-            bag_logits, bag_labels, _ = pl_module.compute_bag_labels_logits_and_attn_maps(batch)
-            if pl_module.n_classes > 1:
-                loss = pl_module.loss_fn_no_reduction(bag_logits, bag_labels.long())
-            else:
-                loss = pl_module.loss_fn_no_reduction(bag_logits.squeeze(1), bag_labels.float())
-            self.loss_cache[ResultsKey.LOSS].extend(loss.tolist())
-            self.loss_cache[ResultsKey.SLIDE_ID].extend([slides[0] for slides in batch[ResultsKey.SLIDE_ID]])
+            self.loss_cache[stage][ResultsKey.LOSS].extend(outputs[ResultsKey.LOSS_PER_SAMPLE])
+            self.loss_cache[stage][ResultsKey.SLIDE_ID].extend([slides[0] for slides in batch[ResultsKey.SLIDE_ID]])
```
dccastro, line 442: Why do the slide IDs need to be retrieved from the input `batch` instead of the `outputs`?

hi-ml-cpath/testhisto/testhisto/utils/test_callbacks.py
```diff
@@ -93,108 +108,130 @@ def test_loss_analysis_epochs_interval(epochs_interval: int) -> None:
...
-    callback.on_train_batch_start(trainer, container.model, batch, 0, None)  # type: ignore
-    _assert_loss_cache_contains_n_elements(callback.loss_cache, batch_size)
+    _assert_loss_cache_contains_n_elements(callback.loss_cache[ModelKey.TRAIN], 0)
+    _assert_loss_cache_contains_n_elements(callback.loss_cache[ModelKey.VAL], 0)
 
-    batch = next(iter(container.data_module.train_dataloader()))
-    callback.on_train_batch_start(trainer, container.model, batch, 1, None)  # type: ignore
-    _assert_loss_cache_contains_n_elements(callback.loss_cache, 2 * batch_size)
+    def _call_on_batch_end_hook(on_batch_end_hook: Callable, batch_idx: int) -> None:
+        batch = next(iter(container.data_module.train_dataloader()))
```
dccastro, line 129: [Minor] Doesn't this always return the same batch every time? If so, won't the cache have duplicates, and is it a problem at all?

dccastro (approved):
Great job as usual, @kenza-bouzid! Approved with some minor optional suggestions and questions.

hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -341,12 +351,16 @@ def _shared_step(self, batch: Dict, batch_idx: int, stage: str) -> BatchResultsT
...
 
-    def training_step(self, batch: Dict, batch_idx: int) -> Tensor:  # type: ignore
+    def training_step(self, batch: Dict, batch_idx: int) -> Union[Tensor, BatchResultsType]:  # type: ignore
         train_result = self._shared_step(batch, batch_idx, ModelKey.TRAIN)
         self.log('train/loss', train_result[ResultsKey.LOSS], on_epoch=True, on_step=True, logger=True,
                  sync_dist=True)
         if self.verbose:
             print(f"After loading images batch {batch_idx} -", _format_cuda_memory_stats())
+        if self.analyse_loss:
+            return {ResultsKey.LOSS: train_result[ResultsKey.LOSS],
```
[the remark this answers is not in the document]
kenza-bouzid: Good point, I wanted to make as little changes as possible to the original behaviour, but this shouldn't hurt any performance, will fix the corresponding tests as well.

hi-ml-cpath/src/health_cpath/utils/callbacks.py
```diff
@@ -163,38 +168,41 @@ def _get_filename(self, filename: str, epoch: int, order: Optional[str] = None)
...
 
-    def get_scatter_plot_file(self, order: str) -> Path:
-        return self.scatter_folder / "slides_with_{}_loss_values.png".format(order)
+    def get_scatter_plot_file(self, order: str, stage: ModelKey) -> Path:
+        return self.get_scatter_folder(stage) / f"slides_with_{order}_loss_values_{stage}.png"
 
-    def get_heatmap_plot_file(self, epoch: int, order: str) -> Path:
-        return self.heatmap_folder / self._get_filename(filename="epoch_{}_{}_slides.png", epoch=epoch, order=order)
+    def get_heatmap_plot_file(self, epoch: int, order: str, stage: ModelKey) -> Path:
+        tmp_filename = "epoch_{}_{}_slides_train.png" if stage == ModelKey.TRAIN else "epoch_{}_{}_slides_val.png"
```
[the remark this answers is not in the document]
kenza-bouzid: Great suggestions, thanks! will definitely add it!

hi-ml-cpath/src/health_cpath/utils/callbacks.py
kenza-bouzid: Oh no particular reason, I just think of slide ids as an "input" to the model that come from the batch, but might be safer to retrieve it from outputs for consistency

hi-ml-cpath/testhisto/testhisto/utils/test_callbacks.py
kenza-bouzid: Yes, I am aware! the reason why I am not using a new batch each time is because I don't have enough batches with the current fixture [mock_data](https://github.com/microsoft/hi-ml/blob/79ced59ca53b3a7b32074734647ffe310d109441/hi-ml-cpath/testhisto/conftest.py#L81) to stimulate 2 train steps + 2 val steps, I had a StopIteration ERROR but maybe I can just increase the number of slides. (we currently have 10 slides, and only 10% for validation, also the reason why I'm using the same dataloader for both train and val).

hi-ml-cpath/src/health_cpath/models/deepmil.py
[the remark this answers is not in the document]
kenza-bouzid: We always return a dict now. Note that it breaks a test in cpath (easy fix like the one here in test_deepmil).

PR #634, opened 2022-10-18
BUG: Ignore PL warning related to ambiguous batch size.

### LINKED ISSUE #633: DeepMIL runs are noisy with Ambiguous batch size warning
We consistently see this warning is all DeepMIL runs due to dynamic bag size (especially during inference).
` UserWarning: Trying to infer the `batch_size` from an ambiguous collection. The batch size we found is 142. To avoid any miscalculations, use `self.log(..., batch_size=batch_size)`.
  "Trying to infer the `batch_size` from an ambiguous collection. The batch size we"`

kenza-bouzid:
Closes #633

### DIFF
--- a/hi-ml/src/health_ml/lightning_container.py
+++ b/hi-ml/src/health_ml/lightning_container.py
@@ -50,7 +50,7 @@ class LightningContainer(WorkflowParams,
         operations on the present object. At the point when called, the datasets are already available in
         the locations given by self.local_datasets. Use this method to prepare datasets or data loaders, for example.
         """
-        pass
+        self.ignore_pl_warnings()
 
     def create_model(self) -> LightningModule:  # type: ignore
         """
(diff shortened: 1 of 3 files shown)


dccastro:
I'm not sure ignoring the warning is the best approach. It's probably better to actually fix it, otherwise there's a risk that we've been looking at biased metrics all along. When logging tensors (e.g. losses; not TorchMetrics objects), Lightning [weighs the step values by batch size](https://github.com/Lightning-AI/lightning/blob/2d3db64b7c00bb6b5f8dcca395a46cc7368df665/src/pytorch_lightning/trainer/connectors/logger_connector/result.py#L248-L251) when aggregating epoch metrics. My suggestion would be to ensure we're able to pass `batch_size=...` wherever we're logging with Lightning to avoid subtle metric aggregation bugs.

hi-ml/src/health_ml/lightning_container.py
```diff
@@ -237,6 +237,11 @@ def on_run_extra_validation_epoch(self) -> None:
             logging.warning("Hook `on_run_extra_validation_epoch` is not implemented by lightning module."
                             "The extra validation epoch won't produce any extra outputs.")
 
+    def ignore_pl_warnings(self) -> None:
```
ant0nsc, line 240: I think that's a bit too specialized. No problem if that is in BaseMIL, but I would not place it here.

ant0nsc (changes requested):
I would suggest to move the ignore method out of the base class, it seems too specific a usecase.

PR #636, opened 2022-10-19
FIX: Fix test_get_dataset

mebristo:
I accidentally created a v2 data asset with the same name as a v1 dataset in our test suite. The data asset cannot be removed or renamed, hence I am renaming the v1 dataset in the test to get it passing again

### DIFF
--- a/hi-ml-azure/testazure/testazure/test_datasets.py
+++ b/hi-ml-azure/testazure/testazure/test_datasets.py
@@ -164,7 +164,7 @@ def test_get_dataset() -> None:
     Test if a dataset that does not yet exist can be created from a folder in blob storage
     """
     # A folder with a single tiny file
-    tiny_dataset = "himl-tiny_dataset"
+    tiny_dataset = "himl_tiny_dataset"
     workspace = DEFAULT_WORKSPACE.workspace
     # When creating a dataset, we need a non-empty name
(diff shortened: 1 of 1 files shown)


hi-ml-azure/testazure/testazure/test_datasets.py
```diff
@@ -164,7 +164,7 @@ def test_get_dataset() -> None:
     Test if a dataset that does not yet exist can be created from a folder in blob storage
     """
     # A folder with a single tiny file
-    tiny_dataset = "himl-tiny_dataset"
+    tiny_dataset = "himl_tiny_dataset"
```
kenza-bouzid, line 167: [Minor] Can we maybe add an explicit v1 here?

kenza-bouzid (approved):

hi-ml-azure/testazure/testazure/test_datasets.py
mebristo: Dataset is a v1 concept - in v2 they're called data assets

mebristo: Or do you mean in the name of the dataset itself? (Rather than the name of the variable)

mebristo: Would we then need to append "v1" to all dataset names?

PR #639, opened 2022-10-21
ENH: Run extra validation epoch on a single gpu

kenza-bouzid:
Kill DDP processes after run_training and initialize a new trainer instance with a single device for inference

### DIFF
--- a/hi-ml/src/health_ml/run_ml.py
+++ b/hi-ml/src/health_ml/run_ml.py
@@ -259,7 +259,7 @@ class MLRunner:
 
     def get_trainer_for_inference(self, checkpoint_path: Optional[Path] = None) -> Trainer:
         # We run inference on a single device because distributed strategies such as DDP use DistributedSampler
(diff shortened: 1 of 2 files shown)


hi-ml/src/health_ml/run_ml.py
```diff
@@ -257,6 +257,25 @@ def is_crossval_disabled_or_child_0(self) -> bool:
             return self.container.crossval_index == 0
         return True
 
+    def get_trainer_for_inference(self, checkpoint_path: Optional[Path] = None) -> Trainer:
+        # We run inference on a single device because distributed strategies such as DDP use DistributedSampler
+        # internally, which replicates some samples to make sure all devices have some batch size in case of
```
ant0nsc, line 262:
```suggestion
        # internally, which replicates some samples to make sure all devices have the same batch size in case of
```

ant0nsc (approved):

PR #641, opened 2022-10-25
BUG: Retrieve tiles_ids from outputs dictionary for loss analysis

kenza-bouzid:
Slides pipeline don't have tile_ids in the batch, we instead retrieve them from results


hi-ml-cpath/src/health_cpath/utils/callbacks.py
```diff
@@ -440,7 +440,7 @@ def update_loss_cache(self, trainer: Trainer, outputs: BatchResultsType, batch:
             self.loss_cache[stage][ResultsKey.ENTROPY].extend(self.compute_entropy(outputs[ResultsKey.CLASS_PROBS]))
             if self.save_tile_ids:
                 self.loss_cache[stage][ResultsKey.TILE_ID].extend(
-                    [self.TILES_JOIN_TOKEN.join(tiles) for tiles in batch[ResultsKey.TILE_ID]]
+                    [self.TILES_JOIN_TOKEN.join(tiles) for tiles in outputs[ResultsKey.TILE_ID]]
```
dccastro, line 443: If we can retrieve also the slide IDs from `outputs`, might it slightly simplify the pipeline to remove `batch` as an input to this method?

dccastro (approved):

PR #643, opened 2022-10-27
ENH: HEDJitter update

maxilse:
Updated the HEDjitter augmentation to reflect the update in https://github.com/scikit-image/scikit-image/blob/main/skimage/color/colorconv.py#L1463-L1504

In addition, I included a normalization step. Our implementation is now following exactly https://github.com/gatsby2016/Augmentation-PyTorch-Transforms/blob/master/myTransforms.py while using only Pytorch (instead of skimage).


hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -37,45 +37,44 @@ def __init__(self, theta: float = 0.) -> None:
...
-        :param rgb_from_stain_mat: Transformation matrix from RGB to HED.
         """
-        alpha = torch.FloatTensor(1, 3).uniform_(1 - theta, 1 + theta)
-        beta = torch.FloatTensor(1, 3).uniform_(-theta, theta)
 
-        # Only perturb the H (=0) and E (=1) channels
-        alpha[0][-1] = 1.
-        beta[0][-1] = 0.
+        alpha = torch.FloatTensor(img.shape[0], 3).uniform_(1 - self.theta, 1 + self.theta)
+        beta = torch.FloatTensor(img.shape[0], 3).uniform_(-self.theta, self.theta)
```
dccastro, line 50: Why not already instantiate `alpha` and `beta` with shape `(batch, 1, 1, 3)`, if below you do `[:, None, None, :]` anyway?

hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -37,45 +37,44 @@ def __init__(self, theta: float = 0.) -> None:
...
         # Combine stains
-        img = 10 ** (-stains @ rgb_from_stain_mat) - 2
+        img = -(stains * (-self.log_adjust)) @ self.rgb_from_hed
+        img = torch.exp(img)
         img = torch.clip(img, 0, 1)
         img = img.permute(0, 3, 1, 2)
 
+        # Normalize
+        imin = torch.amin(img, dim=[1, 2, 3])
+        imax = torch.amax(img, dim=[1, 2, 3])
```
dccastro, line 69: Consider passing `torch.amax(..., keepdim=True)` so you don't have to re-add the missing dimensions below with `[:, None, None, None]`.

hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -37,45 +37,44 @@ def __init__(self, theta: float = 0.) -> None:
...
                                           [-0.60190736, -0.48041419, 1.57358807]])
+        self.log_adjust = torch.log(torch.tensor(1E-6))
 
-    @staticmethod
-    def adjust_hed(img: torch.Tensor,
-                   theta: float,
-                   stain_from_rgb_mat: torch.Tensor,
-                   rgb_from_stain_mat: torch.Tensor
-                   ) -> torch.Tensor:
+    def adjust_hed(self, img: torch.Tensor) -> torch.Tensor:
```
dccastro, line 42: [Minor] Could it not be useful to keep this as a static? Maybe turning `rgb_from_hed`, `hed_from_rgb`, and `log_adjust` into class attributes to limit the number of arguments?

hi-ml/testhiml/testhiml/test_data_augmentations.py
```diff
@@ -68,15 +68,27 @@ def test_stain_normalization() -> None:
...
+          [1.000, 0.0888]]]])
+    expected_output_img2 = torch.Tensor(
+        [[[[0.4210, 0.0015],
+          [0.8476, 0.3973]],
+         [[0.2332, 0.0000],
+          [1.000, 0.1661]],
+         [[0.1316, 0.2157],
+          [0.9711, 0.0608]]]])
+    expected_output_bag = torch.stack([expected_output_img1.squeeze(0),
+                                       expected_output_img2.squeeze(0)])
```
dccastro, line 86: [Minor] Looks like with stacking we don't need the spurious zeroth dimension anymore in the tensor definitions above. Shall we get rid of the extra `[ ]`?

dccastro (approved):

PR #644, opened 2022-10-27
ENH: Extend StainNormalisation Transform to work with Slides Pipeline

kenza-bouzid:
Adjust the input/output shapes.

### DIFF
--- a/hi-ml-cpath/src/health_cpath/utils/wsi_utils.py
+++ b/hi-ml-cpath/src/health_cpath/utils/wsi_utils.py
@@ -16,7 +16,10 @@ def image_collate(batch: List) -> Any:
 
     for i, item in enumerate(batch):
         data = item[0]
-        data[SlideKey.IMAGE] = torch.tensor(np.array([ix[SlideKey.IMAGE] for ix in item]))
+        if isinstance(data[SlideKey.IMAGE], torch.Tensor):
+            data[SlideKey.IMAGE] = torch.stack([ix[SlideKey.IMAGE] for ix in item], dim=0)
+        else:
+            data[SlideKey.IMAGE] = torch.tensor(np.array([ix[SlideKey.IMAGE] for ix in item]))
         data[SlideKey.LABEL] = torch.tensor(data[SlideKey.LABEL])
         batch[i] = data
     return multibag_collate(batch)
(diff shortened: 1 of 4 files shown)


hi-ml-cpath/src/health_cpath/utils/wsi_utils.py
```diff
@@ -16,7 +16,10 @@ def image_collate(batch: List) -> Any:
 
     for i, item in enumerate(batch):
         data = item[0]
-        data[SlideKey.IMAGE] = torch.tensor(np.array([ix[SlideKey.IMAGE] for ix in item]))
+        if isinstance(data[SlideKey.IMAGE], torch.Tensor):
+            data[SlideKey.IMAGE] = torch.stack([ix[SlideKey.IMAGE] for ix in item], dim=0)
```
kenza-bouzid, line 20: Stain norm returns tensors

hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -131,14 +131,20 @@ def stain_normalize(img: torch.Tensor, reference_mean: np.ndarray, reference_std
...
         if img.shape[0] > 1:
             for i in range(img.shape[0]):
                 img_tile = img[i]
                 img[i] = self.stain_normalize(img_tile.unsqueeze(0), self.reference_mean, self.reference_std)
             return img
         else:
-            return self.stain_normalize(img, self.reference_mean, self.reference_std)
+            img = self.stain_normalize(img, self.reference_mean, self.reference_std)
+            if len(original_shape) == 3:
+                return img.squeeze(0)
```
kenza-bouzid, line 146: Collate function above expect images (C, H, W) so we need to unsqueeze the image back

harshita-s: unsqueeze or squeeze? shouldn't this be `if len(original_shape) == 4` if there is already an existing batch dimension and you want to squeeze?

kenza-bouzid: If there's already a batch dimension, I don't want to remove it, I only squeeze when original_shape==3 to put it back to 3 cause I add it line 136

harshita-s (approved):

PR #647, opened 2022-11-03
BUG: Fix HTML report 

harshita-s:
In this PR:
- Fix is introduced to handle missing cross-validation rounds (e.g., when a round fails)
- Extra validation epoch is handled
- Downloading the AML metrics json and saving it as a dataframe is separated into two functions (current version gives a `KeyError` when the json is downloaded for the first time)

### DIFF
--- a/hi-ml-cpath/src/health_cpath/utils/report_utils.py
+++ b/hi-ml-cpath/src/health_cpath/utils/report_utils.py
@@ -148,7 +148,7 @@ def get_hyperdrive_metrics_table(metrics_df: pd.DataFrame, metrics_list: Sequenc
         values: pd.Series = metrics_df.loc[metric]
         mean = values.mean()
         std = values.std()
-        round_values: List[str] = [f"{v:.3f}" if v is not None else "N/A" for v in values]
+        round_values: List[str] = [f"{v:.3f}" if v is not None else str(np.nan) for v in values]
         agg_values: List[str] = [f"{mean:.3f} ± {std:.3f}"]
         row = [metric] + round_values + agg_values
         metrics_rows.append(row)
@@ -238,20 +238,30 @@ def get_formatted_run_info(parent_run: Run) -> str:
     return html
 
 
-def collect_class_info(metrics_df: pd.DataFrame) -> Tuple[int, List[str]]:
+def get_child_runs_hyperparams(metrics_df: pd.DataFrame) -> Dict[int, Dict]:
     """
-    Get the class names from metrics dataframe
-    :param metrics_df: Metrics dataframe, as returned by :py:func:`collect_hyperdrive_metrics()` and
+    Get the hyperparameters of each child run from the metrics dataframe.
+    :param: metrics_df: Metrics dataframe, as returned by :py:func:`collect_hyperdrive_metrics()` and
         :py:func:`~health_azure.aggregate_hyperdrive_metrics()`.
-    :return: Number of classes and list of class names
+    :return: A dictionary of hyperparameter dictionaries for the child runs.
     """
-    hyperparams = metrics_df[0][AMLMetricsJsonKey.HYPERPARAMS]
-    hyperparams_name = hyperparams[AMLMetricsJsonKey.NAME]
-    hyperparams_value = hyperparams[AMLMetricsJsonKey.VALUE]
-    num_classes_index = hyperparams_name.index(AMLMetricsJsonKey.N_CLASSES)
-    num_classes = int(hyperparams_value[num_classes_index])
-    class_names_index = hyperparams_name.index(AMLMetricsJsonKey.CLASS_NAMES)
-    class_names = hyperparams_value[class_names_index]
+    hyperparams_children = {}
+    for child_index in metrics_df.columns:
+        hyperparams = metrics_df[child_index][AMLMetricsJsonKey.HYPERPARAMS]
+        hyperparams_dict = dict(zip(hyperparams[AMLMetricsJsonKey.NAME], hyperparams[AMLMetricsJsonKey.VALUE]))
+        hyperparams_children[child_index] = hyperparams_dict
+    return hyperparams_children
+
+
+def collect_class_info(hyperparams_children: Dict[int, Dict]) -> Tuple[int, List[str]]:
+    """
+    Get the class names from the hyperparameters of child runs.
+    :param hyperparams_children: Dict of hyperparameter dicts, as returned by :py:func:`get_child_runs_hyperparams()`.
+    :return: Number of classes and list of class names.
+    """
+    hyperparams_single_run = list(hyperparams_children.values())[0]
+    num_classes = int(hyperparams_single_run[AMLMetricsJsonKey.N_CLASSES])
+    class_names = hyperparams_single_run[AMLMetricsJsonKey.CLASS_NAMES]
     if class_names == "None":
         class_names = None
     else:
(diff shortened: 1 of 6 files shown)


hi-ml-cpath/src/health_cpath/utils/report_utils.py
```diff
@@ -139,13 +148,16 @@ def get_hyperdrive_metrics_table(metrics_df: pd.DataFrame, metrics_list: Sequenc
         values: pd.Series = metrics_df.loc[metric]
         mean = values.mean()
         std = values.std()
-        row = [metric] + [f"{v:.3f}" for v in values] + [f"{mean:.3f} ± {std:.3f}"]
+        round_values: List[str] = [f"{v:.3f}" if v is not None else f"{v}" for v in values]
```
dccastro, line 151: [Minor] Here I presume `f"{v}"` evaluates to `"None"` for the missing values, right? How about replacing this with a literal, like "N/A", "–", "(missing)", ...?

hi-ml-cpath/src/health_cpath/utils/report_utils.py
```diff
@@ -154,12 +166,21 @@ def get_best_epochs(metrics_df: pd.DataFrame, primary_metric: str, maximise: boo
...
+    best_epochs = {}
+    for i in metrics_df.columns:
+        primary_metric_list = metrics_df[i][primary_metric]
+        if primary_metric_list is not None:
+            # If extra validation epoch was logged (N+1), return only the first N elements
+            primary_metric_list = primary_metric_list[:-1] \
+                if (len(primary_metric_list) == max_epochs_dict[i] + 1) else primary_metric_list
+            best_epochs[i] = int(np.argmax(primary_metric_list) if maximise else np.argmin(primary_metric_list))
+        else:
+            best_epochs[i] = None           # type:ignore
```
dccastro, line 182: Alternatively, why not just omit the key from the dictionary? That way you don't need to check for `None` in `get_best_epoch_metrics()` below.

hi-ml-cpath/src/health_cpath/utils/report_utils.py
```diff
@@ -237,3 +258,21 @@ def collect_class_info(metrics_df: pd.DataFrame) -> Tuple[int, List[str]]:
...
+    Get the maximum number of epochs for each round from metrics dataframe
+    :param metrics_df: Metrics dataframe, as returned by :py:func:`collect_hyperdrive_metrics()` and
+        :py:func:`~health_azure.aggregate_hyperdrive_metrics()`.
+    :return: Number of classes and list of class names
+    """
+    max_epochs_dict = {}
+    for i in metrics_df.columns:
+        hyperparams = metrics_df[i][AMLMetricsJsonKey.HYPERPARAMS]
+        hyperparams_name = hyperparams[AMLMetricsJsonKey.NAME]
+        hyperparams_value = hyperparams[AMLMetricsJsonKey.VALUE]
```
dccastro, line 274: [Optional] One suggestion here is to convert `hyperparams` into a dictionary, like `dict(zip(hyperparams[NAME], hyperparams[VALUE]))`. This can then be made into a more generic and reusable function that returns all hyperparameters for each child run. What do you think?

hi-ml-cpath/src/health_cpath/utils/report_utils.py
```diff
@@ -237,3 +258,21 @@ def collect_class_info(metrics_df: pd.DataFrame) -> Tuple[int, List[str]]:
         class_names = [name.lstrip() for name in class_names[1:-1].replace("'", "").split(',')]
     class_names = validate_class_names(class_names=class_names, n_classes=num_classes)
     return (num_classes, list(class_names))
+
+
+def collect_epoch_info(metrics_df: pd.DataFrame) -> Dict[int, int]:
```
dccastro, line 263: I'd suggest renaming this function `get_max_epochs()` or similar to make its purpose crystal-clear.

hi-ml-cpath/src/health_cpath/utils/report_utils.py
```diff
@@ -237,3 +258,21 @@ def collect_class_info(metrics_df: pd.DataFrame) -> Tuple[int, List[str]]:
...
+    """
+    Get the maximum number of epochs for each round from metrics dataframe
+    :param metrics_df: Metrics dataframe, as returned by :py:func:`collect_hyperdrive_metrics()` and
+        :py:func:`~health_azure.aggregate_hyperdrive_metrics()`.
+    :return: Number of classes and list of class names
```
dccastro, line 268: This docstring needs updating.

hi-ml-cpath/testhisto/testhisto/utils/test_report_utils.py
```diff
@@ -212,7 +215,9 @@ def test_collect_hyperdrive_metrics(metrics_df: pd.DataFrame, tmp_path: Path, ov
 
 @pytest.mark.parametrize('maximise', [True, False])
 def test_get_best_epochs(metrics_df: pd.DataFrame, maximise: bool) -> None:
-    best_epochs = get_best_epochs(metrics_df, 'val/accuracy', maximise=maximise)
+    max_epochs_dict = {0: 3, 1: 3, 3: 3}
```
dccastro, line 218: We seem to be missing two test cases for the fixes introduced in this PR:
- Interrupted run with higher `max_epochs` than available values.
- Failed run with no values at all (`None`).

One idea to implement these is to adapt the existing `metrics_df` fixture. Thoughts?

dccastro (changes requested):
Many thanks for patching these corner cases. My main comment is that we should probably write some tests to make sure they're properly covered and won't break again in the future. Otherwise, just a few minor optional suggestions.

hi-ml-cpath/src/health_cpath/utils/report_utils.py
harshita-s: Yes, it will be None. Now replaced by "N/A"

hi-ml-cpath/src/health_cpath/utils/report_utils.py
harshita-s: good idea, renamed now

hi-ml-cpath/src/health_cpath/utils/report_utils.py
harshita-s: thanks for spotting, changed now!

hi-ml-cpath/src/health_cpath/utils/report_utils.py
```diff
@@ -154,12 +166,21 @@ def get_best_epochs(metrics_df: pd.DataFrame, primary_metric: str, maximise: boo
...
     :param primary_metric: Name of the reference metric to optimise.
+    :max_epochs_dict: A dictionary of the maximum number of epochs in each cross-validation round.
     :param maximise: Whether the given metric should be maximised (minimised if `False`).
     :return: Dictionary mapping each hyperdrive child index to its best epoch.
     """
-    best_fn = np.argmax if maximise else np.argmin
-    best_epochs = metrics_df.loc[primary_metric].apply(best_fn)
-    return best_epochs.to_dict()
+    best_epochs = {}
+    for i in metrics_df.columns:
```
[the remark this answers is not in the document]
harshita-s: good idea, renamed now

hi-ml-cpath/src/health_cpath/utils/report_utils.py
```diff
@@ -154,12 +166,21 @@ def get_best_epochs(metrics_df: pd.DataFrame, primary_metric: str, maximise: boo
...
+        primary_metric_list = metrics_df[i][primary_metric]
+        if primary_metric_list is not None:
+            # If extra validation epoch was logged (N+1), return only the first N elements
+            primary_metric_list = primary_metric_list[:-1] \
+                if (len(primary_metric_list) == max_epochs_dict[i] + 1) else primary_metric_list
```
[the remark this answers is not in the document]
harshita-s: Only if we have more metric elements than the maximum number of epochs (e.g., when we have an extra validation epoch), it will drop the last element. Otherwise, the full list should be considered. Could you please explain the comment a bit more?

hi-ml-cpath/src/health_cpath/utils/report_utils.py
harshita-s: The functions to plot curves expect that the best epochs dictionary has the same number of elements as the number of rounds. I think we should use `None` if the script can't find a best epoch for a round.

hi-ml-cpath/src/health_cpath/utils/report_utils.py
harshita-s: This is a good idea. I have now created a new function to get the hyperparameters, since these are required in `get_max_epochs` and `collect_class_info`

hi-ml-cpath/testhisto/testhisto/utils/test_report_utils.py
harshita-s: Sounds good. I have adapted the fixtures (`metrics_df` and `max_epochs_dict`) to consider the two test cases.

dccastro (approved):

PR #649, opened 2022-11-07
FIX: Fix bug in saving validation outputs

mebristo:
We were previously resetting accuracy to get around a known bug in torchmetrics but the reset occurred before saving, which corrupted validation accuracy plots. It seems this is now fixed and so we can remove the reset. Also update torchmetrics to the latest


kenza-bouzid (approved):
LGTM, thanks Melissa for fixing this issue.

dccastro (approved):

PR #664, opened 2022-11-15
ENH: Upgrade azureml packages

kenza-bouzid:
Our logs are very noisy at the moment with packages compatibility warnings. This should fix it.


hi-ml-cpath/requirements_run.txt
```diff
@@ -1,4 +1,3 @@
-azure-storage-blob==12.5.0
```
kenza-bouzid, line 1: we already have it in hi-ml-azure

mebristo (approved):

ant0nsc (approved):

PR #653, opened 2022-11-09
ENH: Merge Main into Transfer Main

kenza-bouzid:
Sync 32 commits from main

PR #676, opened 2022-11-23
BUG: Add symbolic link that had gotten deleted by accident

ant0nsc:
Undo an accident from #653

### DIFF
--- a/docs/source/README.md
+++ b/docs/source/README.md
@@ -1,0 +1,1 @@
+../../hi-ml-multimodal/README.md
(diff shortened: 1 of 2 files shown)


ozan-oktay (approved):

fepegar (changes requested):
Thanks, Anton. I think the name of the symlink is expected to be `multimodal.md`.

fepegar (approved):
Looking good: https://hi-ml--676.org.readthedocs.build/en/676/multimodal.html

PR #689, opened 2022-11-29
ENH: Enable adding additional env variables from container

kenza-bouzid:
add get_additional_environment_variables to Lightning container to be able to flexibly add experiments specific env variables, e.g. Azure mounting env variables for large file experiments... this can be overridden in the config file directly.

### DIFF
--- a/hi-ml-azure/src/health_azure/utils.py
+++ b/hi-ml-azure/src/health_azure/utils.py
@@ -104,7 +104,6 @@ DEFAULT_ENVIRONMENT_VARIABLES = {
     "MKL_SERVICE_FORCE_INTEL": "1",
     # Switching to a new software stack in AML for mounting datasets
     "RSLEX_DIRECT_VOLUME_MOUNT": "true",
-    "RSLEX_DIRECT_VOLUME_MOUNT_MAX_CACHE_SIZE": "1",
     "DATASET_MOUNT_CACHE_SIZE": "1",
 }
 
--- a/hi-ml/src/health_ml/runner.py
+++ b/hi-ml/src/health_ml/runner.py
@@ -227,6 +227,8 @@ class Runner:
         entry_script = Path(sys.argv[0]).resolve()
         script_params = sys.argv[1:]
 
+        environment_variables = self.additional_environment_variables()
+
         # Get default datastore from the provided workspace. Authentication can take a few seconds, hence only do
         # that if we are really submitting to AzureML.
         workspace: Optional[Workspace] = None
@@ -277,7 +279,7 @@ class Runner:
                 aml_workspace=workspace,
                 ml_client=ml_client,
                 compute_cluster_name=self.experiment_config.cluster,
-                environment_variables=self.additional_environment_variables(),
+                environment_variables=environment_variables,
                 default_datastore=datastore,
                 experiment_name=self.lightning_container.effective_experiment_name,
                 input_datasets=input_datasets,  # type: ignore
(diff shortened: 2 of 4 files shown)


hi-ml-azure/src/health_azure/utils.py
```diff
@@ -104,7 +104,6 @@
     "MKL_SERVICE_FORCE_INTEL": "1",
     # Switching to a new software stack in AML for mounting datasets
     "RSLEX_DIRECT_VOLUME_MOUNT": "true",
-    "RSLEX_DIRECT_VOLUME_MOUNT_MAX_CACHE_SIZE": "1",
```
kenza-bouzid, line 107: deprecated according to AML team

ant0nsc: We can probably also retire the other RSLEX... thing

hi-ml/src/health_ml/runner.py
```diff
@@ -297,7 +299,7 @@ def after_submission_hook(azure_run: Run) -> None:
             azure_run_info = submit_to_azure_if_needed(
                 input_datasets=input_datasets,  # type: ignore
                 submit_to_azureml=False,
-                environment_variables=environment_variables,
+                environment_variables=self.additional_environment_variables(),
```
dccastro, line 302: [Minor] Does `additional_environment_variables()` need to be called down here inside each `if/else` case instead of once above (`environment_variables = ...`)?

ant0nsc (approved):

hi-ml/testhiml/testhiml/test_runner.py
```diff
@@ -113,6 +113,30 @@ def test_additional_aml_run_tags(mock_runner: Runner) -> None:
...
+        with patch("health_ml.runner.Runner.run_in_situ"):
+            with patch("health_ml.runner.Runner.parse_and_load_model"):
+                with patch("health_ml.runner.Runner.validate"):
+                    with patch.object(sys, "argv", arguments):
+                        mock_container = create_autospec(LightningContainer)
+                        mock_container.get_additional_environment_variables = MagicMock(return_value={"foo": "bar"})
+                        mock_runner.lightning_container = mock_container
+                        mock_runner.run()
+        mocks["submit_to_azure_if_needed"].assert_called_once()
+        assert DEBUG_DDP_ENV_VAR in mocks["submit_to_azure_if_needed"].call_args[1]["environment_variables"]
```
dccastro, line 135: [Minor] Maybe extract `mocks["submit_to_azure_if_needed"].call_args[1]["environment_variables"]` to a variable, to simplify?

dccastro (approved):

hi-ml/src/health_ml/runner.py
kenza-bouzid: good point, we can call it once above 
but note that this call is only reached once anyways, we're either running outside or in aml

hi-ml-azure/src/health_azure/utils.py
kenza-bouzid: Oh no actually, RSLEX_DIRECT_VOLUME_MOUNT might be essential. I quote from azure doc 
"For read only mount environment variable RSLEX_DIRECT_VOLUME_MOUNT="true" should be set before starting mount."
"For writable mount environment variable RSLEX_DIRECT_VOLUME_WRITABLE_MOUNT="true" should be set before starting mount."

Not sure, what the default is, but to be on the safe side let's keep it read only.

PR #696, opened 2022-12-19
ENH: Enable overriding tiling on the fly steps

kenza-bouzid:
add get_tiling_transforms to be able to flexibly override tiling on the fly steps in case we need to add additional preprocessing before tiling, like normalizing the background...

### DIFF
--- a/hi-ml-cpath/src/health_cpath/datamodules/base_module.py
+++ b/hi-ml-cpath/src/health_cpath/datamodules/base_module.py
@@ -5,7 +5,7 @@
 import torch
 from enum import Enum
 from pathlib import Path
-from typing import Any, Callable, Dict, Generic, Optional, Sequence, Tuple, TypeVar, Union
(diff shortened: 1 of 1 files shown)


hi-ml-cpath/src/health_cpath/datamodules/base_module.py
```diff
@@ -318,3 +311,13 @@ def _get_dataloader(self, dataset: SlidesDataset, stage: ModelKey, shuffle: bool
             generator=generator,
             **dataloader_kwargs,
         )
+
+    def _get_tiling_on_the_fly_steps_transforms(self, stage: ModelKey) -> List[Callable]:
```
ant0nsc, line 315: nitpick on names - this tiling will always be "on the fly". Also, if people are meant to override it, it should not be private. How about `get_tiling_transforms`?

ant0nsc (approved):
Looking good, just one suggestion regarding naming

dccastro (approved):

PR #420, opened 2022-06-08
DEL: Retire the --azureml flag of the runner

kenza-bouzid:
Once cluster name is given, it should run in AML no need to have an extra flag.

PR #709, opened 2022-12-21
BUG: Fix broken `--azureml` flag

### LINKED ISSUE #690: `--azureml` flag not working as expected
It seems that the `--azureml` flag is not working, at least not in the the [first steps docs](https://hi-ml.readthedocs.io/en/latest/first_steps.html) explains it.

If I try running my script with the `--azureml` flag it runs locally, unless I specify `submit_to_azureml=True` in the call to `submit_to_azureml_if_needed()`. After having a look at the code in [himl.py](https://github.com/microsoft/hi-ml/blob/main/hi-ml-azure/src/health_azure/himl.py#L834) this behaviour makes sense, as the code is run locally `if not submit_to_azureml`, where `submit_to_azureml` is set to `None` even if the `--azureml` flag is used. 

Not sure if I'm missing something here, but it appears the logic for checking for `--azureml` in `sys.argv` is missing.

peterhessey:
Closes #690.

Adds back in accidentally-removed `--azureml` flag logic and adds test to catch it breaking in the future.

### DIFF
--- a/hi-ml-azure/src/health_azure/himl.py
+++ b/hi-ml-azure/src/health_azure/himl.py
@@ -831,6 +831,8 @@ def submit_to_azure_if_needed(  # type: ignore
             return _generate_v2_azure_datasets(cleaned_input_datasets, cleaned_output_datasets)
     # This codepath is reached when executing outside AzureML. Here we first check if a script submission to AzureML
     # is necessary. If not, return to the caller for local execution.
+    if submit_to_azureml is None:
+        submit_to_azureml = AZUREML_COMMANDLINE_FLAG in sys.argv[1:]
     if not submit_to_azureml:
         # Set the environment variables for local execution.
         environment_variables = {
--- a/hi-ml-azure/src/health_azure/examples/elevate_this.py
+++ b/hi-ml-azure/src/health_azure/examples/elevate_this.py
@@ -26,7 +26,8 @@ def main() -> None:
     _ = submit_to_azure_if_needed(
         compute_cluster_name="lite-testing-ds2",
         wait_for_completion=True,
-        wait_for_completion_show_output=True)
+        wait_for_completion_show_output=True,
+        strictly_aml_v1=True)
 
     parser = ArgumentParser()
     parser.add_argument("-m", "--message", type=str, required=True, help="The message to print out")
--- a/hi-ml-azure/testazure/testazure/test_data/simple/hello_world_template.txt
+++ b/hi-ml-azure/testazure/testazure/test_data/simple/hello_world_template.txt
@@ -52,7 +52,7 @@ def main() -> None:
         wait_for_completion={{ wait_for_completion }},
         wait_for_completion_show_output={{ wait_for_completion_show_output }},
         tags={{ tags }},
-        submit_to_azureml={{ submit_to_azureml}},
+        submit_to_azureml={{ submit_to_azureml }},
         strictly_aml_v1={{ strictly_aml_v1 }})
 
     parser = ArgumentParser()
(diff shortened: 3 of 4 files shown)


kenza-bouzid (approved):

mebristo:
We decided to remove this flag and using the --compute flag instead (if a compute has been provided and is not 'local' we can assume the user wants to submit to AML). This was to reduce the number of command line args. Is it needed?

kenza-bouzid:
> We decided to remove this flag and using the --compute flag instead (if a compute has been provided and is not 'local' we can assume the user wants to submit to AML). This was to reduce the number of command line args. Is it needed?

Melissa, I think we only retired it in the runner  #420

mebristo:
@kenza-bouzid  not sure I understand - what other scripts than the runner would use this command line arg?

kenza-bouzid:
> @kenza-bouzid not sure I understand - what other scripts than the runner would use this command line arg?

@mebristo We sometimes submit custom scripts to azureml (tiling, create montages...) using submit_to_azure_if_needed with the current file as entry_script.

mebristo:
kenza-bouzid  but `submit_to_azure_if_needed` has an arg 'submit_to_azureml', so as far as I can see the command line arg still isn't needed?

ant0nsc:
> kenza-bouzid but `submit_to_azure_if_needed` has an arg 'submit_to_azureml', so as far as I can see the command line arg still isn't needed?

@mebristo You're right - but in order to use that arg, you already need to have an argument parser. The idea with this hardcoded magic flag was to keep the required extra code at a minimum. 

Happy to hear other ideas how to support the idea of "minimally invasive change" for getting people onto Azure - it's not clear that we have chosen the best one.

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -831,6 +831,8 @@ def submit_to_azure_if_needed(  # type: ignore
             return _generate_v2_azure_datasets(cleaned_input_datasets, cleaned_output_datasets)
     # This codepath is reached when executing outside AzureML. Here we first check if a script submission to AzureML
     # is necessary. If not, return to the caller for local execution.
+    if submit_to_azureml is None:
+        submit_to_azureml = AZUREML_COMMANDLINE_FLAG in sys.argv[1:]
```
ant0nsc, line 835: There's one more thing we need to do: Before script submission, modify the argument list from `sys.argv` and remove this magic flag (line 926). Otherwise, the argument parser would need to be able to cope with the existence of `--azureml`

ant0nsc (changes requested):
One more small change necessary, see comment.

mebristo:
@ant0nsc would it not work to continue with the strategy of looking for a `--compute` flag that was neither None nor 'local'?

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -921,7 +923,8 @@ def submit_to_azure_if_needed(  # type: ignore
             )
             if entry_script is None:
                 entry_script = Path(sys.argv[0])
-            script_params = script_params or sys.argv[1:]
+
+            script_params = script_params or [param for param in sys.argv[1:] if param != AZUREML_COMMANDLINE_FLAG]
```
ant0nsc, line 927: you will also need a matching change for submission using the V1 SDK. Consequently, it might make sense to move that messing with `script_params` to before line 883, so that both codepaths are affected in the same way. Please also add tests.

ant0nsc (changes requested):
sorry, my comment was not comprehensive enough - V1 codepath is not yet covered by your latest change

mebristo:
@peterhessey Sorry it's actually called `--cluster`!

ant0nsc (approved):

PR #718, opened 2023-01-05
ENH: Reduce logging noise in test suite

### LINKED ISSUE #708: Reduce logging noise in pytest runs
We have debug-level logs from `msrest` and other libraries in the output for pytest, making results hard to navigate. Should call the equivalent of `package_setup_and_hacks` in all `conftest.py`

ant0nsc:
Output of the test suite is cluttered and difficult to navigate in the browser. Reduce output from dependent packages.

Closes #708

### DIFF
--- a/hi-ml-azure/testazure/testazure/test_himl.py
+++ b/hi-ml-azure/testazure/testazure/test_himl.py
@@ -1587,7 +1587,7 @@ def test_submit_to_azure_if_needed_v2() -> None:
 
     with patch.multiple(
         "health_azure.himl",
-        package_setup=DEFAULT,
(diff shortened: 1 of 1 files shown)


kenza-bouzid (approved):
LGTM!

peterhessey (changes requested):
LGTM in terms of functionality. 

I feel like there should be a generic util function added for setting log levels (e.g. pass in a dict of package names and corresponding desired log level) that can be called by each of the `package_setup()` functions, as repeating `logger.setLevel` on multiple lines seems unnecessary and is harder to read than a dictionary.

peterhessey (approved):

PR #721, opened 2023-01-05
ENH: Recreate datamodule for extra validation and inference

kenza-bouzid:
Recreate the datamodule for extra validation and inference, this is useful to reflect any changes in datamodule params (max_bag_size_inf=0 for whole slide inference)


harshita-s (approved):

ant0nsc (approved):
The change makes sense - but I fail to see where the code to modify the datamodule parameters would sit?

kenza-bouzid:
> The change makes sense - but I fail to see where the code to modify the datamodule parameters would sit?

we have a hook on_run_extra_validation_epoch that is called after training is finished. This can be overridden at the container level with any additional changes to the datamodule: batch_size/bag_size...

PR #724, opened 2023-01-09
ENH: Allow return of intermediate layers from ResNet

markpinnock:
This PR allows the returning of intermediate layers for use as skip connections in e.g. U-Net by passing the argument "return_skip=True" during the forward pass.


fepegar:
Thanks, @markpinnock. The mypy check is failing because you should explicitly annotate the return type with all the elements in the tuple. Something like this, assuming three tensors in the tuple:

```python
TypeSkipConnections = Tuple[torch.Tensor, torch.Tensor, torch.Tensor]
```

The notebook issue seems unrelated to this PR.

hi-ml-multimodal/src/health_multimodal/image/model/resnet.py
```diff
@@ -20,7 +22,9 @@ class ResNetHIML(ResNet):
     def __init__(self, **kwargs: Any) -> None:
         super().__init__(**kwargs)
 
-    def forward(self, x: torch.Tensor) -> torch.Tensor:
+    def forward(self, x: torch.Tensor,
+                return_skip: bool = False) -> Union[torch.Tensor, TypeSkipConnections]:
```
ant0nsc, line 26: I think it would be helpful to add a docstring here that describes the arguments and in particular the returned tuple in detail.

hi-ml-multimodal/src/health_multimodal/image/model/resnet.py
```diff
@@ -20,7 +22,15 @@ class ResNetHIML(ResNet):
...
```
ozan-oktay, line 26: As far as I know ResNet doesn't have skip connections as in U-Net. These are intermediate layers. Maybe you could rename it as `return_intermediate_layers`

hi-ml-multimodal/src/health_multimodal/image/model/resnet.py
```diff
@@ -31,7 +41,10 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
         x3 = self.layer3(x2)
         x4 = self.layer4(x3)
 
-        return x4
+        if return_skip:
```
ozan-oktay, line 44: Maybe a simple unit test for this would be helpful:

Input (128x128x3) random tensor -> making sure that the output shape is consistent, e.g. 64x64xn1_features, ... 32x32xn2_features

Many thanks for creating this PR.

hi-ml-multimodal/src/health_multimodal/image/model/resnet.py
fepegar: Another option could be `return_all_feature_maps`, as in `pl_bolts`: https://github.com/Lightning-AI/lightning-bolts/blob/64c4bac1aabeea88a4d577e2522752f5ec685e9a/pl_bolts/models/self_supervised/resnets.py#L261

I suspect their implementation might be more efficient in terms of memory. If we don't care about intermediate activations, maybe overwriting `x0` is better than storing the values of `x0`, `x1`, etc. in memory. I guess that's why a lot of code goes

```
x = layer1(x)
x = layer2(x)
etc
```

ozan-oktay: They are all stored in memory as part of the computation graph for backprop unless the `no_grad` decorator is passed as in inference mode only. I guess memory wouldn't be a big concern at test time.

We didn't use pl_bolts on purpose, their 3party dependencies are not updated on a regular basis.

ozan-oktay: Thanks Fernando. I think either `return_all_feature_maps` or `return_intermediate_layers` should work fine. Let's update it and then we can merge the PR in soon.

hi-ml-multimodal/src/health_multimodal/image/model/resnet.py
ozan-oktay: Thanks for this Mark!

fepegar (approved):
Looks good to me. Thanks, @markpinnock!

ozan-oktay (approved):
👍

PR #732, opened 2023-01-14
Subsample tiles without replacement

harshita-s:
Current transform `SubSampled` to sample and shuffle the tiles does so with replacement. In this PR, we ensure that tiles are sampled without replacement.


ant0nsc (approved):
Can you maybe add a test for this fix? The bug highlights that even the simplest pieces of code can contain critical bugs.

PR #734, opened 2023-01-16
ENH: Simplify MLRunner Inference 

kenza-bouzid:
- [Important] Fix inference only mode checkpoint loading
- Avoid load model checkpoints manually
- Simplify model inference
- Add validation as part of inference only mode 
- Update/Simplify tests

### DIFF
--- a/hi-ml/src/health_ml/deep_learning_config.py
+++ b/hi-ml/src/health_ml/deep_learning_config.py
@@ -175,7 +175,8 @@ class WorkflowParams(param.Parameterized):
     regression_metrics: str = param.String(default=None, doc="A list of names of metrics to compare")
     run_inference_only: bool = param.Boolean(False, doc="If True, run only inference and skip training after loading"
                                                         "model weights from the specified checkpoint in "
-                                                        "`src_checkpoint` flag. If False, run training and inference.")
+                                                        "`src_checkpoint` flag. Inference is run on both validation "
+                                                        "and test sets. If False, run training and inference.")
     resume_training: bool = param.Boolean(False, doc="If True, resume training from the src_checkpoint.")
     tag: str = param.String(doc="A string that will be used as the display name of the run in AzureML.")
     experiment: str = param.String(default="", doc="The name of the AzureML experiment to use for this run. If not "
--- a/hi-ml/src/health_ml/run_ml.py
+++ b/hi-ml/src/health_ml/run_ml.py
@@ -80,6 +80,7 @@ class MLRunner:
         Driver class to run a ML experiment. Note that the project root argument MUST be supplied when using hi-ml
         as a package!
 
+        :param experiment_config: The ExperimentConfig object to use for training.
         :param container: The LightningContainer object to use for training.
         :param project_root: Project root. This should only be omitted if calling run_ml from the test suite. Supplying
         it is crucial when using hi-ml as a package or submodule!
@@ -96,6 +97,7 @@ class MLRunner:
         self.trainer: Optional[Trainer] = None
         self.azureml_run_for_logging: Optional[Run] = None
         self.mlflow_run_for_logging: Optional[str] = None
+        self.inference_ckpt: Optional[str] = None  # Passed to trainer.validate and trainer.test in inference mode
 
     def set_run_tags_from_parent(self) -> None:
         """
@@ -165,9 +167,9 @@ class MLRunner:
             self.set_run_tags_from_parent()
 
     def get_multiple_trainloader_mode(self) -> str:
-        # Workaround for a bug in PL 1.5.5: We need to pass the cycle mode for the training data as a trainer argument
-        # because training data that uses a CombinedLoader is not split correctly in DDP. This flag cannot be passed
-        # through the get_trainer_arguments method of the container because cycle mode is not yet available.
+        """Workaround for a bug in PL 1.5.5: We need to pass the cycle mode for the training data as a trainer argument
+        because training data that uses a CombinedLoader is not split correctly in DDP. This flag cannot be passed
+        through the get_trainer_arguments method of the container because cycle mode is not yet available."""
         multiple_trainloader_mode = "max_size_cycle"
         try:
             from SSL.data.datamodules import CombinedDataModule  # type: ignore
@@ -229,14 +231,6 @@ class MLRunner:
             )
             logging.info(f"Environment variables: {rank_info}. trainer.global_rank: {self.trainer.global_rank}")
 
-    def load_model_checkpoint(self) -> None:
-        if not self.container.run_inference_only:
-            self.checkpoint_handler.additional_training_done()
-        checkpoint_path_for_inference = self.checkpoint_handler.get_checkpoint_to_test()
-        self.container.load_model_checkpoint(checkpoint_path_for_inference)
-        best_epoch = torch.load(checkpoint_path_for_inference).get("epoch", -1)
-        logging.info(f"Checkpoint saved at epoch: {best_epoch}")
-
     def after_ddp_cleanup(self, old_environ: Dict) -> None:
         """
         Run processes cleanup after ddp context to prepare for single device inference.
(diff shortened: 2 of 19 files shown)


hi-ml/src/health_ml/run_ml.py
```diff
@@ -274,81 +268,78 @@ def is_crossval_disabled_or_child_0(self) -> bool:
             return self.container.crossval_index == 0
         return True
 
-    def get_trainer_for_inference(self, checkpoint_path: Optional[Path] = None) -> Trainer:
-        # We run inference on a single device because distributed strategies such as DDP use DistributedSampler
-        # internally, which replicates some samples to make sure all devices have the same batch size in case of
-        # uneven inputs.
+    def set_trainer_for_inference(self) -> None:
```
kenza-bouzid, line 271: so that we set self.trainer for both validation and test

hi-ml/src/health_ml/run_ml.py
```diff
@@ -274,81 +268,78 @@ def is_crossval_disabled_or_child_0(self) -> bool:
...
+        """ Prepare the runner for inference: validation or test. The following steps are performed:
+        1. Get the checkpoint to use for inference. This is either the checkpoint from the last training epoch or the
+        one specified in src_checkpoint argument.
+        2. If the container has a run_extra_val_epoch method, call it to run an extra validation epoch.
+        3. Create a new trainer instance for inference. This is necessary because the trainer is created with a single
+        device in contrast to training that uses DDP if multiple GPUs are available.
+        4. Create a new data module instance for inference to account for any requested changes in the dataloading
+        parameters (e.g. batch_size, max_num_workers, etc) as part of on_run_extra_validation_epoch.
+        """
+        self.inference_ckpt = str(self.checkpoint_handler.get_checkpoint_to_test())
```
kenza-bouzid, line 295: this is to reduce duplications in run_validation and run_inference

hi-ml/src/health_ml/run_ml.py
```diff
@@ -274,81 +268,78 @@ def is_crossval_disabled_or_child_0(self) -> bool:
...
     def run_validation(self) -> None:
-        """
-        Run validation on the validation set for all models to save time/memory consuming outputs.
-        """
-        self.container.on_run_extra_validation_epoch()
-        trainer = self.get_trainer_for_inference(checkpoint_path=None)
-        with change_working_directory(self.container.outputs_folder):
-            trainer.validate(self.container.model, datamodule=self.container.get_data_module())
-
-    def validate_model_weights(self) -> None:
```
kenza-bouzid, line 323: no need to validate this manually, rely on PL checkpoints weights loading. another advatage here is that we shouldn't worry about container/model states anymore cause PL will only load weights. When we load chcekpoint ourselve we should be careful to do it before calling on_run_extra_val_epoch otherwise model's state gets reset from the checkpoint

hi-ml/src/health_ml/run_ml.py
```diff
@@ -274,81 +268,78 @@ def is_crossval_disabled_or_child_0(self) -> bool:
             return self.container.crossval_index == 0
         return True
 
-    def get_trainer_for_inference(self, checkpoint_path: Optional[Path] = None) -> Trainer:
-        # We run inference on a single device because distributed strategies such as DDP use DistributedSampler
-        # internally, which replicates some samples to make sure all devices have the same batch size in case of
-        # uneven inputs.
+    def set_trainer_for_inference(self) -> None:
+        """ Set the runner's trainer for inference: validation or test.
```
ant0nsc, line 272:
```suggestion
        """ Set the runner's PL Trainer object that should be used when running inference on the validation or test set.
```

hi-ml/src/health_ml/run_ml.py
```diff
@@ -396,26 +387,23 @@ def run(self) -> None:
...
+                self.checkpoint_handler.additional_training_done()
+                # Kill all processes besides rank 0 after training is done to start inference on a single device
                 self.after_ddp_cleanup(old_environ)
 
-                # load model checkpoint for custom inference or additional validation step
-                if self.container.has_custom_test_step() or self.container.run_extra_val_epoch:
-                    self.load_model_checkpoint()
+            self.init_inference()
+
+            with logging_section("Model Validation"):
```
ant0nsc, line 404: Lower case for consistency with the other sections?
```suggestion
            with logging_section("Model validation"):
```

ant0nsc (approved):
Looking good!
Would it make sense to also attach a test that asserts on metrics? We have a test somewhere that trains the HelloWorld model. Can we use the checkpoint from that training run and then run in inference-only mode, checking that we get the same metrics?

hi-ml/src/health_ml/run_ml.py
```diff
@@ -274,81 +268,78 @@ def is_crossval_disabled_or_child_0(self) -> bool:
...
-        number_mismatch = 0
-        for name, param in self.container.model.named_parameters():
-            if not torch.allclose(weights[name].cpu(), param):
-                logging.warning(f"Parameter {name} does not match between model and checkpoint.")
-                number_mismatch += 1
-        logging.info(f"Number of mismatched parameters: {number_mismatch}")
+        """Run validation on the validation set for all models to save time/memory consuming outputs. This is done in
+        inference only mode or when the user has requested an extra validation epoch. The cwd is changed to the outputs
+        folder """
+        if self.container.run_extra_val_epoch or self.container.run_inference_only:
```
mebristo, line 321: If run_inference_only is set, why do we want to run validation?

ant0nsc: Maybe we want to have a commandline flag there?

hi-ml/src/health_ml/run_ml.py
```diff
@@ -274,81 +268,78 @@ def is_crossval_disabled_or_child_0(self) -> bool:
...
+        self.trainer, _ = create_lightning_trainer(
             container=self.container,
-            resume_from_checkpoint=checkpoint_path,
             num_nodes=1,
             azureml_run_for_logging=self.azureml_run_for_logging,
             mlflow_run_for_logging=mlflow_run_id
         )
-        return trainer
+
+    def init_inference(self) -> None:
```
mebristo, line 285: Should the title of this method reflect that it is the required initialisation for validation, as well as inference?

hi-ml/src/health_ml/run_ml.py
```diff
@@ -274,81 +268,78 @@ def is_crossval_disabled_or_child_0(self) -> bool:
...
-            assert checkpoint_path is not None
-        else:
-            self.validate_model_weights()
-
-        trainer, _ = create_lightning_trainer(
+        self.trainer, _ = create_lightning_trainer(
             container=self.container,
-            resume_from_checkpoint=checkpoint_path,
             num_nodes=1,
             azureml_run_for_logging=self.azureml_run_for_logging,
```
mebristo, line 281: Do you know if we have a test that ensures that metrics for the 3 different stages (training, validation, inference) still get logged to the same run?

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -175,7 +175,8 @@ class WorkflowParams(param.Parameterized):
     regression_metrics: str = param.String(default=None, doc="A list of names of metrics to compare")
     run_inference_only: bool = param.Boolean(False, doc="If True, run only inference and skip training after loading"
                                                         "model weights from the specified checkpoint in "
-                                                        "`src_checkpoint` flag. If False, run training and inference.")
+                                                        "`src_checkpoint` flag. Inference is run on both validation "
```
mebristo, line 178: Could we have some more explanation somewhere on why inference is done on validation set?

kenza-bouzid: We need it for the experiments we're running at the moment. For example, whole slide inference (that is quite expensive) we sometimes run it  only after training is finished in inference only mode and would like to check validation results as well

hi-ml/src/health_ml/run_ml.py
kenza-bouzid: I think so, `test_get_mlflow_run_id_from_previous_loggers` added in #627

hi-ml/src/health_ml/run_ml.py
kenza-bouzid: I think that we are misusing the term inference. What we are actually doing is an evaluation on the test and validation set. Inference is done with trainer.predict when we don't have a ground truth at all 
image
Whereas test and validate 
image
image
image

hi-ml/src/health_ml/run_ml.py
kenza-bouzid: see comment above about init_inference 
I think we should rename inference to evaluation
Also instead of adding an extra flag what about using a `run_mode`
where, 
RunMode: 
- all: training, validation, testing
- validation 
- testing
- evaluation: both validation and testing?

hi-ml/src/health_ml/run_ml.py
mebristo: I meant more like a test for  `set_trainer_for_inference` specifically that checks that the trainer at each stage has the expected properties (mlflow_run_id and azureml_run_for_logging, num_nodes etc)

kenza-bouzid: Good point, will add a test

hi-ml/src/health_ml/run_ml.py
kenza-bouzid: Will address this in a separate PR

mebristo (approved):

PR #735, opened 2023-01-17
ENH: Remove mypy cache folder in `make clean`

ant0nsc:
There can be `mypy` cache folders at multiple levels of folder hierarchy, remove them all

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,8 +49,8 @@ conda: conda_update call_pip_local
 
 # clean build artifacts
 clean:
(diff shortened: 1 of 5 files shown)


peterhessey (approved):
LGTM 👍

fepegar (approved):
LGTM, but doesn't mypy cache help mypy run faster?

ant0nsc:
> LGTM, but doesn't mypy cache help mypy run faster?

It does indeed. The present makefile _already_ removes the mypy cache, but it is missing some mypy caches in subfolders, that's what this PR adds.

PR #741, opened 2023-01-19
BUG: Training HelloWorld using AML v2 SDK hangs

### LINKED ISSUE #739: Bug: Submitting a single node job via AML v2 SDK does not work
Submitting HelloWorld via v2 hangs, jobs are not even starting the main script

ant0nsc:
This PR fixes a bug introduced in #730. It also adds smoke tests to run training of the `HelloWorld` model in AML, using v1 and v2 SDK, using 1 or 2 nodes.

Closes #739


hi-ml-azure/src/health_azure/himl.py
```diff
@@ -493,14 +493,14 @@ def submit_run_v2(workspace: Optional[Workspace],
...
         the completion of this run (if True).
-    :param wait_for_completion_show_output: If wait_for_completion is True this parameter indicates whether to show the
-        run output on sys.stdout.
     :param workspace_config_path: If not provided with an AzureML Workspace, then load one given the information in this
         config
     :param ml_client: An Azure MLClient object for interacting with Azure resources.
     :param hyperparam_args: A dictionary of hyperparameter search args to pass into a sweep job.
     :param num_nodes: The number of nodes to use for the job in AzureML.
-    :param processes_per_node: The number of processes per node to use for the job in AzureML.
+    :param pytorch_processes_per_node: For plain PyTorch multi-GPU processing: The number of processes per node. This
```
mebristo, line 501: This function shouldn't be called at all in a v1 job so this statement about being ignored in v1 is perhaps confusing

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -560,12 +564,18 @@ def submit_run_v2(workspace: Optional[Workspace],
...
-            },
+            distribution=distribution,  # type: ignore
         )
 
+    if hyperparam_args:
+        param_sampling = hyperparam_args[PARAM_SAMPLING_ARG]
+
+        for sample_param, choices in param_sampling.items():
+            input_datasets_v2[sample_param] = choices.values[0]
+            cmd += f" --{sample_param}=" + "${{inputs." + sample_param + "}}"
```
mebristo, line 575: Do we want to manually add this to the command? I thought AML did already this for each of the children

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -824,7 +818,10 @@ def submit_to_azure_if_needed(  # type: ignore
         will also register the data in this folder as an AzureML dataset.
     :param output_datasets: The script will create a temporary folder when running in AzureML, and while the job writes
         data to that folder, upload it to blob storage, in the data store.
-    :param num_nodes: The number of nodes to use in distributed training on AzureML.
+    :param num_nodes: The number of nodes to use in distributed training on AzureML. When using a value > 1, an
```
mebristo, line 821: "When using a value > 1, an distributed MPI job will be started" <- this is only true if pytorch_processes_per_node_v2=None

Also, what about the third case where num_nodes=1 and pytorch_processes_per_node_v2=None - is this considered a PyTorch job? Is it compatible with PyTorch Lightning?

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -539,16 +539,20 @@ def submit_run_v2(workspace: Optional[Workspace],
...
-    processes_per_node = processes_per_node if processes_per_node >= 1 else 1
-
-    if hyperparam_args:
-        param_sampling = hyperparam_args[PARAM_SAMPLING_ARG]
-
-        for sample_param, choices in param_sampling.items():
-            input_datasets_v2[sample_param] = choices.values[0]
-            cmd += f" --{sample_param}=" + "${{inputs." + sample_param + "}}"
+    if pytorch_processes_per_node is not None:
+        pytorch_processes_per_node = pytorch_processes_per_node if pytorch_processes_per_node >= 1 else 1
```
peterhessey, line 543: Is this in case a value <= 0 is provided? If so I think it might be better to raise an error / include an assert here instead.

peterhessey (approved):
LGTM, only one minor comment

hi-ml/Makefile
```diff
@@ -75,3 +75,26 @@ pytest_and_coverage: pip_test call_pytest_and_coverage
...
+--conda_env ../hi-ml-cpath/environment.yml --cluster=pr-gpu --max_run_duration=15m --wait_for_completion
+endef
+
+# HelloWorld model training on a single node, submitted using the v1 SDK
+smoke_helloworld_v1:
+	python src/health_ml/runner.py --model=health_ml.HelloWorld --strictly_aml_v1=True --tag smoke_helloworld_v1 ${SHARED_ARGS}
+
+# HelloWorld model training on a single node, submitted using the v2 SDK
+smoke_helloworld_v2:
+	python src/health_ml/runner.py --model=health_ml.HelloWorld --strictly_aml_v1=False --tag smoke_helloworld_v1 ${SHARED_ARGS}
```
mebristo, line 92: I think this tag should be smoke_hello_world_v2

hi-ml/src/health_ml/runner.py
```diff
@@ -268,7 +268,6 @@ def after_submission_hook(azure_run: Run, *args: Any) -> None:
                 after_submission=after_submission_hook,
                 tags=self.additional_run_tags(script_params),
                 strictly_aml_v1=self.experiment_config.strictly_aml_v1,
-                processes_per_node_v2=self.lightning_container.num_gpus_per_node(),
```
mebristo, line 271: Should this arg have been renamed to "pytorch_processes_per_node_v2" rather than removed completely?

ant0nsc: No, it should be removed completely, we want an MPI job rather than a plain Pytorch job like you are running.

mebristo: So it will not be possible to run a plain pytorch multi-gpu job using the runner script?

ant0nsc: No, it will not be possible. The runner is not using plain PyTorch, but PL. So it's more: You shouldn't even attempt to run a plain pytorch job with the runner.

hi-ml/Makefile
ant0nsc: good spot! done

hi-ml-azure/src/health_azure/himl.py
ant0nsc: I've cleaned that up, disallowing all those wrong values.

PR #545, opened 2022-08-01
MNT: Prepare hi-ml-multimodal for PyPI

fepegar:
- Simplified `setup.py` in `hi-ml-multimodal` as we are not using CI for these purposes.
- Set initial PyPI version to `0.1.0` as recommended in [Semantic Versioning 2.0.0](https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase)
- Added `MANIFEST.in`, needed to include files such as `requirements_*.txt`, `README` and `LICENSE` in the package distribution
- Tried TestPyPI by running the commands below (from [Packaging Python Projects](https://packaging.python.org/en/latest/tutorials/packaging-projects/))

```console
cd hi-ml-multimodal
python3 -m pip install --upgrade build
python3 -m build
python3 -m pip install --upgrade twine
python3 -m twine upload --repository testpypi dist/*
```

PR #750, opened 2023-01-20
MNT: Bump hi-ml-multimodal patch version

fepegar:
Bumping version to include @markpinnock's changes from #724.

Additionally, I've added some convenience tools for semi-automatic bumping with `make`. For example:

```shell
make bump_version part=patch
```

Next, I will run the commands in https://github.com/microsoft/hi-ml/pull/545#issue-1324136707 (which could also become a `make` command) to publish the new version.

TODO:
- [x] Add `__init__` to config file and re-run
- [x] Disable tagging in config


ozan-oktay (approved):

peterhessey (approved):
LGTM!

PR #752, opened 2023-01-23
ENH: Gaussian blur and pre-processing transforms in SSL

harshita-s:
In this PR:
- `get_transforms` is modified to consider both pre-processing transforms and augmentations. The pre-processing transforms and augmentations can be overridden in the child SSL containers.
- Gaussian blur usage is corrected, as it gives desired result when applied to HWC (channel-last) images in OpenCV


hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -171,17 +171,34 @@ def __init__(self, kernel_size: int, p: float = 0.5, min: float = 0.1, max: floa
...
-            sigma = (self.max - self.min) * np.random.random_sample() + self.min
-            sample = np.array(sample.squeeze())  # type: ignore
-            sample = cv2.GaussianBlur(sample, (self.kernel_size, self.kernel_size), sigma)
-            sample = torch.Tensor(sample).unsqueeze(0)
-
+        if prob < p:
+            sigma = (max - min) * np.random.random_sample() + min
+            sample = sample.permute([0, 2, 3, 1]).squeeze().numpy()   # only 3 channels, color channel last
+            sample = cv2.GaussianBlur(sample, (kernel_size, kernel_size), sigma)
+            sample = torch.Tensor(sample).unsqueeze(0).permute(0, 3, 1, 2)
```
ant0nsc, line 181: I know that you are just moving code around - but while you're touching it, can we maybe clean it up a bit?
- The check to apply could be written via a [binomial generator](https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.binomial.html#numpy.random.Generator.binomial) as `if random.Generator.binomial(n=1, p=p):`
- It would be more readable to write `sigma = random.Generator.uniform(low=min,high=max)`
- Add documentation for the arguments of the function

ant0nsc, line 181: This operations to go from channel first to channel last, and then back - are we doing that somewhere else too? If yes, this would be a good time to introduce 2 helper functions that can be tested in isolation and then re-used.

hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -171,17 +171,34 @@ def __init__(self, kernel_size: int, p: float = 0.5, min: float = 0.1, max: floa
...
 
+    def __call__(self, img: torch.Tensor) -> torch.Tensor:
+        original_shape = img.shape
+        if len(original_shape) == 3:
+            img = img.unsqueeze(0)  # add batch dimension if missing
+        # if the input is a bag of images, gaussian blur needs to run on each image separately
+        if img.shape[0] > 1:
+            for i in range(img.shape[0]):
+                img_tile = img[i]
+                img[i] = self.apply_gaussian_blur(sample=img_tile.unsqueeze(0), kernel_size=self.kernel_size, p=self.p,
```
ant0nsc, line 192: Maybe I don't get something important here, feel free to ping me for a chat, but: I think this is sequence of operations is very wasteful. Apply_gaussian_blur can deal with tensors that have a batch dimension. Why are we going over the image batch here, effectively removing the batch dimension, then adding it back on via unsqueeze?

ant0nsc (approved):
Looks good overall - a couple of comments on readability, and maybe a chance for speedups.

hi-ml/src/health_ml/utils/data_augmentations.py
harshita-s: The opencv Gaussian blur works with single images, similar to our stain normalization transform. we are applying similar checks for stain normalization.

hi-ml/src/health_ml/utils/data_augmentations.py
dccastro: Relatedly, do you remember if there's any reason we're using OpenCV for this and not `torchvision`'s [`gaussian_blur`](https://pytorch.org/vision/main/generated/torchvision.transforms.functional.gaussian_blur.html) (which operates on NCHW tensors)?

hi-ml-cpath/src/SSL/configs/HistoSimCLRContainer.py
```diff
@@ -51,14 +51,15 @@ def _get_transforms(self, augmentation_config: Optional[CfgNode],
 
     @staticmethod
     def get_transforms(apply_augmentations: bool) -> ImageTransformationPipeline:
-        transforms: List[Any] = []
+        transforms: List[Any] = [StainNormalization()]         # Common transforms
```
kenza-bouzid, line 54: do we want to make this change at this level or rather override it in specific configs? 
What about adding a function get_preprocessing_tranforms that returns an empty list by default which we can override in the container, I don't think StainNormalization was part of SimCLR augmentations. 
You can also add a function get_augmentations_transforms() to replace lines 58-62 so that we can flexibly change in sub classed containers without too much duplications

hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -171,17 +171,34 @@ def __init__(self, kernel_size: int, p: float = 0.5, min: float = 0.1, max: floa
...
+                                                  min=self.min, max=self.max)
+            return img
+        else:
+            img = self.apply_gaussian_blur(sample=img, kernel_size=self.kernel_size, p=self.p,
+                                           min=self.min, max=self.max)
+            if len(original_shape) == 3:
```
kenza-bouzid, line 198: might be out of scope but it would be nice to write a wrapper function or even transform to fix the reshaping, the same way we have `transform_dict_adaptor`

hi-ml/src/health_ml/utils/data_augmentations.py
harshita-s: I agree that the squeeze and unsqueeze operations could be simplified.

hi-ml/src/health_ml/utils/data_augmentations.py
harshita-s: I am not sure why we are not using it, but that would definitely simplify things. For our runs, the torchvision version takes much more time (~2x), perhaps this could be the reason?

kenza-bouzid (approved):

hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -171,17 +171,34 @@ def __init__(self, kernel_size: int, p: float = 0.5, min: float = 0.1, max: floa
         self.kernel_size = kernel_size
         self.p = p
 
-    def __call__(self, sample: torch.Tensor) -> torch.Tensor:
+    @staticmethod
+    def apply_gaussian_blur(sample: torch.Tensor, kernel_size: int, p: float, min: float, max: float) -> torch.Tensor:
```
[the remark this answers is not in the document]
harshita-s: It is similar to the stain normalization method in this file, perhaps @dccastro knows why `stain_normalize` was made a static method?

hi-ml/src/health_ml/utils/data_augmentations.py
harshita-s: As discussed, will introduce the helper functions in a separate PR as it will affect both stain normalization and gaussian blur transforms

hi-ml/src/health_ml/utils/data_augmentations.py
harshita-s: I'll open another PR for the changes, since it affects both stain norm and Gaussian blur transforms

hi-ml/src/health_ml/utils/data_augmentations.py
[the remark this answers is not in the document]
harshita-s: It is because in case we want to apply it without instantiating the transform, as informed by @dccastro

hi-ml/src/health_ml/utils/data_augmentations.py
harshita-s: exactly, I have noted this, we could address in a separate PR.

hi-ml/src/health_ml/utils/data_augmentations.py
harshita-s: Thank you, I have made the suggested changes

hi-ml-cpath/src/SSL/configs/HistoSimCLRContainer.py
```diff
@@ -49,16 +49,21 @@ def _get_transforms(self, augmentation_config: Optional[CfgNode],
                 val_transforms = DualViewTransformWrapper(val_transforms)  # type: ignore
         return train_transforms, val_transforms
 
-    @staticmethod
-    def get_transforms(apply_augmentations: bool) -> ImageTransformationPipeline:
-        transforms: List[Any] = []
+    def get_preprocessing_transforms(self) -> List[Any]:
+        return([])
+
+    def get_augmentations(self) -> List[Any]:
```
kenza-bouzid, line 55: [Minor] Can you please change Any by Callable?

hi-ml-cpath/src/SSL/configs/HistoSimCLRContainer.py
harshita-s: I have now made the suggested change, thank you!

hi-ml-cpath/src/SSL/configs/HistoSimCLRContainer.py
harshita-s: sure, done

PR #761, opened 2023-01-23
ENH: Enable distributed inference

kenza-bouzid:
Add --max_num_gpus_inference flag to be able to use all gpus during inference. DDP processes are only killed if max_num_gpus_inference == 1


hi-ml/src/health_ml/run_ml.py
```diff
@@ -225,11 +226,9 @@ def after_ddp_cleanup(self, old_environ: Dict) -> None:
         # We can now use the global_rank of the Lightning model, rather than environment variables, because DDP has set
         # all necessary properties.
         if self.container.model.global_rank != 0:
-            logging.info(f"Terminating training thread with rank {self.container.model.global_rank}.")
+            print_message_with_rank_pid(f"Terminating training thread with rank {self.container.model.global_rank}.")
```
kenza-bouzid, line 229: this wasn't logged at all because PL restricts logs from rank 0 only

hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -520,6 +526,14 @@ class TrainerParams(param.Parameterized):
                                                  "training and shuffle=False during validation. Default to True. Set to"
                                                  "False to set your own sampler.")
 
+    def validate(self) -> None:
+        if self.max_num_gpus_inference > 1 and self.pl_replace_sampler_ddp:
+            logging.warning("The 'pl_replace_sampler_ddp' flag is set to True, but the 'max_num_gpus_inference' > 1. "
+                            "This might bias the inference results due to duplicate samples. Consider setting "
+                            "--pl_replace_sampler_ddp=False if you want to use multiple GPUs for inference."
+                            "pytorch_lightning.overrides.distributed.UnrepeatedDistributedSampler cab be used instead "
```
ant0nsc, line 534:
```suggestion
                            "pytorch_lightning.overrides.distributed.UnrepeatedDistributedSampler can be used instead "
```

hi-ml/src/health_ml/run_ml.py
ant0nsc: At present, print_message... also prints out a big DEBUG sign, which I feel will look a bit out of place here. If we think the information about killing process is important, then I'd remove that DEBUG sign

hi-ml/src/health_ml/run_ml.py
```diff
@@ -244,6 +243,21 @@ def after_ddp_cleanup(self, old_environ: Dict) -> None:
         if torch.distributed.is_initialized():  # type: ignore
             torch.distributed.destroy_process_group()  # type: ignore
 
+    def end_training(self, old_environ: Dict) -> None:
```
ant0nsc, line 246: Consider renaming `old_environ` to something more telling, like `environ_before_training`?

hi-ml/src/health_ml/run_ml.py
```diff
@@ -258,7 +272,7 @@ def set_trainer_for_inference(self) -> None:
         internally, which replicates some samples to make sure all devices have the same batch size in case of
         uneven inputs which biases the results."""
         mlflow_run_id = get_mlflow_run_id_from_trainer(self.trainer)
-        self.container.max_num_gpus = 1
+        self.container.max_num_gpus = self.container.max_num_gpus_inference
```
ant0nsc, line 275: Rather than modifying max_num_gpus in place, make it an argment to `create_lightning_trainer`?

ant0nsc (approved):
Great job! Does not look complicated at all anymore

hi-ml/src/health_ml/run_ml.py
```diff
@@ -244,6 +243,21 @@ def after_ddp_cleanup(self, old_environ: Dict) -> None:
         if torch.distributed.is_initialized():  # type: ignore
             torch.distributed.destroy_process_group()  # type: ignore
 
+    def end_training(self, old_environ: Dict) -> None:
+        """Cleanup after training is done. This is called after the trainer has finished fit to the data. This is called
```
mebristo, line 247: Nit: "after the trainer has finished fitting"

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -220,11 +220,41 @@ def test_run_training() -> None:
...
+            runner.run_training()
+
+        with patch.object(runner.checkpoint_handler, "additional_training_done") as mock_additional_training_done:
+            with patch.object(runner, "after_ddp_cleanup") as mock_after_ddp_cleanup:
+                with patch("health_ml.run_ml.cleanup_checkpoints") as mock_cleanup_ckpt:
+                    old_environ = {"old": "environ"}
+                    runner.end_training(old_environ=old_environ)
+                    mock_additional_training_done.assert_called_once()
+                    mock_cleanup_ckpt.assert_called_once()
+                    if max_num_gpus_inf == 1:
```
mebristo, line 243: Maybe we could also call "assert_not_called" if max_num_gpus_inf is not equal to 1? Also, above we have another value, -1, for max_num_gpus_inf - what is the expected behaviour there? We don't seem to test that case here

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -220,11 +220,41 @@ def test_run_training() -> None:
...
 @pytest.mark.parametrize("run_extra_val_epoch", [True, False])
 @pytest.mark.parametrize("run_inference_only", [True, False])
-def test_init_inference(run_inference_only: bool, run_extra_val_epoch: bool, ml_runner_with_run_id: MLRunner) -> None:
+def test_init_inference(
+    run_inference_only: bool, run_extra_val_epoch: bool, max_num_gpus_inf: int, ml_runner_with_run_id: MLRunner
+) -> None:
     ml_runner_with_run_id.container.run_inference_only = run_inference_only
     ml_runner_with_run_id.container.run_extra_val_epoch = run_extra_val_epoch
+    ml_runner_with_run_id.container.max_num_gpus_inference = max_num_gpus_inf
+    assert ml_runner_with_run_id.container.max_num_gpus == -1
```
mebristo, line 257: Minor comment, could we add a comment here explaining that -1 is the default value for max_num_gpus? At first sight I thought it had something to do with the values passed in for max_num_gpus_inf

mebristo: Also, -1 might not be an insightful value for max_num_gpus_inf, since max_num_gpus would equal -1 both before and after init_inference got called - maybe it's worth using a different value for extra peace of mind?

hi-ml/src/health_ml/run_ml.py
kenza-bouzid: good point, will add a level argument so that we can specify if it's a debug or info or warning

mebristo (approved):

hi-ml/src/health_ml/run_ml.py
kenza-bouzid: done

hi-ml/src/health_ml/run_ml.py
kenza-bouzid: good point will do

hi-ml/testhiml/testhiml/test_run_ml.py
kenza-bouzid: that's a very good point, I think assert_not_called would capture the expected behaviour nicely, that means if the max_num_gpus_inf != 1 we don't want to kill ddp processes and carry on with validation and test

hi-ml/testhiml/testhiml/test_run_ml.py
kenza-bouzid: > Minor comment, could we add a comment here explaining that -1 is the default value for max_num_gpus? At first sight I thought it had something to do with the values passed in for max_num_gpus_inf

sure! I wanted to reflect that the value will change at the end of training

hi-ml/src/health_ml/run_ml.py
kenza-bouzid: done thanks

hi-ml/src/health_ml/run_ml.py
kenza-bouzid: As discussed, will handle this in a separate PR.

PR #765, opened 2023-01-24
BUG: Gather results causes OOM even when save_intermediate_outputs is False

kenza-bouzid:
We discovered a bug after merging #754 val_set_is_dist was used as a flag to gather_results. It was defined as
 val_set_is_dist  = pl_replace_sampler_ssp and max_num_gpus > 1 
 However max_num_gpus  is set to -1 by default. So val_set_is_dist  was always False. 
This blows up the memory when we use very large bag_size and we're not interester in intermediate validation outputs. 
Therefore we only gather if save_intermediate_outputs is True or if it's an extra validation epoch.
A test is added to make sure gather_results is called when expected only

### DIFF
--- a/hi-ml-cpath/src/health_cpath/utils/output_utils.py
+++ b/hi-ml-cpath/src/health_cpath/utils/output_utils.py
@@ -355,13 +355,13 @@ class DeepMILOutputsHandler:
         :param epoch: Current epoch number.
         :param on_extra_val: Whether this is an extra validation epoch (e.g. after training).
         """
-        epoch_results = gather_results(epoch_results)
+        if self.save_intermediate_outputs or on_extra_val:
+            epoch_results = gather_results(epoch_results)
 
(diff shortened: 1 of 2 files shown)


hi-ml-cpath/src/health_cpath/utils/output_utils.py
```diff
@@ -355,13 +355,13 @@ def save_validation_outputs(self, epoch_results: EpochResultsType, metrics_dict:
         :param epoch: Current epoch number.
         :param on_extra_val: Whether this is an extra validation epoch (e.g. after training).
```
kenza-bouzid, line 356: This is a very complicated logic, we should simplify this in the future. Creating an issue

kenza-bouzid: #766

ant0nsc (approved):

harshita-s (approved):

Issue #753, opened 2023-01-23
Job submission via SDK v2 does not create a data asset if it does not yet exist


ant0nsc:
Repro:
- Try to use a `DatasetConfig` with `submit_to_azure_if_neede` with a dataset that does not yet exist.
- Set `strictly_aml_v1` to False

The code will fail in `create_v2_inputs` here:
```
        version = input_dataset.version or 1
        data_asset: Data = ml_client.data.get(input_dataset.name, version=str(version))
        data_path = data_asset.id or ""
```

```
  File "hi-ml/hi-ml/src/health_ml/runner.py", line 246, in submit_to_azureml_if_needed
    azure_run_info = submit_to_azure_if_needed(
  File "hi-ml/hi-ml-azure/src/health_azure/himl.py", line 967, in submit_to_azure_if_needed
    input_datasets_v2 = create_v2_inputs(ml_client, cleaned_input_datasets)
  File hi-ml/hi-ml-azure/src/health_azure/himl.py", line 715, in create_v2_inputs
    data_asset: Data = ml_client.data.get(input_dataset.name, version=str(version))
  File "site-packages/azure/ai/ml/operations/_data_operations.py", line 148, in get
    data_version_resource = self._operation.get(
  File "site-packages/azure/core/tracing/decorator.py", line 78, in wrapper_use_tracer
    return func(*args, **kwargs)
  File "site-packages/azure/ai/ml/_restclient/v2022_05_01/operations/_data_versions_operations.py", line 455, in get
    map_error(status_code=response.status_code, response=response, error_map=error_map)
  File "site-packages/azure/core/exceptions.py", line 107, in map_error
    raise error
azure.core.exceptions.ResourceNotFoundError: (UserError) Data version crop_and_convert_he_10x_to_tiff_no_hardcoded_bckd_2023_01_17:1 (dataContainerName:version) not found.
Code: UserError
Message: Data version crop_and_convert_he_10x_to_tiff_no_hardcoded_bckd_2023_01_17:1 (dataContainerName:version) not found.
```

PR #772, opened 2023-01-27
BUG: Fix create v2 asset bug

### LINKED ISSUE #757: Bug: Retrieving a V2 data asset reads version 1 by default
See https://github.com/microsoft/hi-ml/blob/main/hi-ml-azure/src/health_azure/himl.py#L709
Code should read the latest version unless specified otherwise

peterhessey:
Closes #757 and #753

### DIFF
--- a/hi-ml-azure/src/health_azure/datasets.py
+++ b/hi-ml-azure/src/health_azure/datasets.py
@@ -140,19 +140,19 @@ def _get_latest_v2_asset_version(ml_client: MLClient, data_asset_name: str) -> s
     :return: The latest version of the data asset if found, else None.
     """
     data_assets = ml_client.data.list(name=data_asset_name)
-    highest_version = 0
-    for data_asset in data_assets:
+    highest_version = float('-inf')
 
+    for data_asset in data_assets:
         try:
             data_asset_version = int(data_asset.version)
         except ValueError as val_er:
             logging.warning(f"Failed to convert data asset version to int: {val_er}")
-            data_asset_version = -1
+            continue
 
         if data_asset_version > highest_version:
             highest_version = data_asset_version
 
-    if highest_version == 0:
+    if highest_version == float('-inf'):
         raise ResourceNotFoundError(f"No data asset found with the provided name: {data_asset_name}")
 
     return str(highest_version)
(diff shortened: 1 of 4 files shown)


hi-ml-azure/src/health_azure/himl.py
```diff
@@ -704,8 +705,14 @@ def create_v2_inputs(ml_client: MLClient, input_datasets: List[DatasetConfig]) -
     inputs: Dict[str, Input] = {}
     for i, input_dataset in enumerate(input_datasets):
         input_name = f"INPUT_{i}"
-        version = input_dataset.version or 1
-        data_asset: Data = ml_client.data.get(input_dataset.name, version=str(version))
+        version = input_dataset.version
+        # if version is None, this function gets the latest version
+        data_asset: Data = _get_or_create_v2_data_asset(
```
ant0nsc, line 710: Can you make sure we are printing progress information here, like "now trying to access data asset foo version None in workspace bar"

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -704,8 +705,17 @@ def create_v2_inputs(ml_client: MLClient, input_datasets: List[DatasetConfig]) -
...
+        version = input_dataset.version
+        # if version is None, this function gets the latest version
+        logging.info(
+            f"Trying to access data asset {input_dataset.name} version {version}, datastore {input_dataset.datastore}"
+        )
+        data_asset: Data = _get_or_create_v2_data_asset(
+            ml_client,
+            input_dataset.datastore,
+            input_dataset.name,
+            version=str(version),
```
kenza-bouzid, line 717: if version is None this will be casted as 'None', so the check line 176 won't be valid, is that intended?
image

kenza-bouzid, line 717:
```suggestion
            version=str(version) if version else version,
```

kenza-bouzid (changes requested):

hi-ml-azure/src/health_azure/himl.py
peterhessey: Nope not intended, just me being lazy when trying to fix pyright. Good spot, thank you!

hi-ml-azure/src/health_azure/datasets.py
```diff
@@ -152,6 +152,9 @@ def _get_latest_v2_asset_version(ml_client: MLClient, data_asset_name: str) -> s
         if data_asset_version > highest_version:
             highest_version = data_asset_version
 
+    if highest_version == 0:
+        raise ResourceNotFoundError(f"No data asset found with the provided name: {data_asset_name}")
```
ant0nsc, line 156: Nitpick - this could strictly speaking also happen if one of the data_asset.version == "0", might be safer to work with None as the default?

kenza-bouzid (approved):

PR #774, opened 2023-01-27
ENH: Make display names a first class citizen

ant0nsc:
Display names for the submitted jobs can be passed in via an argument now, rather than being set in the submission hook


hi-ml-azure/src/health_azure/himl.py
```diff
@@ -648,6 +635,8 @@ def submit_run(workspace: Workspace,
         the completion of this run (if True).
     :param wait_for_completion_show_output: If wait_for_completion is True this parameter indicates whether to show the
         run output on sys.stdout.
+    :param display_name: The name that will be displayed in the AML UI. If not provided, the unique run ID will be
+        used as the display name.
```
fepegar, line 639: Does this mean we don't get the random AML names e.g. `hungry_beaver`?

ant0nsc: you're right, i have  not phrased that correctly

ant0nsc: fixed

fepegar: Phew. I really enjoyed those :D

fepegar (approved):

mebristo (approved):

PR #778, opened 2023-01-31
ENH: Add extra slides to attention heatmaps plots

kenza-bouzid:
Add option to include an extra slide to the the attention heatmap plots.


harshita-s (approved):

hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
```diff
@@ -228,6 +229,9 @@ def get_data_module(self) -> HistoDataModule:
     def get_slides_dataset(self) -> Optional[SlidesDataset]:
         return None
 
+    def get_ihc_slides_dataset(self) -> Optional[SlidesDataset]:
+        return None
+
```
ant0nsc, line 234: The naming is very specific to our use case. How about a name like "extra_slides_for_plotting"? This is more generic, and at the same time also indicates clearly that these slides are only used for plotting.

hi-ml-cpath/src/health_cpath/utils/plots_utils.py
```diff
@@ -216,13 +219,17 @@ def __init__(
...
-        """Returns the slide dictionary for a given slide node"""
-        assert self.slides_dataset is not None, "Cannot plot attention heatmap or wsi without slides dataset"
-        slide_index = self.slides_dataset.dataset_df.index.get_loc(slide_node.slide_id)
+    def get_slide_dict(self, slide_node: SlideNode, slides_dataset: SlidesDataset) -> SlideDictType:
+        """Returns the slide dictionary for a given slide node from a slides dataset.
+
+        :param slide_node: The slide node that encapsulates the slide metadata.
+        :param slides_dataset: The slides dataset that contains the slide image and other metadata.
+        """
+        slide_index = slides_dataset.dataset_df.index.get_loc(slide_node.slide_id)
```
ant0nsc, line 230: What if the slide_id is not found? would be good to have an error message that also prints out the missing slide_id.

hi-ml-cpath/src/health_cpath/utils/plots_utils.py
```diff
@@ -216,13 +219,17 @@ def __init__(
         self.should_upscale_coords = loading_params.should_upscale_coordinates()
         self.loading_params.set_roi_type_to_foreground()
         self.slides_dataset: Optional[SlidesDataset] = None
+        self.ihc_slides_dataset: Optional[SlidesDataset] = None
 
-    def get_slide_dict(self, slide_node: SlideNode) -> SlideDictType:
-        """Returns the slide dictionary for a given slide node"""
-        assert self.slides_dataset is not None, "Cannot plot attention heatmap or wsi without slides dataset"
-        slide_index = self.slides_dataset.dataset_df.index.get_loc(slide_node.slide_id)
+    def get_slide_dict(self, slide_node: SlideNode, slides_dataset: SlidesDataset) -> SlideDictType:
```
ant0nsc, line 224: This is now effectively a static method. You could preserve the previous usage pattern by using `slides_dataset` if passed as an argument, and otherwise using `self.slides_dataset` (just a thought, feel free to disagree!)

ant0nsc (approved):
Looking good overall! Have a look at naming to make the method names less specific to our current use case. The rest is just my 5cents, feel free to disagree :-)

hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
kenza-bouzid: Very good point, will change it to extra!

hi-ml-cpath/src/health_cpath/utils/plots_utils.py
kenza-bouzid: The reason I made slides_dataset a param is to be able to reuse this method for the ihc_dataset as well line 258. Also Note that it's not a static method we are still using self.loading_params line 233

hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
harshita-s: I agree!

hi-ml-cpath/src/health_cpath/utils/plots_utils.py
kenza-bouzid: Very good point, my jobs all failed because of a missing slide, thanks for catching it

PR #780, opened 2023-01-31
ENH: Add ROC curve to plot options

harshita-s:
ROC curve is added as a plot option. Tests are updated for ROC and PR curves


hi-ml-cpath/src/health_cpath/utils/plots_utils.py
```diff
@@ -73,6 +73,26 @@ def save_pr_curve(results: ResultsType, figures_dir: Path, stage: str = '') -> N
...
 
+def save_roc_curve(results: ResultsType, figures_dir: Path, stage: str = '') -> None:
+    """Plots and saves ROC curve figure in its dedicated directory. This implementation
+    only works for binary classification.
+''
+    :param results: Dict of lists that contains slide_level results
+    :param figures_dir: The path to the directory where to save the figure
+    :param stage: Test or validation, used to name the figure. Empty string by default.
+    """
+    true_labels = [i.item() if isinstance(i, Tensor) else i for i in results[ResultsKey.TRUE_LABEL]]
```
ant0nsc, line 84: This fragment is repeated very often (line 86, 104, 105). Add a function for that?

ant0nsc (approved):
Looking good! Small opportunity to apply DRY (don't repeat yourself)

kenza-bouzid (approved):
LGTM

hi-ml-cpath/src/health_cpath/utils/plots_utils.py
harshita-s: Thanks for the suggestion, I have now replaced these lines with function calls

hi-ml-cpath/src/health_cpath/utils/plots_utils.py
```diff
@@ -187,6 +207,14 @@ def make_figure_dirs(subfolder: str, parent_dir: Path) -> Path:
     return figures_dir
 
 
+def get_list_from_results_dict(results: ResultsType, results_key: ResultsKey) -> List[Any]:
+    """ Get a specific results list from the slide_level results dictionary
```
kenza-bouzid, line 211: can you please mention that we extract items from tensors  here so that it's compatible with inputs formats of scikit learn functions?

harshita-s: done, thanks

PR #781, opened 2023-01-31
ENH: Enable resubmitting single cross validation runs manually

kenza-bouzid:
Add resubmit_crossval_child flag to be able to submit a single fold run. This is useful in case one of the crossvalidation folds fails and needs to be resubmitted manually.


hi-ml/src/health_ml/deep_learning_config.py
```diff
@@ -156,6 +156,12 @@ class WorkflowParams(param.Parameterized):
     crossval_index: int = param.Integer(default=0, bounds=(0, None),
                                         doc="When doing cross validation, this is the index of the current "
                                             "split. Valid values: 0 .. (crossval_count -1)")
+    resubmit_crossval_child: bool = param.Boolean(default=False,
+                                                  doc="If True, the current run is a cross validation child run that "
+                                                      "has been resubmitted manually along with the flags "
+                                                      "--crossval_count > 1 and --crossval_index=i. This is useful in "
+                                                      "case one of the crossvalidation folds fails and needs to be "
+                                                      "resubmitted manually. Default False")
```
ant0nsc, line 164: I think a solution that does not require another parameter would be better in the long run. How about you make crossval_index an optional integer, defaulting to None? If crossval_index is None, it indicates a crossval parent run, that should kick off a hyperdrive run.

ant0nsc (approved):
The change is fine in its own right, but I have some concerns about adding yet another parameter.  Generally, I find it better to express all cases with a minimum set of parameters. The more parameters there are, the more corner cases there are that need to be handled/ruled out.
I would suggest to abstract all case decisions into properties or simple functions, even if it only involves a simple comparison (like we have now for `is_crossvalidation_enabled`)

hi-ml/src/health_ml/deep_learning_config.py
kenza-bouzid: totally agree!! when I did a quick search of crossval_index use, it looked like it might have some side effects, after careful review your suggestion should work, will make the change. Thanks for the suggestion :)

harshita-s (approved):

ant0nsc (approved):

PR #777, opened 2023-01-31
STY: Add pre-commit hook to add trailing commas

fepegar:
As requested by @ant0nsc in https://github.com/microsoft/hi-ml/pull/776#discussion_r1091625644

Why is this useful? (Pasting from https://github.com/asottile/add-trailing-comma)

## multi-line method invocation style -- why?

```python
# Sample of *ideal* syntax
function_call(
    argument,
    5 ** 5,
    kwarg=foo,
)
```

- the initial paren is at the end of the line
- each argument is indented one level further than the function name
- the last parameter (unless the call contains an unpacking
  (`*args` / `**kwargs`)) has a trailing comma

This has the following benefits:

- arbitrary indentation is avoided:

    ```python
    # I hear you like 15 space indents
    # oh your function name changed? guess you get to reindent :)
    very_long_call(arg,
                   arg,
                   arg)
    ```
- adding / removing a parameter preserves `git blame` and is a minimal diff:

    ```diff
     # with no trailing commas
     x(
    -    arg
    +    arg,
    +    arg2
     )
    ```

    ```diff
     # with trailing commas
     x(
         arg,
    +    arg2,
     )
    ```

### trailing commas for function calls

```diff
 x(
     arg,
-    arg
+    arg,
 )
```

### trailing commas for function calls with unpackings

If `--py35-plus` is passed, `add-trailing-comma` will also perform the
following change:

```diff
 x(
-    *args
+    *args,
 )
 y(
-    **kwargs
+    **kwargs,
 )
```

Note that this would cause a **`SyntaxError`** in earlier python versions.

### trailing commas for tuple / list / dict / set literals

```diff
 x = [
-    1, 2, 3
+    1, 2, 3,
 ]
```

### trailing commas for function definitions

```diff
 def func(
         arg1,
-        arg2
+        arg2,
 ):
```

```diff
 async def func(
         arg1,
-        arg2
+        arg2,
 ):
```

### trailing commas for function definitions with unpacking arguments

If `--py36-plus` is passed, `add-trailing-comma` will also perform the
following change:

```diff
 def f(
-    *args
+    *args,
 ): pass


 def g(
-    **kwargs
+    **kwargs,
 ): pass


 def h(
-    *, kw=1
+    *, kw=1,
 ): pass
```

Note that this would cause a **`SyntaxError`** in earlier python versions.

### trailing commas for `from` imports

```diff
 from os import (
     path,
-    makedirs
+    makedirs,
 )
```

### trailing comma for class definitions

```diff
[...]

PR #782, opened 2023-01-31
STY: Add Black pre-commit hook

fepegar:
More info at
- #777


fepegar:
@ant0nsc @peterhessey do you think this makes sense? As much as I like Black, these changes will show up a lot in Git blame. If we don't mind that, I guess this is ready for review.

peterhessey:
> @ant0nsc @peterhessey do you think this makes sense? As much as I like Black, these changes will show up a lot in Git blame. If we don't mind that, I guess this is ready for review.

I think so yeah, haven't tested it yet but looks like [we can add commits to ignore in git blame](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html).

Just trying to iron out the kinks to currently, the pre-commit keeps getting upset

fepegar:
It wouldn't work for the blame view in GitHub, though. But maybe that's ok.

fepegar:
> Are you sure? [This article](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) implies that it should work!

I was not. Great! Thanks for sharing that. In that case I strongly support this PR.

fepegar:
@peterhessey, we might want to use `bugbear` to deal with line lengths: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

peterhessey:
> @peterhessey, we might want to use Bugbear to deal with line lengths: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

Good shout, thanks! You got any idea what's going on with the pre-commit? It seems to re-style files and push them, but then the `black` pre-commit fails again (even though it's supposedly fixed the files) after that push

fepegar:
Maybe some conflict with autopep8? I think we can remove that if we have Black anyway.

peterhessey:
Once this PR is reviewed + approved I will squash it down into 3-4 commits before merging all the commits (not squashed together) into the main branch. This is in order to preserve the ignoring of the styling commit as otherwise it will be all squashed into a new commit hash and will require an additional PR for fixing the git blame.

Makefile
```diff
@@ -77,19 +88,19 @@ call_pyright:
 pyright: conda call_pyright
 
 # run basic checks, assuming test requirements already installed
-check: flake8 mypy
+check: flake8 mypy black
```
ant0nsc, line 91: shouldn't black run before flake8? Flake8 might detect errors that black will later remove

docs/source/examples/10/sample.py
```diff
@@ -41,12 +41,12 @@ def main() -> None:
...
-                        help='Kernel type to be used in the algorithm')
-    parser.add_argument('--penalty', type=float, default=1.0,
-                        help='Penalty parameter of the error term')
-    parser.add_argument('--cross_validation_split_index', help="An index denoting which split of the dataset this"
-                                                               "run represents in k-fold cross-validation")
+    parser.add_argument('--kernel', type=str, default='linear', help='Kernel type to be used in the algorithm')
+    parser.add_argument('--penalty', type=float, default=1.0, help='Penalty parameter of the error term')
+    parser.add_argument(
+        '--cross_validation_split_index',
+        help="An index denoting which split of the dataset this" "run represents in k-fold cross-validation",
```
ant0nsc, line 48:
```suggestion
        help="An index denoting which split of the dataset this run represents in k-fold cross-validation",
```

hi-ml-azure/run_pytest.py
```diff
@@ -68,15 +68,17 @@ class RunPytestConfig(param.Parameterized):
...
+        doc="The maximum runtime that is allowed for this job in AzureML. This is given as a floating"
+        "point number with a string suffix s, m, h, d for seconds, minutes, hours, day. Examples: '3.5h', '2d'",
     )
     add_to_sys_path: str = param.String(
         default="",
-        doc="A folder name that should be added to sys.path. The folder name should be relative to repository root."
+        doc="A folder name that should be added to sys.path. The folder name should be relative to repository root.",
+    )
+    strictly_aml_v1: bool = param.Boolean(
+        default=True, doc="If True (default), use AzureML v1 SDK. If False, use " "the v2 of the SDK"
```
ant0nsc, line 80:
```suggestion
        default=True, doc="If True (default), use AzureML v1 SDK. If False, use the v2 of the SDK"
```

hi-ml-azure/src/health_azure/himl_tensorboard.py
```diff
@@ -27,14 +27,16 @@
...
 
 class HimlTensorboardConfig(azure_util.AmlRunScriptConfig):
-    log_dir: Path = param.ClassSelector(class_=Path, default=Path("outputs"), instantiate=False,
-                                        doc="Path to directory in which Tensorboard  files"
-                                            "(summarywriter and TB logs) are stored")
+    log_dir: Path = param.ClassSelector(
+        class_=Path,
+        default=Path("outputs"),
+        instantiate=False,
+        doc="Path to directory in which Tensorboard  files" "(summarywriter and TB logs) are stored",
```
ant0nsc, line 34:
```suggestion
        doc="Path to directory in which Tensorboard  files (summarywriter and TB logs) are stored",
```

ant0nsc, line 34: Can you do a quick search/replace for `" "` across the whole repo?

ant0nsc (approved):
Great stuff - really happy that you also worked out the `git blame` question! 2 small comments only

Makefile
peterhessey: In this context `black` doesn't make any changes as it is running with the flag `--diff --check`! so in theory the order shouldn't matter here

fepegar (commented):
Great work, @peterhessey! But we're missing the most important bit :D

[Code style: black](https://github.com/psf/black)

peterhessey:
> Great work, @peterhessey! But we're missing the most important bit :D
> 
> [Code style: black](https://github.com/psf/black)

Just realised I managed to lose all my documentation when rebasing a million times 😮‍💨 Will re-do the docs and add this in too!

hi-ml/Makefile
```diff
@@ -55,8 +55,12 @@ mypy:
 	mypy --install-types --show-error-codes --non-interactive --package health_ml
 	mypy --install-types --show-error-codes --non-interactive --package testhiml
 
+# run black check
+black:
+	find . -name '*.py' -exec black --line-length=120 --skip-string-normalization --check --diff {} +
```
ant0nsc, line 60: Generally, I think it would be better to configure black via a config file, so that it use used consistently also if we invoke it via vscode (file `pyproject.toml` for example, 
```
[tool.black]
line-length = 120
target-version = ['py39']
```

ant0nsc (approved):
Looking good, just one suggestion about configuring black

ant0nsc:
> Reading through [this article](https://akrabat.com/ignoring-revisions-with-git-blame/) it sounds like I need to manually configure `git blame` to take that `.git-blame-ignore-revs` into account? @peterhessey ?

Oh, I see, [this page] says that github does that automagically, great!

kenza-bouzid (approved):
Looks great! It's so satisfying 🤩

docs/source/coding_guidelines.md
```diff
@@ -35,15 +35,32 @@ Please follow the general [PEP 8](https://peps.python.org/pep-0008/) Python rule
...
 We highly recommend to run all those tools _before_ pushing the latest changes to a PR.
 If you have `make` installed, you can run both tools in one go via `make check` (from the repository root folder).
 
+### Black styling
+
+We use `black` for styling all of our code. To pass the tests carried out in `make check` (or `make black`) you can run the following command to reformat your changed file as per our guidelines:
+
+  ```shell
+  black <path_to_python_file> --line-length=120 --skip-string-normalization
+  ```
```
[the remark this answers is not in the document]
peterhessey: Have just added some more documentation on this :)

docs/source/coding_guidelines.md
```diff
@@ -50,16 +50,31 @@ If you have `make` installed, you can run both tools in one go via `make check`
...
 
   ```shell
-  black **/*.py --line-length=120 --skip-string-normalization
+  black .
   ```
 
-If you do not use this command, the pre-commit hook on your PR will attempt to do this automatically.
+#### Black VS Code Integration
+
+If you're using VS Code as your IDE, you can avoid having to call `black` every time you want to style some files by installing the [`black` formatter extension](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter) and adding the following to your vscode `settings.json` file:
```
kenza-bouzid, line 64: can you please add this to our .vscode/settings.json in all hi-ml-* submodules

PR #794, opened 2023-02-07
BUG: Fix ordering of `local_datasets`

### LINKED ISSUE #792: Local datasets order not correct
When accessing the `local_datasets` parameter of the lightning container, the list is not in the same order as expected from `create_v2_inputs/outputs`. We speculated that this is as dictionary keys are un-ordered, but [key order is preserved as of python 3.7+](https://stackoverflow.com/questions/5629023/order-of-keys-in-dictionaries-in-old-versions-of-python).

peterhessey:
Closes #792

### DIFF
--- a/hi-ml-azure/src/health_azure/himl.py
+++ b/hi-ml-azure/src/health_azure/himl.py
@@ -1089,7 +1089,10 @@ def _extract_v2_data_asset_from_env_vars(asset_num: int, asset_type_identifier: 
 
     asset_environment_variable = f"AZURE_ML_{asset_type_identifier}{asset_type_identifier}{asset_num}"
     asset_path_str = os.environ.get(asset_environment_variable)
-    assert asset_path_str is not None, f"Expected {asset_environment_variable} to be set in environment variables."
+    if asset_path_str is None:
+        raise ValueError(
+            f"Cannot find {asset_environment_variable} in environment variables, cannot retrieve data asset path."
+        )
 
     return Path(asset_path_str)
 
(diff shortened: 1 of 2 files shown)


hi-ml-azure/src/health_azure/himl.py
```diff
@@ -1078,24 +1079,19 @@ def _get_dataset_names_from_string(sys_arg: str, pattern: str) -> Path:
...
+    :return: The path to the data asset
     """
-    returned_input_datasets: List[Path] = []
-    returned_output_datasets: List[Path] = []
 
-    input_pattern_string = "AZURE_ML_INPUT_" + V2_INPUT_PATTERN + r"\d+"
-    output_pattern_string = r"AZURE_ML_OUTPUT_" + V2_OUTPUT_PATTERN + r"\d+"
+    asset_environment_variable = f"AZURE_ML_{asset_type_identifier}{asset_type_identifier}{asset_num}"
+    asset_path_str = os.environ.get(asset_environment_variable)
+    assert asset_path_str is not None, f"Expected {asset_environment_variable} to be set in environment variables."
```
ant0nsc, line 1092: Here, I would rather raise an exception. Dont' forget that asserts are not always executed.

ant0nsc (approved):
lgtm!

hi-ml-azure/src/health_azure/himl.py
peterhessey: Did not know that. Will change it now

kenza-bouzid (approved):

PR #797, opened 2023-02-08
BUG: Add flag for identity-based data access

### LINKED ISSUE #795: Defaulting to Identity-Based data access breaks cross-region compute jobs
It seems that if identity-based access is not available, the v2 API does not fall back to using account keys / SAS tokens. As such, when the submission fails to cross-region compute because of identity-based access not working, then the whole job fails instead of defaulting to account keys / SAS tokens.

A flag should be added to make the identity-based access an opt-in feature, instead defaulting to account keys / SAS tokens still.

peterhessey:
Closes #795

### DIFF
--- a/hi-ml-azure/testazure/testazure/test_himl.py
+++ b/hi-ml-azure/testazure/testazure/test_himl.py
@@ -817,7 +817,7 @@ def test_submit_run_v2(tmp_path: Path) -> None:
                 display_name=dummy_display_name,
                 distribution=MpiDistribution(process_count_per_instance=1),
                 instance_count=1,
-                identity=UserIdentityConfiguration(),
+                identity=None,
             )
 
             # job with hyperparameter sampling:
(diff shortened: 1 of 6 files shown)


hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -1167,7 +1167,7 @@ def test_invoking_hello_world_using_azureml_flag(tmp_path: Path) -> None:
     """
     Test that invoking hello_world.py with the --azureml flag will submit to AzureML and not run locally.
 
-    :param tmp_path: PyTest test fixture for temporary path.
+    :param tmp_path: Py_ test fixture for temporary path.
```
kenza-bouzid, line 1170: fix docu

hi-ml/src/health_ml/experiment_config.py
```diff
@@ -46,6 +46,11 @@ class ExperimentConfig(param.Parameterized):
                                                          "should only be used when debugging issues")
     strictly_aml_v1: bool = param.Boolean(default=False, doc="If True, use AzureML v1 SDK. If False (default), use "
                                                              "the v2 of the SDK")
+    identity_based_auth: bool = param.Boolean(
+        default=False,
+        doc="If True, use identity based authentication to access blob storage data via datastores. If False (default),"
```
kenza-bouzid, line 51: Mention that this is only supported when using v2 sdk?

kenza-bouzid (approved):

ant0nsc (approved):

PR #803, opened 2023-02-17
ENH: Background normalization transform

harshita-s:
In this PR, background normalization transform is added and corresponding tests are also added.


hi-ml/src/health_ml/utils/data_augmentations.py
```diff
@@ -72,10 +72,20 @@ def adjust_hed(self, img: torch.Tensor) -> torch.Tensor:
         return img
 
     def __call__(self, img: torch.Tensor) -> torch.Tensor:
-        if img.shape[1] != 3:
-            raise ValueError("HED jitter can only be applied to images with 3 channels (RGB).")
-
-        return self.adjust_hed(img)
+        original_shape = img.shape
+        if len(original_shape) == 3:
+            img = img.unsqueeze(0)  # add batch dimension if missing
```
kenza-bouzid, line 77: [Minor] can we add some tests for this?

kenza-bouzid (approved):
LGTM, it would be great to add a test for HEDJitted applied to tiles without batch dimension

hi-ml/src/health_ml/utils/data_augmentations.py
harshita-s: yes, added, thanks

hi-ml-cpath/src/health_cpath/models/transforms.py
```diff
@@ -312,3 +312,56 @@ def __call__(self, data: Mapping) -> Mapping:
         with elapsed_timer(message):
             out_data = self.transform(data)
         return out_data
+
+
+class NormalizeBackgroundd(MapTransform):
+    """Normalize the background of an image by deviding the image by the background intensity. The background
+    intensities are either defined by the user or are computed as the q-th percentile of the image."""
```
ant0nsc, line 319: Maybe you could add what effect this transform has? You're saying how it is computed - but what happens to pixels that have the background value after the transform is applied?

ant0nsc (approved):
Looking good! Some minor suggestions to improve docs and quality of tests.

hi-ml-cpath/src/health_cpath/models/transforms.py
harshita-s: thank you. I have now updated the docs

hi-ml-cpath/src/health_cpath/models/transforms.py
```diff
@@ -312,3 +312,56 @@ def __call__(self, data: Mapping) -> Mapping:
...
+        :param image_key: The key of the image to normalize.
+        :param background_keys: The keys of the precomputed background intensities that should be part of the batch
+        metadata. If None, the background intensities are computed using the q-th percentile of the image.
+        :param q_percentile: The q-th percentile to use to compute the background intensities. Only used if
+        background_keys are None. Use 50 for median, 0 for min, 100 for max.
+        """
+        self.image_key = image_key
+        assert q_percentile or background_keys, "Either background_keys or q_percentile must be set."
+        if background_keys:
+            assert len(background_keys) == 3, "Number of background keys must be 3 (for RGB images)."
```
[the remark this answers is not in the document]
harshita-s: Thank you for the comment, fixed now.

Issue #802, opened 2023-02-17
ImageInferenceEngine can only consume a very specific ImageModel class


corcra:
I want to use the `ImageInferenceEngine` from hi-ml-multimodal with an image model that can't cleanly be fit into the existing `ImageModel` class.

I'd like to propose something like:
- We define a more abstract class for ImageModels that _just_ requires the methods actually used by `ImageInferenceEngine` / `ImageTextInferenceEngine` (from what I can see, this is `forward`, `get_patchwise_projected_embeddings`, and the `training` attribute). The inference engine can then just expect something from this class as an input.
- The existing `ImageModel` class can subclass this, so no expected change to current behaviour.
- I can write a custom subclass for my special model and be able to use it in the inference engine.

fepegar:
Sounds good!

Issue #725, opened 2023-01-09
Packages not installed correctly on phrase grounding notebook


fepegar:
From https://github.com/microsoft/hi-ml/actions/runs/3873708628/jobs/6604951491

```python-traceback
---------------------------------------------------------------------------
Exception encountered at "In [5]":
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
/tmp/ipykernel_1821/1519321605.py in <module>
      2 from pathlib import Path
      3 
----> 4 import torch
      5 
      6 from health_multimodal.text import get_cxr_bert_inference

ModuleNotFoundError: No module named 'torch'
```

I've been able to reproduce manually on Binder. I haven't been able to figure it out. PIP says modules are installed correctly, but they don't show up when runnin `!pip list -v` and it's not possible to import them.

fepegar:
I just tried again and didn't get the error. Will try again in a few days.

fepegar:
Thanks to #817, we now know why this happens. The notebook is trying the package from a branch in a fork which doesn't exist here. Screenshot from [this run](https://github.com/microsoft/hi-ml/actions/runs/4293280709/jobs/7480762273):

![image](https://user-images.githubusercontent.com/12688084/221868439-c809b139-e7ad-4122-af05-00d19b74b407.png)

PR #806, opened 2023-02-20
ENH: ImageInferenceModel to consume more flexible image models

corcra:
Addressing #802.

Description:
I've introduced an abstract parent class ("MinimalImageModel" - open to suggestions for a better title) which the existing `ImageModel` inherits from.
The ImageInferenceEngine then just checks that the image model is of `MinimalImageModel` type.

I'm aware I haven't added any tests - I'm not sure how best to test this. Would it be weird to define a child class of `MinimalImageModel` _inside_ a test and confirm that the ImageInferenceEngine is happy?


fepegar:
[Converting to draft]

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -74,7 +75,23 @@ class ImageModelOutput():
...
+class MinimalImageModel(nn.Module):
+    """ An abstract class defining methods required by the ImageInferenceEngine """
+
+    def __init__(self, **kwargs: Any):
+        super().__init__()
+        assert hasattr(self, "training")
+
+    @abstractmethod
+    def forward(self, x: torch.Tensor) -> ImageModelOutput:
+        pass
```
fepegar, line 87:
```suggestion
        raise NotImplementedError
```

(same for other methods)

corcra:
I'm not sure why the notebook is failing given I haven't touched it 🤔

fepegar:
> I'm not sure why the notebook is failing given I haven't touched it 🤔

It's unrelated to this PR. More info here:
- https://github.com/microsoft/hi-ml/pull/808

Can you pleaser review that PR (do you think that image is appropriate?)? I tried to add you as a reviewer but wasn't able to.

fepegar:
The mysterious notebook error might be related to
- https://github.com/microsoft/hi-ml/issues/725

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -75,7 +77,23 @@ class ImageModelOutput():
...
-class ImageModel(nn.Module):
+class BaseImageModel(nn.Module, ABC):
+    """ An abstract class defining methods required by the ImageInferenceEngine """
+
+    def __init__(self, **kwargs: Any):
+        super().__init__()
+        assert hasattr(self, "training")
+
+    @abstractmethod
+    def forward(self, x: torch.Tensor) -> ImageModelOutput:
```
fepegar, line 88: Will this be compatible with a child class taking multiple images?

Also applies to the other method below.

fepegar: I guess a solution would be using `*args` and `**kwargs`, although mypy will probably complain. I suppose there are more elegant solutions, though.

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -75,7 +77,23 @@ class ImageModelOutput():
     projected_patch_embeddings: torch.Tensor
 
 
-class ImageModel(nn.Module):
+class BaseImageModel(nn.Module, ABC):
+    """ An abstract class defining methods required by the ImageInferenceEngine """
+
+    def __init__(self, **kwargs: Any):
+        super().__init__()
+        assert hasattr(self, "training")
```
fepegar, line 85: Why is this necessary?

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -75,7 +77,23 @@ class ImageModelOutput():
     projected_patch_embeddings: torch.Tensor
 
 
-class ImageModel(nn.Module):
+class BaseImageModel(nn.Module, ABC):
+    """ An abstract class defining methods required by the ImageInferenceEngine """
```
fepegar, line 81:
```suggestion
    """Abstract class for image models."""
```

hi-ml-multimodal/src/health_multimodal/image/model/model.py
fepegar: I suspect this `__init__` method can just be removed completely, but I might be wrong.

hi-ml-multimodal/src/health_multimodal/image/model/model.py
corcra: I've pushed a possible solution where we use an `ImageModelInput` class, a child class that takes >1 image could then subclass this to add new inputs. This is more explicit than using `*args/**kwargs` but it feels a bit clunky.

hi-ml-multimodal/src/health_multimodal/image/model/model.py
corcra: I put it in initially because we use the `training` attribute in the inference engine, but I now realise we can assume it has one by virtue of being a `nn.module`, so lets remove this `__init__`!

hi-ml-multimodal/src/health_multimodal/image/model/model.py
fepegar: How about just removing the method? Then children will have freedom to implement their `forward`s however they want.

fepegar:
> I'm not sure why the notebook is failing given I haven't touched it 🤔

FYI: https://github.com/microsoft/hi-ml/issues/725#issuecomment-1448184728

hi-ml-multimodal/src/health_multimodal/image/model/model.py
corcra: If the forward doesn't return something with at least some of the fields in `ImageModelOutput`, our inference engine methods will fail later.

fepegar: Sure, but the children classes could take care of that.

fepegar: Anyway, this is not terribly important and could be changed in a future PR.

fepegar: Actually, I've changed my mind. Looking at the code, this breaks the current API of the model, which is not great, and passing some custom class to the model instead of just a tensor is not very friendly or PyTorch-standard. So I think it might be better to use the *args syntax or simply delegate the creation of the `forward` method to the children classes.

hi-ml-multimodal/src/health_multimodal/image/inference_engine.py
```diff
@@ -78,7 +79,7 @@ def get_projected_global_embedding(self, image_path: Path) -> torch.Tensor:
                  where joint_feature_dim is the dimensionality of the joint latent space.
         """
         input_image, _ = self.load_and_transform_input_image(image_path, self.transform)
-        projected_img_emb = self.model.forward(input_image).projected_global_embedding
+        projected_img_emb = self.model.forward(ImageModelInput(image=input_image)).projected_global_embedding
```
Shruthi42, line 82: Could we break this up into multiple lines? Creating the image model input first then doing the forward pass, etc.

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -139,7 +158,7 @@ def get_patchwise_projected_embeddings(self, input_img: torch.Tensor, normalize:
         :returns projected_embeddings: tensor of embeddings in shape [batch, n_patches_h, n_patches_w, feature_size].
         """
         assert not self.training, "This function is only implemented for evaluation mode"
-        outputs = self.forward(input_img)
+        outputs = self.forward(ImageModelInput(image=input_img))
```
Shruthi42, line 161: Similar to the comment above, let's split this up into different lines.

hi-ml-multimodal/src/health_multimodal/image/model/__init__.py
```diff
@@ -3,13 +3,15 @@
...
 
+from .model import BaseImageModel
 from .model import ImageModel
 from .model import ResnetType
 from .model import get_biovil_resnet
 from .model import CXR_BERT_COMMIT_TAG
 from .model import BIOMED_VLP_CXR_BERT_SPECIALIZED
 
 __all__ = [
+    "BaseImageModel",
```
Shruthi42, line 14: We might also want to add `ImageModelInput` here, any code that subclasses `BaseImageModel` will need to access this class.

hi-ml-multimodal/src/health_multimodal/image/inference_engine.py
fepegar: I suggest keeping the current API so we don't break it: https://github.com/microsoft/hi-ml/pull/806#discussion_r1124704433

hi-ml-multimodal/src/health_multimodal/image/model/model.py
fepegar: Sounds good!

hi-ml-multimodal/src/health_multimodal/image/model/__init__.py
fepegar: [Resolving as we've changed method and this is now outdated]

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -112,7 +125,9 @@ def train(self, mode: bool = True) -> Any:
             self.projector.train(mode=False)
         return self
 
-    def forward(self, x: torch.Tensor) -> ImageModelOutput:
+    def forward(self, *args: Any, **kwargs: Any) -> ImageModelOutput:
+        x = kwargs["x"]
+        assert isinstance(x, torch.Tensor)
```
fepegar, line 130: Why do we need these changes?

hi-ml-multimodal/src/health_multimodal/image/inference_engine.py
fepegar: [Resolving as it's now outdated]

hi-ml-multimodal/src/health_multimodal/image/model/model.py
fepegar: [Resolving as it's now outdated]

hi-ml-multimodal/src/health_multimodal/image/model/model.py
fepegar: [Discussed offline]

corcra: Discussed offline - we don't, let's simplify :)

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -125,9 +125,7 @@ def train(self, mode: bool = True) -> Any:
             self.projector.train(mode=False)
         return self
 
-    def forward(self, *args: Any, **kwargs: Any) -> ImageModelOutput:
-        x = kwargs["x"]
-        assert isinstance(x, torch.Tensor)
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> ImageModelOutput:  # type: ignore
```
fepegar, line 128: Is the `x` to enforce that at least one tensor is in the `args`?

fepegar: Could you please add the specific mypy error that is being ignored? E.g. `# type: ignore[arg-type]` (not sure about the exact error code you're getting, which I guess is the point of adding it. It's safer than a blanket `# type: ignore`.

fepegar (changes requested):

fepegar (approved):

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -112,7 +125,7 @@ def train(self, mode: bool = True) -> Any:
             self.projector.train(mode=False)
         return self
 
-    def forward(self, x: torch.Tensor) -> ImageModelOutput:
+    def forward(self, x: torch.Tensor, *args: Any, **kwargs: Any) -> ImageModelOutput:  # type: ignore[override]
```
Shruthi42, line 128: Why are we allowing and then ignoring potential args and kwargs here? If someone passes in two tensors to this function, there's no indication that this is not supported - it's very possible this fails silently completely unknown to the user.

We are designing this code to allow users to use classes that inherit from `BaseImageModel` interchangeably, so there's a reasonable chance that a user swaps out one class with another and the model forward pass is run incorrectly but silently.

fepegar: I agree. I thought we were ignoring mypy errors for the base class, not for this.

corcra: I've removed the unused args/kwargs now. The mypy issue occurs in the child class:
`model.py:128: error: Signature of "forward" incompatible with supertype "BaseImageModel"` 
... so I think we need to ignore it here. This might be a bit annoying as I guess all children of `BaseImageModel` will need to do something similar to appease mypy.

Shruthi42 (approved):

fepegar:
It looks like this PR will be superseded by
- #830 

which has diverged anyway. Shall we close this / put it on hold?

corcra:
> It looks like this PR will be superseded by
> 
> * [ENH: Add BioViL-T image and text encoders #830](https://github.com/microsoft/hi-ml/pull/830)
> 
> which has diverged anyway. Shall we close this / put it on hold?

Yes, sounds good. I'll close this in favour of #830.

PR #807, opened 2023-02-20
ENH: Improve support for choosing image model weights

### LINKED ISSUE #617: Stop hardcoding `pretrained` kwarg for image model
```python
import torch
github = 'microsoft/hi-ml:main'
model = torch.hub.load(github, 'biovil_resnet', pretrained=False)
```

I would expect `model` to contain randomly-initialized weights, but in

https://github.com/microsoft/hi-ml/blob/681e4281d0439dc554a410924b1fd5878cc31360/hi-ml-multimodal/src/health_multimodal/image/model/model.py#L165

we enforce ImageNet weights.

The easiest solution is to remove hard-coded argument:

```python
encoder = encoder_class(**kwargs) 
```

A nicer solution might be creating a new parameter, maybe `weights`, that can be, e.g., `random`, `imagenet` or `biovil`. I think TorchVision now uses a similar syntax:

```python
resnet50(weights=ResNet50_Weights.IMAGENET1K_V2)
```

@Shruthi42 @ozan-oktay

markpinnock:
This PR allows the user to determine whether to use ResNet weights pre-trained on ImageNet or BioViL, or whether to have a randomly initialised model: fixes #617  .

### DIFF
--- a/hi-ml-multimodal/src/health_multimodal/image/model/pretrained.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/model/pretrained.py
@@ -111,7 +111,7 @@ def get_image_encoder(weights: str) -> ImageModel:
             img_encoder_type=ImageEncoderType.RESNET50,
             joint_feature_size=JOINT_FEATURE_SIZE,
             pretrained_model_path=None,
-    )
+        )
     elif weights == Weights.IMAGENET:
         image_model = get_imagenet_init_encoder()
     elif weights == Weights.BIOVIL:
--- a/hi-ml-multimodal/src/health_multimodal/image/model/pretrained.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/model/pretrained.py
@@ -99,3 +99,26 @@ def get_imagenet_init_encoder() -> ImageModel:
     )
     image_model.encoder.encoder.load_state_dict(state_dict)
     return image_model
+
+
+def get_image_encoder(weights: str) -> ImageModel:
+    """Instantiate image model with random or pre-trained weights.
+    :param weights: Select one of `random`, `imagenet`, `biovil`, `biovil_t`
+    """
+
+    if weights == Weights.RANDOM:
+        image_model = ImageModel(
+            img_encoder_type=ImageEncoderType.RESNET50,
+            joint_feature_size=JOINT_FEATURE_SIZE,
+            pretrained_model_path=None,
+    )
+    elif weights == Weights.IMAGENET:
+        image_model = get_imagenet_init_encoder()
+    elif weights == Weights.BIOVIL:
+        image_model = get_biovil_image_encoder()
+    elif weights == Weights.BIOVIL_T:
+        image_model = get_biovil_t_image_encoder()
+    else:
+        raise ValueError(f"Weights option not found: {weights}")
+
+    return image_model
--- a/hi-ml-multimodal/src/health_multimodal/image/model/pretrained.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/model/pretrained.py
@@ -8,10 +8,12 @@ from __future__ import annotations
 import tempfile
 from pathlib import Path
 
+from torch.hub import load_state_dict_from_url
 from torchvision.datasets.utils import download_url
+from torchvision.models.resnet import model_urls
 
 from .model import ImageModel
-from .types import ImageEncoderType
+from .types import ImageEncoderType, Weights
 
 
 JOINT_FEATURE_SIZE = 128
@@ -83,4 +85,17 @@ def get_biovil_t_image_encoder() -> ImageModel:
         joint_feature_size=JOINT_FEATURE_SIZE,
         pretrained_model_path=biovilt_checkpoint_path,
     )
+    return image_model
+
+
+def get_imagenet_init_encoder() -> ImageModel:
+    """Download ImageNet pre-trained weights and instantiate the image model."""
+
+    state_dict = load_state_dict_from_url(model_urls[ImageEncoderType.RESNET50])
+    image_model = ImageModel(
+        img_encoder_type=ImageEncoderType.RESNET50,
+        joint_feature_size=JOINT_FEATURE_SIZE,
+        pretrained_model_path=None,
+    )
+    image_model.encoder.encoder.load_state_dict(state_dict)
     return image_model
--- a/hi-ml-multimodal/src/health_multimodal/image/model/pretrained.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/model/pretrained.py
@@ -13,7 +13,7 @@ from torchvision.datasets.utils import download_url
 from torchvision.models.resnet import model_urls
 
 from .model import ImageModel
-from .types import ImageEncoderType, Weights
+from .types import ImageEncoderType, ImageEncoderWeightTypes
 
 
 JOINT_FEATURE_SIZE = 128
@@ -106,17 +106,17 @@ def get_image_encoder(weights: str) -> ImageModel:
     :param weights: Select one of `random`, `imagenet`, `biovil`, `biovil_t`
     """
 
-    if weights == Weights.RANDOM:
+    if weights == ImageEncoderWeightTypes.RANDOM:
         image_model = ImageModel(
             img_encoder_type=ImageEncoderType.RESNET50,
             joint_feature_size=JOINT_FEATURE_SIZE,
             pretrained_model_path=None,
         )
-    elif weights == Weights.IMAGENET:
+    elif weights == ImageEncoderWeightTypes.IMAGENET:
         image_model = get_imagenet_init_encoder()
-    elif weights == Weights.BIOVIL:
+    elif weights == ImageEncoderWeightTypes.BIOVIL:
         image_model = get_biovil_image_encoder()
-    elif weights == Weights.BIOVIL_T:
+    elif weights == ImageEncoderWeightTypes.BIOVIL_T:
         image_model = get_biovil_t_image_encoder()
     else:
         raise ValueError(f"Weights option not found: {weights}")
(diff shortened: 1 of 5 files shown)


hi-ml-multimodal/test_multimodal/image/model/test_model.py
```diff
@@ -156,3 +156,92 @@ def test_hubconf() -> None:
...
+    biovil_params = biovil_model.encoder.state_dict()
+    expected_params = expected_model.encoder.state_dict()
+
+    for (k, v), (exp_k, exp_v) in zip(
+        biovil_params.items(), expected_params.items()
+    ):
+        assert k == exp_k
+
+        # BN running means are different
+        if 'running' not in k:
```
fepegar, line 179: Why are they different?

hi-ml-multimodal/test_multimodal/image/model/test_model.py
```diff
@@ -156,3 +156,92 @@ def test_hubconf() -> None:
...
+    biovil_model = ImageModel(img_model_type="resnet50",
+                              joint_feature_size=joint_feature_size,
+                              weights="biovil")
+
+    biovil_params = biovil_model.encoder.state_dict()
+    expected_params = expected_model.encoder.state_dict()
+
+    for (k, v), (exp_k, exp_v) in zip(
+        biovil_params.items(), expected_params.items()
+    ):
```
fepegar, line 175: Can you use `biovil_model.named_parameters()` instead?

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -82,11 +82,12 @@ def __init__(self,
                  joint_feature_size: int,
                  freeze_encoder: bool = False,
                  pretrained_model_path: Optional[Union[str, Path]] = None,
+                 weights: Weights = Weights.IMAGENET,
                  **downstream_classifier_kwargs: Any):
         super().__init__()
 
         # Initiate encoder, projector, and classifier
-        self.encoder = ImageEncoder(img_model_type)
+        self.encoder = ImageEncoder(img_model_type, weights)
```
fepegar, line 90:
```suggestion
        self.encoder = ImageEncoder(img_model_type, weights=weights)
```

hi-ml-multimodal/test_multimodal/image/model/test_model.py
fepegar: [Same for tests below]

hi-ml-multimodal/test_multimodal/image/model/test_model.py
markpinnock: So this actually doesn't apply for the BioViL version as all parameters are over-written - I have removed that from the BioViL unit test. However, it does occur when initialising an ImageNet-pretrained ImageModel - I had noticed the same behaviour in the unchanged repo previously but didn't look into it. Specifically it is running/tracked parameters that are affected.

I've compared the ImageModel.encoder parameters with a model from "resnet50()" during various stages of initialisation and they diverge in get_encoder_output_dim(self.encoder) (line 90), specifically at line 215. Basically what I think is happening is that the parameters are identical until self.encoder is called on the random tensor in this function - the running/tracked params then get updated, but all weights and non-tracked params stay the same as no training has occurred.

I tried moving self.freeze_encoder/self.train() (lines 97/98) up above line 90, but that caused all sorts of other failures :(

fepegar: Thanks for the investigation. I'll open a PR that will hopefully fic this issue and then we can update this branch.

hi-ml-multimodal/test_multimodal/image/model/test_model.py
markpinnock: Ok, will do

hi-ml-multimodal/test_multimodal/image/model/test_model.py
fepegar: PR opened:
- https://github.com/microsoft/hi-ml/pull/809

fepegar: I've merged the changes to this branch. Can you please see if we can safely remove those exceptional checks?

hi-ml-multimodal/src/health_multimodal/image/model/resnet.py
```diff
@@ -48,33 +56,39 @@ def forward(self, x: torch.Tensor,
...
 
 
-def resnet18(pretrained: bool = False, progress: bool = True, **kwargs: Any) -> ResNetHIML:
+def resnet18(pretrained: bool = False, progress: bool = True,
+             weights: Weights = Weights.RANDOM, **kwargs: Any) -> ResNetHIML:
     r"""ResNet-18 model from
     `"Deep Residual Learning for Image Recognition" <https://arxiv.org/pdf/1512.03385.pdf>`_.
 
     :param pretrained: If ``True``, returns a model pre-trained on ImageNet.
+    :param weights: Returns model pre-trained on ``biovil``, ``imagenet`` or with ``random`` weights.
```
fepegar, line 77: Should come after `progress`.

fepegar: Same below.

hi-ml-multimodal/test_multimodal/image/model/test_model.py
```diff
@@ -159,6 +159,81 @@ def test_hubconf() -> None:
...
+                                joint_feature_size=joint_feature_size,
+                                weights="random")
+
+    for random_param_1, random_param_2 in zip(
+        random_model_1.encoder.encoder.named_parameters(),
+        random_model_2.encoder.encoder.named_parameters()
+    ):
+        assert random_param_1[0] == random_param_2[0]
+
+        # BN and down_sample initialisations are the same in random models
```
fepegar, line 232: Do you know why they are the same?

markpinnock: I assumed batchnorm was the same as in Tensorflow, where beta and gamma are initialised to 0 and 1 respectively, but I'll check

fepegar: Ah that would make sense! Thanks, I was curious.

fepegar:
Converting to draft until we've updated from `main`, after merging the high-priority PR
- #830

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -98,6 +99,9 @@ def __init__(self,
         self.freeze_encoder = freeze_encoder
         self.train()
 
+        if weights == Weights.BIOVIL and pretrained_model_path is None:
```
Shruthi42, line 102: `get_biovil_resnet` from `image.model` does exactly this - why not use that function to load BioViL instead of adding logic to this class?

hi-ml-multimodal/src/health_multimodal/image/model/resnet.py
```diff
@@ -48,33 +56,39 @@ def forward(self, x: torch.Tensor,
 
 
 def _resnet(arch: str, block: Type[Union[BasicBlock, Bottleneck]], layers: List[int],
-            pretrained: bool, progress: bool, **kwargs: Any) -> ResNetHIML:
+            pretrained: bool, weights: Weights, progress: bool, **kwargs: Any) -> ResNetHIML:
```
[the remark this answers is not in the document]
markpinnock: Hi @Shruthi42 - thank you for reviewing. The issue is that the `ImageEncoder` class has `pretrained=True` hard-coded when calling either `resnet18` or `resnet50`:
https://github.com/microsoft/hi-ml/blob/273e84e6bbf777dc93669c88c2200111fd7fcd48/hi-ml-multimodal/src/health_multimodal/image/model/model.py#L165-L166
This means that we are not able to use a randomly initialised `ImageModel` without then adding an extra method to iterate over the weights and re-initialise (as the default is ImageNet). The idea was that this addition could allow this to be an option while still being compatible with previous versions.

hi-ml-multimodal/src/health_multimodal/image/model/model.py
markpinnock: We are currently using the `ImageModel` class directly, as this makes selecting random weights easier just by passing an argument directly from the config file to `ImageModel` when initialising. The alternative was calling `get_biovil_resnet` and then having to call an extra method to randomly re-initialise the model weights, which felt a bit messy...

ozan-oktay:
@markpinnock and @fepegar: Thanks for creating the PR and proposing this feature. I would like to recommend that nn.Module class definition and state dictionary initialisation are conceptually different topics, and it's conventionally handled separately in DL libraries. Moving forward we probably should remove the `pretrained_model_path` argument from `ImageModel` class instead of adding more argument because "weights or bnorm stats" do not characterise how computation graph is defined.

For your specific use case, I would recommend writing a simple helper function like the one below, which should enable your use-case and help us to have a more modular design. In case, I am missing a feature that you intend to have, please reply to this thread so that we can identify an alternative solution. Thanks.

```
def load_pretrained_encoder(specified_state_dict_path, model_args_dict):
     my_module = ImageModel(**model_args_dict)
     state_dict = torch.load(specified_state_dict_path)
     my_module.encoder.load_state_dict(state_dict)
     return my_module
```

hi-ml-multimodal/src/health_multimodal/image/model/pretrained.py
```diff
@@ -84,3 +86,39 @@ def get_biovil_t_image_encoder() -> ImageModel:
         pretrained_model_path=biovilt_checkpoint_path,
     )
     return image_model
+
+
+def get_imagenet_init_encoder() -> ImageModel:
```
ozan-oktay, line 91: Would you mind adding a simple unit-test checking if this function returns the expected image encoder?

ozan-oktay (approved):

fepegar (approved):

fepegar:
Thanks for your contribution, @markpinnock!

PR #809, opened 2023-02-20
FIX: Set evaluation mode to compute number of dimensions

fepegar:
When guessing the number of output dimensions in the image encoder, the model is in training mode and the batch norm stats get updated. This shouldn't really happen and makes the checks in https://github.com/microsoft/hi-ml/pull/807/files#r1112200449 fail. This PR adds a line to ensure the encode is in `eval` mode before calculating the number of output dims.

### DIFF
--- a/hi-ml-multimodal/src/health_multimodal/image/model/model.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/model/model.py
@@ -87,6 +87,7 @@ class ImageModel(nn.Module):
 
         # Initiate encoder, projector, and classifier
         self.encoder = ImageEncoder(img_model_type)
+        self.encoder.eval()  # to get the output size in the next line
         self.feature_size = get_encoder_output_dim(self.encoder)
         self.projector = MLP(input_dim=self.feature_size, output_dim=joint_feature_size,
                              hidden_dim=joint_feature_size, use_1x1_convs=True)
--- a/hi-ml-multimodal/src/health_multimodal/image/model/model.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/model/model.py
@@ -88,7 +88,6 @@ class ImageModel(nn.Module):
 
         # Initiate encoder, projector, and classifier
         self.encoder = ImageEncoder(img_model_type)
-        self.encoder.eval()  # to get the output size in the next line
         self.feature_size = get_encoder_output_dim(self.encoder)
         self.projector = MLP(input_dim=self.feature_size, output_dim=joint_feature_size,
                              hidden_dim=joint_feature_size, use_1x1_convs=True)
--- a/hi-ml-multimodal/src/health_multimodal/image/model/model.py
+++ b/hi-ml-multimodal/src/health_multimodal/image/model/model.py
@@ -8,8 +8,9 @@ from __future__ import annotations
 import enum
 import tempfile
 from pathlib import Path
+from contextlib import contextmanager
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -87,6 +87,7 @@ def __init__(self,
 
         # Initiate encoder, projector, and classifier
         self.encoder = ImageEncoder(img_model_type)
+        self.encoder.eval()  # to get the output size in the next line
```
ozan-oktay, line 90: By convention `nn.modules` are initialised with `self.training` set to `True`.

Instead of hardcoding this here, perhaps we could handle this inside the `get_encoder_output_dim` method also in case other classes use this function later.

One idea is that -- In line 214, you could add the following:
```
        initial_state = self.encoder.training
        module.eval()
        module.forward(...)
        module.train(initial_state) # Restore the state of encoder again
```

fepegar: Thanks, I thought of doing that, potentially with a context manager. But I think in this case it shouldn't be necessary as we have `self.train()` in the `__init__` of the class, just a few lines below: https://github.com/microsoft/hi-ml/pull/809/files#diff-ddbcf0079364b4612521f029a11af075c8293ec9c06f147e9cedafa17f6c42bdR99

fepegar: I've created a context manager in [4bdc2e7](https://github.com/microsoft/hi-ml/pull/809/commits/4bdc2e7bf8cd61bb53431c827955679229e21b4c) to handle this in a generic way.

Shruthi42: Looks good to me, if we remove this line and use the context manager you created instead.

Shruthi42 (approved):

hi-ml-multimodal/src/health_multimodal/image/model/model.py
```diff
@@ -212,5 +214,18 @@ def get_encoder_output_dim(module: torch.nn.Module) -> int:
...
     # Extract the number of output feature dimensions
-    representations = module(x)
+    with restore_training_mode(module):
+        module.eval()
+        representations = module(x)
     return representations.shape[1]
+
+
+@contextmanager
+def restore_training_mode(module: nn.Module) -> Generator[None, None, None]:
```
Shruthi42, line 224: @fepegar This method could be useful in other codebases. Wondering if it should be placed elsewhere, instead of in image/model.

fepegar: It should, I agree. But the import from `hubconf.py` is a bit brittle as it's designed so only `torch` and `torchvision` are needed, so I'd rather work on that on a different PR so we can unblock @markpinnock's PR (https://github.com/microsoft/hi-ml/pull/807). Does that make sense?

hi-ml-multimodal/src/health_multimodal/image/model/model.py
fepegar: Thanks! I forgot to remove it. Done in 9533bead.

ozan-oktay (approved):
👍

hi-ml-multimodal/src/health_multimodal/image/model/model.py
Shruthi42: I see, this sounds good.

PR #810, opened 2023-02-21
ENH: Add Swin Transformer Encoder and Projection layer

kenza-bouzid:
Add Swin transformer encoder and an intermediate projection layer


ant0nsc (approved):

hi-ml-cpath/src/health_cpath/models/encoders.py
```diff
@@ -105,6 +106,15 @@ def __init__(self, tile_size: int, n_channels: int = 3) -> None:
         super().__init__(resnet50, tile_size, n_channels, apply_imagenet_preprocessing=False)
 
 
+class SwinTransformer_NoPreproc(ImageNetEncoder):
+    def __init__(self, tile_size: int, n_channels: int = 3) -> None:
```
harshita-s, line 110: please add some docstrings explaining this encoder

hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
```diff
@@ -89,6 +93,9 @@ def get_encoder(self, outputs_folder: Optional[Path]) -> TileEncoder:
         elif self.encoder_type == Resnet50_NoPreproc.__name__:
             encoder = Resnet50_NoPreproc(tile_size=self.tile_size, n_channels=self.n_channels)
 
+        elif self.encoder_type == SwinTransformer_NoPreproc.__name__:
+            encoder = SwinTransformer_NoPreproc(tile_size=self.tile_size, n_channels=self.n_channels)
```
harshita-s, line 97: can we cover this encoder in a test?

hi-ml-cpath/src/health_cpath/models/encoders.py
```diff
@@ -105,6 +106,15 @@ def __init__(self, tile_size: int, n_channels: int = 3) -> None:
...
+        super().__init__(swin_tiny_patch4_window7_224, tile_size, n_channels, apply_imagenet_preprocessing=False)
+
+    def _get_encoder(self) -> Tuple[torch.nn.Module, int]:
+        pretrained_model = self.create_feature_extractor_fn(pretrained=True, num_classes=0)
+        return pretrained_model, pretrained_model.num_features  # type: ignore
```
harshita-s, line 115: why do we need to ignore type here?

hi-ml-cpath/src/health_cpath/models/deepmil.py
```diff
@@ -89,7 +89,9 @@ def __init__(self,
 
         # Model components
         self.encoder = encoder_params.get_encoder(outputs_folder)
-        self.aggregation_fn, self.num_pooling = pooling_params.get_pooling_layer(self.encoder.num_encoding)
+        self.projector = encoder_params.get_projection_layer(self.encoder.num_encoding)
+        num_encoding = encoder_params.projection_dim if encoder_params.projection_dim > 0 else self.encoder.num_encoding
```
harshita-s, line 93: could you explain why this is needed in a comment, perhaps

hi-ml-cpath/src/health_cpath/models/encoders.py
kenza-bouzid: it doesn't recognize num_features

hi-ml-cpath/src/health_cpath/models/deepmil.py
kenza-bouzid: If the projection_dim is 0 than the projection layer is an Identity layer so num_encoding won't change, however, if encoder_params.projection_dim > 0, we reduce the encoding dimension so we need to use encoder_params.projection_dim

hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
kenza-bouzid: We don't currently have tests that cover any of the encoders, we can maybe address it in a separate PR, but anyways I'm not sure we need to test these as it's external code, we're just import predefined encoders from libraries that already test the encoders expected behaviour

hi-ml-cpath/src/health_cpath/models/encoders.py
kenza-bouzid: I think the name is self explanatory, we don't have docs for Resnet encoders either

hi-ml-cpath/src/health_cpath/models/deepmil.py
harshita-s: could you please add it as a comment (what I implied above)

hi-ml-cpath/src/health_cpath/models/encoders.py
harshita-s: I think we should explain what version this is `swin_tiny_patch4_window7_224` what does 4, 7, tiny refer to?

harshita-s: Resnets are self explanatory (with suffix 18 and 50) but there are multiple versions of swin transformers. I think the name should be `SwinT_Transformer` or something similar

hi-ml-cpath/src/health_cpath/models/encoders.py
harshita-s: if num_features are int, we shouldn't need it, don't understand the problem?

hi-ml-cpath/src/health_cpath/utils/deepmil_utils.py
harshita-s: ok, thanks, codecov complains so I was wondering if we should add some tests

harshita-s (approved):

hi-ml-cpath/src/health_cpath/models/encoders.py
kenza-bouzid: Agreed, patch 4 mean that we using a patch size 4x4 for patching the input images into small tokens window 7 is the size of the sliding window 7x7 within which attention is computed, 224 is the input image size, tiny is for the model's size.
I'll add this in docs. Thanks

PR #824, opened 2023-03-02
DOC: Use symlinks to build docs instead of copying files

### LINKED ISSUE #823: Missing sections in the documentation
Three pages are missing from the [documentation](https://hi-ml.readthedocs.io/en/latest/):

- Main page for hi-ml-multimodal
- Contributing
- Changelog

This happens because they are not found by Read the Docs. It works locally because they are copied into the `docs` folder using `make html`, but that command is not used when building on RTD. There are warnings in the docs building process, but we ignore them.

Compare 

https://github.com/microsoft/hi-ml/blob/687fa69592c37aea3baf032342583ed1ed9ba5c8/docs/source/index.rst?plain=1#L60-L100

with 

image

fepegar:
Fixes #823.

There was an issue with the Azure ML Python SDK v2 disliking symlinks in the source dir, but that seems to have been resolved now so this PR removes the copying operation and replaces the targets with symlinks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -145,9 +145,6 @@ tensorboard_logs/
 docs/source/api
 !docs/source/api/api.rst
 !docs/source/api/multimodal.rst
-# This file is copied from repository root to docs/source by the makefile
-/docs/source/CHANGELOG.md
-/docs/source/CONTRIBUTING.md
 
 node_modules/
 !.github/actions/format_coverage/dist/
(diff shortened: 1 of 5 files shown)


fepegar:
The missing sections are back, so these changes worked.

image

peterhessey (approved):
LGTM

ant0nsc (approved):
Great! I verified that the repo now does not cause any problems with AzureML v2 SDK job submissions.

fepegar:
Thank you both!

PR #828, opened 2023-03-03
ENH: Add LoadMaskSubROId transform

kenza-bouzid:
LoadMaskSubROId loads a sub section from a slide based a foreground png mask with categorical labels for each tissue section.

### DIFF
--- a/hi-ml-cpath/src/health_cpath/preprocessing/loading.py
+++ b/hi-ml-cpath/src/health_cpath/preprocessing/loading.py
@@ -224,7 +224,7 @@ class LoadMaskSubROId(MapTransform, RandomizableTransform, BaseLoadROId):
     Operates on dictionaries, replacing the file paths in `image_key` loaded array, in (C, H, W) format. Also adds the
     following meta-data entries:
     - `'location'` (tuple): top-right coordinates of the bounding box
-    - `'size'` (tuple): width and height of the bounding box
+    - `'size'` (tuple): A tuple with (width, height) of the bounding box
     - `'level'` (int): chosen magnification level
     - `'scale'` (float): corresponding scale, loaded from the file
     """
--- a/hi-ml-cpath/src/health_cpath/preprocessing/loading.py
+++ b/hi-ml-cpath/src/health_cpath/preprocessing/loading.py
@@ -158,8 +158,8 @@ class LoadROId(MapTransform, BaseLoadROId):
 
 
 class LoadMaskROId(MapTransform, BaseLoadROId):
-    """Transform that loads a pathology slide and mask, cropped to the mask bounding box (ROI) defined by the mask.
-
+    """Operates on dictionaries, replacing the file paths in `image_key` and `mask_key` with the
+    respective loaded arrays, in (C, H, W) format. Also adds the following meta-data entries:
     Operates on dictionaries, replacing the file paths in `image_key` loaded array, in (C, H, W) format. Also adds the
     following meta-data entries:
     - `'location'` (tuple): top-right coordinates of the bounding box
--- a/hi-ml-cpath/src/health_cpath/preprocessing/loading.py
+++ b/hi-ml-cpath/src/health_cpath/preprocessing/loading.py
@@ -2,6 +2,7 @@
 #  Copyright (c) Microsoft Corporation. All rights reserved.
 #  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 #  ------------------------------------------------------------------------------------------
+from git import Sequence
 import param
 import numpy as np
 import skimage.filters
@@ -10,8 +11,9 @@ from enum import Enum
 from health_ml.utils import box_utils
 from health_cpath.utils.naming import SlideKey
 from monai.data.wsi_reader import WSIReader
-from monai.transforms import MapTransform, LoadImaged
-from typing import Any, Callable, Dict, Optional, Tuple
+from monai.transforms import MapTransform, LoadImaged, RandomizableTransform
+from PIL import Image
+from typing import Any, Callable, Dict, Optional, Tuple, Union
 from health_azure.logging import print_message_with_rank_pid
 
 
@@ -41,23 +43,30 @@ def segment_foreground(slide: np.ndarray, threshold: Optional[float] = None) -> 
 
 class ROIType(str, Enum):
     """Options for the ROI selection. Either a bounding box defined by foreground or a mask can be used."""
-    FOREGROUND = 'foreground'
-    MASK = 'segmentation_mask'
-    WHOLE = 'whole_slide'
+
+    FOREGROUND = "foreground"
+    MASK = "segmentation_mask"
+    WHOLE = "whole_slide"
+    MASKSUBROI = "sub_section"
 
 
 class WSIBackend(str, Enum):
     """Options for the WSI reader backend."""
-    OPENSLIDE = 'OpenSlide'
-    CUCIM = 'cuCIM'
+
+    OPENSLIDE = "OpenSlide"
+    CUCIM = "cuCIM"
 
 
 class BaseLoadROId:
     """Abstract base class for loading a region of interest (ROI) from a slide. The ROI is defined by a bounding box."""
 
     def __init__(
-        self, backend: str = WSIBackend.CUCIM, image_key: str = SlideKey.IMAGE, level: int = 1, margin: int = 0,
-        backend_args: Dict = {}
+        self,
+        backend: str = WSIBackend.CUCIM,
+        image_key: str = SlideKey.IMAGE,
+        level: int = 1,
+        margin: int = 0,
+        backend_args: Dict = {},
     ) -> None:
         """
         :param backend: The WSI reader backend to use. One of 'OpenSlide' or 'cuCIM'. Default: 'cuCIM'.
@@ -81,7 +90,7 @@ class BaseLoadROId:
         h, w = self.reader.get_size(slide_obj, level=level)
         return box_utils.Box(0, 0, w, h)
 
-    def _get_bounding_box(self, slide_obj: Any, slide_id: int) -> box_utils.Box:
+    def _get_bounding_box(self, slide_obj: Any, slide_id: Union[str, int]) -> box_utils.Box:
         """Estimate bounding box at the lowest resolution (i.e. highest level) of the slide."""
         highest_level = self.reader.get_level_count(slide_obj) - 1
         scale = self.reader.get_downsample_ratio(slide_obj, highest_level)
@@ -138,8 +147,9 @@ class LoadROId(MapTransform, BaseLoadROId):
             scale = self.reader.get_downsample_ratio(image_obj, self.level)
             scaled_bbox = level0_bbox / scale
 
-            data[self.image_key], _ = self.reader.get_data(image_obj, location=origin, level=self.level,
-                                                           size=(scaled_bbox.h, scaled_bbox.w))
+            data[self.image_key], _ = self.reader.get_data(
+                image_obj, location=origin, level=self.level, size=(scaled_bbox.h, scaled_bbox.w)
+            )
             data[SlideKey.ORIGIN] = origin
             data[SlideKey.SCALE] = scale
             data[SlideKey.FOREGROUND_THRESHOLD] = self.foreground_threshold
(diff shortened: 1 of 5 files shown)


hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -192,8 +202,6 @@ def __call__(self, data: Dict) -> Dict:
                 size=(scaled_bbox.h, scaled_bbox.w),
                 level=self.level,
             )
-            mask, _ = self.reader.get_data(mask_obj, **get_data_kwargs)  # type: ignore
```
kenza-bouzid, line 195: removed mask to free up memory, no neeed to store the mask data, it's like storing a secon wsi

harshita-s: Not sure why this change was made. the name of the transform `LoadMaskROId` suggests that mask will be loaded. A mask is much lighter than loading WSI. If there is no mask, this won't work right?

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -204,38 +212,127 @@ def __call__(self, data: Dict) -> Dict:
...
+
+    def _get_foreground_mask(self, mask_path: str, level: int = 0) -> np.ndarray:
+        """Return a binary foreground mask using the specified label or a random label if roi_label is None."""
+        mask = np.asarray(Image.open(mask_path))[..., 0]
+        section = self._get_sub_section_label(mask)
+        foreground_mask = mask == section
+        return foreground_mask
+
+    def _get_bounding_box(self, foreground_mask: np.ndarray, slide_id: Union[str, int] = "") -> box_utils.Box:
+        """Estimate bounding box at the lowest resolution (i.e. highest level) of the slide."""
```
harshita-s, line 270: do you mean at level 0? Isn't this highest resolution?

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -204,38 +212,127 @@ def __call__(self, data: Dict) -> Dict:
...
+        wsi_mag_at_level: float = 10.0,
+        mask_mag: float = 1.25,
+        roi_label: Optional[Sequence[int]] = None,
+        **kwargs: Any,
+    ) -> None:
+        """
+        :param image_key: Image key in the input and output dictionaries. Default: 'image'.
+        :param mask_key: Mask key in the input and output dictionaries. Default: 'mask'.
+        :param wsi_mag_at_level0: WSI magnification at level 0. Default: 10x.
+        :param wsi_mag_at_level: WSI magnification at the chosen level. Default: 10x.
```
harshita-s, line 244: why are two different magnifications required here?

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
kenza-bouzid: Oh sorry I didn't update the docs when I copied the function definition from above, will do!

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
kenza-bouzid: Good point we only need wsi_mag_at_level0, I can infer the other from the wsi_obj, thanks, changed it now!

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
kenza-bouzid: Oh in this setting the mask is as large as the whole WSI, remember that these mask are tiffiles and here in this line we load them at the the given level. I don't think it matters much for PANDA for example because the slides are relatively smaller but in the case of very large size (25k, 25k) storing the masks affect the memory consumption quite a lot which prevent us from using all cpu cores for preprocessing

kenza-bouzid: Also I read LoadMaskRoid as loading a masked region of interest but not the mask necessarily, do you see any use case where will need the mask to be loaded?

harshita-s: Shouldn't we load the mask for compatibility with PANDA experiments?

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -204,38 +214,124 @@ def __call__(self, data: Dict) -> Dict:
...
+    - `'scale'` (float): corresponding scale, loaded from the file
+    """
+
+    def __init__(
+        self,
+        image_key: str = SlideKey.IMAGE,
+        mask_key: str = SlideKey.MASK,
+        wsi_mag_at_level0: float = 10.0,
+        mask_mag: float = 1.25,
+        roi_label: Optional[Sequence[int]] = None,
```
ant0nsc, line 238: Type annotations for roi_label and usage are not consistent

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -204,38 +214,124 @@ def __call__(self, data: Dict) -> Dict:
...
+            random label is sampled from the foreground tissue uniformly.
+        :param kwargs: Additional arguments for `BaseLoadROId`.
+        """
+        MapTransform.__init__(self, [image_key, mask_key], allow_missing_keys=False)
+        BaseLoadROId.__init__(self, image_key=image_key, **kwargs)
+        self.mask_key = mask_key
+        self.scale_level0 = wsi_mag_at_level0 / mask_mag
+        self.roi_label = roi_label
+
+    def _get_sub_section_label(self, mask: np.ndarray) -> int:
```
ant0nsc, line 256: this should be `Sequence[int`?

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -204,38 +214,124 @@ def __call__(self, data: Dict) -> Dict:
...
+        """Return the label of the sub section to crop. If roi_label is None, a random label is sampled from the
+        foreground tissue uniformly. Otherwise the sub ROI is sampled from the foreground of the specified label."""
+        return self.roi_label or self.R.choice(np.unique(mask).tolist())
+
+    def _get_foreground_mask(self, mask_path: str, level: int = 0) -> np.ndarray:
+        """Return a binary foreground mask using the specified label or a random label if roi_label is None."""
+        mask = np.asarray(Image.open(mask_path))[..., 0]
+        section = self._get_sub_section_label(mask)
+        foreground_mask = mask == section
```
ant0nsc, line 265: Are you sure this `==` operation works when `section` is a list of integers, as in line 259?

ant0nsc (changes requested):
I have a suspicion that that the transform might break if you supply a `roi_label` that is not an integer? Can you comment if I'm misunderstanding something here, or add tests for this case?

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -2,6 +2,7 @@
 #  Copyright (c) Microsoft Corporation. All rights reserved.
 #  Licensed under the MIT License (MIT). See LICENSE in the repo root for license information.
 #  ------------------------------------------------------------------------------------------
+from git import Sequence
```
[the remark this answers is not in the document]
kenza-bouzid: Oups, that was supposed to be from typing (auto import)

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -41,23 +43,30 @@ def segment_foreground(slide: np.ndarray, threshold: Optional[float] = None) ->
...
+    MASKSUBROI = "sub_section"
 
 
 class WSIBackend(str, Enum):
     """Options for the WSI reader backend."""
-    OPENSLIDE = 'OpenSlide'
-    CUCIM = 'cuCIM'
+
+    OPENSLIDE = "OpenSlide"
+    CUCIM = "cuCIM"
```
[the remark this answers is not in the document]
kenza-bouzid: They don't have constants for it 
They're using literals
image

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -204,38 +214,124 @@ def __call__(self, data: Dict) -> Dict:
         return data
 
 
+class LoadMaskSubROId(MapTransform, RandomizableTransform, BaseLoadROId):
+    """Transform that loads a pathology slide cropped to a sub-region of the mask bounding box (ROI).
+    The masks are supposed to be in png format, with the background in black (0) and tissue sections in categorical
+    values (e.g., 128, 255) along the RGB channels. The sub ROI is sampled from the foreground by randomly selecting a
+    label from unique values in the mask if roi_label is None. Otherwise the sub ROI is sampled from the foreground of
```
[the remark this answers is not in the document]
kenza-bouzid: we pick it from the R channel (assuming that they are consistent), I should precise it in the docs, would prefer having a parameter for thta?

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -204,38 +214,124 @@ def __call__(self, data: Dict) -> Dict:
...
+    the specified label.
+
+    Operates on dictionaries, replacing the file paths in `image_key` loaded array, in (C, H, W) format. Also adds the
+    following meta-data entries:
+    - `'location'` (tuple): top-right coordinates of the bounding box
+    - `'size'` (tuple): width and height of the bounding box
+    - `'level'` (int): chosen magnification level
```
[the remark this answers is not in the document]
kenza-bouzid: yes, level is not magnification, it's the "magnification level" in the pyramid.
we don't have a way of retrieving magnification from the WSI object other then reading it from the properties which depends on the format anf reader (openslide vs cucim, ...), so it might be tricky to come up with something generic for all...

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
kenza-bouzid: Thanks for catching it, I used to have a list of roi_labels than switch it to a single label

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
```diff
@@ -204,38 +214,124 @@ def __call__(self, data: Dict) -> Dict:
...
+        roi_label: Optional[Sequence[int]] = None,
+        **kwargs: Any,
+    ) -> None:
+        """
+        :param image_key: Image key in the input and output dictionaries. Default: 'image'.
+        :param mask_key: Mask key in the input and output dictionaries. Default: 'mask'.
+        :param wsi_mag_at_level0: WSI magnification at level 0. Default: 10x.
+        :param mask_mag: Mask magnification. Default: 1.25x.
+        :param roi_label: The label corresponding to the foreground tissue to sub sectionned. If None (default), a
+            random label is sampled from the foreground tissue uniformly.
```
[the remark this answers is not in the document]
kenza-bouzid: sorry that was confusing it's actually only a single value

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
kenza-bouzid: no, see comment above, roi_label type should Optionl[int]

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
ant0nsc: Ah, makes sense now, I remember in the version of the code that I reviewed there was still a `.tolist()` called somewhere, which I thought is the last operation, but it's R.choice

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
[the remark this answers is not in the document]
ant0nsc: OK. Would it be helpful to have the actual magnification available?

ant0nsc (approved):
Looks good! Sorry I got confused about the `.tolist()` in my first review

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
kenza-bouzid: My bad, the typing was confusing

hi-ml-cpath/src/health_cpath/preprocessing/loading.py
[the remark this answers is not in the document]
kenza-bouzid: it's not used anywhere else in the pipeline, we always use level..

harshita-s (approved):

PR #840, opened 2023-03-16
ENH: Allow variants of model parameters in the same class

ant0nsc:
Based on a commandline flag, models can now switch between different hyperparameters


harshita-s (approved):

harshita-s:
Looks good, thank you for adding model variants!

PR #853, opened 2023-03-24
ENH: Expose v1 dataset overwrite

### LINKED ISSUE #839: Expose overwrite option for output datasts in SDK v1

peterhessey:
Closes #839

### DIFF
--- a/docs/source/datasets.md
+++ b/docs/source/datasets.md
@@ -43,7 +43,7 @@ To simplify usage, the `hi-ml` package creates AzureML datasets for you. All you
 From that point on, you can drop a folder of files in the container that holds your data. Within the `hi-ml` package,
 just reference the name of the folder, and the package will create a dataset for you, if it does not yet exist.
 
-## Using the datasets (SDK v1)
+## Using the datasets - SDK v1
 
 The simplest way of specifying that your script uses a folder of data from blob storage is as follows: Add the
 `input_datasets` argument to your call of `submit_to_azure_if_needed` like this:
@@ -53,7 +53,8 @@ from health_azure import submit_to_azure_if_needed
 run_info = submit_to_azure_if_needed(...,
                                      input_datasets=["my_folder"],
                                      default_datastore="my_datastore",
-                                     strictly_aml_v1=True)
+                                     strictly_aml_v1=True,
+                                    )
 input_folder = run_info.input_datasets[0]
 ```
 
@@ -82,7 +83,8 @@ run_info = submit_to_azure_if_needed(...,
                                      input_datasets=["my_folder"],
                                      output_datasets=["new_dataset"],
                                      default_datastore="my_datastore",
-                                     strictly_aml_v1=True))
+                                     strictly_aml_v1=True,
+                                    )
 input_folder = run_info.input_datasets[0]
 output_folder = run_info.output_datasets[0]
 ```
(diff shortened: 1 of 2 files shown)


ant0nsc (changes requested):
I think that's somewhat incomplete? There's no easy way of using that additional parameter. It would make sense to add `overwrite` in the `DatasetConfig` class.

peterhessey:
> I think that's somewhat incomplete? There's no easy way of using that additional parameter. It would make sense to add `overwrite` in the `DatasetConfig` class.

Ah okay, didn't realise we wanted it exposed all the way up to the runner.py call!

ant0nsc (approved):
Looking a lot better! Are there v1/v2 SDK differences now? if yes, please add docu or bring codepaths in line.

ant0nsc:
Could also be good info about that flag to add that to the `datasets.md` docu file

docs/source/datasets.md
```diff
@@ -66,20 +75,24 @@ dataset (or even more than one).
...
 You can use that via the `output_datasets` argument:
+
 ```python
 from health_azure import submit_to_azure_if_needed
 run_info = submit_to_azure_if_needed(...,
                                      input_datasets=["my_folder"],
                                      output_datasets=["new_dataset"],
-                                     default_datastore="my_datastore")
+                                     default_datastore="my_datastore",
+                                     strictly_aml_v1=True))
```
kenza-bouzid, line 85:
```suggestion
                                     strictly_aml_v1=True,
                                     )
```

kenza-bouzid: can you please add trailing commas after all  strictly_aml_v1=True,

docs/source/datasets.md
```diff
@@ -132,24 +150,45 @@ Note that mounting the dataset locally is only supported on Linux because it req
...
 This is also true when running locally - if `local_folder` is not specified and an AzureML workspace can be found, then the dataset will be downloaded or mounted to the `target_folder`.
 
+### Overwriting existing output datasets
+
+When creating an output dataset with the same name as an existing dataset, the default behaviour of `hi-ml` is to overwrite the existing dataset to avoid uploading errors. If you wish to disable this behaviour, it can be controlled using the `overwrite_existing` parameter:
+
+```python
+from health_azure import DatasetConfig, submit_to_azure_if_needed
+output_dataset = DatasetConfig(name="my_folder",
+                              datastore="my_datastore",
```
kenza-bouzid, line 176: fix tab allignement

kenza-bouzid (approved):

docs/source/datasets.md
```diff
@@ -132,24 +150,45 @@ Note that mounting the dataset locally is only supported on Linux because it req
...
+                                     strictly_aml_v1=True)
 # Input_folder will now be "/tmp/mnist"
 input_folder = run_info.input_datasets[0]
 ```
 
 This is also true when running locally - if `local_folder` is not specified and an AzureML workspace can be found, then the dataset will be downloaded or mounted to the `target_folder`.
 
+### Overwriting existing output datasets
+
+When creating an output dataset with the same name as an existing dataset, the default behaviour of `hi-ml` is to overwrite the existing dataset to avoid uploading errors. If you wish to disable this behaviour, it can be controlled using the `overwrite_existing` parameter:
```
ant0nsc, line 171: Not sure I understand "to avoid uploading errors"? You could add something that overwriting is helpful if you first run of a script fails, and you want to re-run.

PR #854, opened 2023-03-27
BUG: Prevent creating v2 data asset without specifying datastore

### LINKED ISSUE #786: Creating a new data assets uses default datastore by default
We've encountered this issue recently:

```shell
Dataset initialization failed: UserErrorException:
	Message: Failed with visit error: Failed with execution error: error in streaming from input data sources
	VisitError(ExecutionError(StreamError(NotFound)))
=> Failed with execution error: error in streaming from input data sources
	ExecutionError(StreamError(NotFound))
	InnerException None
	ErrorResponse 
{
    "error": {
        "code": "UserError",
        "message": "Failed with visit error: Failed with execution error: error in streaming from input data sources\n\tVisitError(ExecutionError(StreamError(NotFound)))\n=> Failed with execution error: error in streaming from input data sources\n\tExecutionError(StreamError(NotFound))"
    }
}
```

This happened because the data asset was created using the workspace's default datastore. On shared workspaces like (hai4, mprr1ws), the default datastore doesn't contain our private data. 

One should specify --datastore when submitting the first ever job that creates the data assets as we can not delete/edit data assets once they are created.

Using the default workspace's datastore by default is not always desired, can we come up with another strategy for default datastores?

 Possible options:
 - The user has to specify a datastore for the first job: in get_or_create_data_asset only create the data asset if datastore is specified, this would force the user to be careful the first time and specify the right datastore
 - Before using the default workspace's datastore: can we check that the dataset exists already? if not throw an error: `Input Dataset doesn't exist in default datastore, please specify the right datastore` ?  
 
 cc @ant0nsc @peterhessey

peterhessey:
Closes #786 

It should now be impossible to create data new data assets without also specifying a datastore by its specific name, in both v1 and v2 pathways.

Default datastore interactions in all other code paths have been left unchanged.


ant0nsc (approved):
One of the tests is not doing what you think it should...
Also, is there a mention of default datastores anywhere in the documentation? If so, this should be cleaned up too.

kenza-bouzid (approved):
LGTM, we've discussed once to add a special char for default dataset (maybe -1 or 0), to make it easy to use default datastore, is that still planned?

peterhessey:
> LGTM, we've discussed once to add a special char for default dataset (maybe -1 or 0), to make it easy to use default datastore, is that still planned?

I discussed this with Anton and we agreed that if the user is creating a new data asset then they _must_ specify the name of the datastore to be on the safe side. Having the option of using a default datastore with a flag or special char potentially keeps the door open for accidental creation in the wrong datastore still which is what this PR is trying to circumvent

PR #859, opened 2023-03-29
ENH: Ability to evaluate model on a new dataset

ant0nsc:
Renamed the `MLRunner` to `TrainingRunner`. Introduced a new `EvalRunner`. Shared functionality moved from `MLRunner` to `RunnerBase`.


hi-ml/src/health_ml/runner.py
```diff
@@ -325,15 +326,29 @@ def run_in_situ(self, azure_run_info: AzureRunInfo) -> None:
...
+                set_environment_variables_for_multi_node()
+            self.ml_runner = TrainingRunner(
+                experiment_config=self.experiment_config,
+                container=self.lightning_container,
+                project_root=self.project_root,
+            )
+            self.ml_runner.validate()
+            self.ml_runner.run_and_cleanup(azure_run_info)
+        elif self.experiment_config.mode == RunnerMode.EVAL:
+            self.eval_runner = EvalRunner(
```
kenza-bouzid, line 342: use same attribute self.ml_runner for both (see comment above about typing) and call 
self.ml_runner.validate()
self.ml_runner.run_and_cleanup(azure_run_info)
outside the if else statement (TODO for DRY)

hi-ml/testhiml/testhiml/test_eval_runner.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+@pytest.fixture(scope="function")
+def hello_world_checkpoint(ml_runner_with_container: TrainingRunner, tmp_path: Path) -> Path:
+    container = ml_runner_with_container.container
+    container.set_output_to(tmp_path)
+    container.max_epochs = 5
+    ml_runner_with_container.run()
+    # Read out the trained checkpoint and use the runner in eval mode
+    return container.get_checkpoint_to_test()
```
kenza-bouzid, line 28: You can also use the checkpoint in 
hi-ml/hi-ml/testhiml/test_data/hello_world_checkpoint.ckpt 
it will be faster than retraining

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -66,12 +71,15 @@ def ml_runner_with_container() -> Generator:
 
 
 @pytest.fixture()
-def ml_runner_with_run_id() -> Generator:
+def ml_runner_with_src_ckpt() -> Generator:
+    """
+    A fixture with a training runner for the HelloWorld model that has a src_checkpoint set.
```
kenza-bouzid, line 76: can we give this fixture a more specific name, src_ckpt can be one of (run_id, local path, url) hence the previous name

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -457,12 +495,12 @@ def test_model_weights_when_resume_training() -> None:
     experiment_config = ExperimentConfig(model="HelloWorld")
     container = HelloWorld()
     container.max_num_gpus = 0
-    container.src_checkpoint = CheckpointParser(mock_run_id(id=0))
```
kenza-bouzid, line 460: I'm not sure this change has the same behaviour, we want to make sure we're able to resume from a previous aml run, str(hello_world_checkpoint) is a local path outside AML

hi-ml/src/health_ml/eval_runner.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def validate(self) -> None:
+        """Checks if the fields of the class are set up correctly."""
+        if self.container.src_checkpoint is None or self.container.src_checkpoint.checkpoint == "":
+            raise ValueError(
+                "To use model evaluation, you need to provide a checkpoint to use, via the --src_checkpoint argument."
+            )
+
+    def run(self) -> None:
+        """Start the core workflow that the class implements: Initialize the trainer and run inference on the
```
mebristo, line 20: To the untrained eye it might be confusing why we'd need to 'initialize the trainer' here in the EvalRunner - can we explain why this is necessary here? Also we could mention that if the container has an extra_val_epoch method it will be run first before inference starts

kenza-bouzid (approved):
Great work, only minor comments!

hi-ml/src/health_ml/eval_runner.py
kenza-bouzid: Oh I don't thing extra_val_epoch is run here, only run_infererence, this is for an evaluation on a external dataset so all data goes into the test_datamodule

hi-ml/src/health_ml/training_runner.py
```diff
@@ -0,0 +1,238 @@
...
+
+            self.end_training(environ_before_training)
+
+        self.init_inference()
+
+        with logging_section("Model validation"):
+            self.run_validation()
+
+        with logging_section("Model inference"):
+            self.run_inference()
```
mebristo, line 236: Do we still want to call `init_inference` and `run_inference` in TrainingRunner?

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -18,34 +18,38 @@
...
+from testhiml.utils.fixed_paths_for_tests import full_test_data_path, mock_run_id
 
 no_gpu = not is_gpu_available()
+hello_world_checkpoint = full_test_data_path(suffix="hello_world_checkpoint.ckpt")
 
 
-@pytest.fixture(scope="module")
-def ml_runner_no_setup() -> MLRunner:
+@pytest.fixture()
+def ml_runner_no_setup(tmp_path: Path) -> TrainingRunner:
```
mebristo, line 39: For consistency, should the names of these fixtures be updated to 'training_runner_...)

hi-ml/testhiml/testhiml/test_run_ml.py
mebristo, line 1: Since run_ml.py no longer exists, maybe this file could be renamed to test_training_runner.py?

hi-ml/testhiml/testhiml/test_run_ml.py
```diff
@@ -408,45 +446,45 @@ def test_run(run_inference_only: bool, run_extra_val_epoch: bool, ml_runner_with
 
 
 @pytest.mark.parametrize("run_extra_val_epoch", [True, False])
-def test_run_inference_only(run_extra_val_epoch: bool, ml_runner_with_run_id: MLRunner) -> None:
+def test_run_inference_only(run_extra_val_epoch: bool, ml_runner_with_src_ckpt: TrainingRunner) -> None:
```
mebristo, line 449: Would it make more sense for this to live in `test_eval_runner.py`? Is there actually  a use case for creating a TrainingRunner when run_inference_only = True?

hi-ml/src/health_ml/eval_runner.py
mebristo: But `init_inference` is called here and within that (lines 204-205 of runner_base.py), `on_run_extra_validation_epoch` is called (if `run_extra_val_epoch`)...or am I missing something?

hi-ml/src/health_ml/runner.py
```diff
@@ -104,8 +105,8 @@ def __init__(self, project_root: Path):
         self.project_root = project_root
         self.experiment_config: ExperimentConfig = ExperimentConfig()
         self.lightning_container: LightningContainer = None  # type: ignore
-        # This field stores the MLRunner object that has been created in the most recent call to the run() method.
-        self.ml_runner: Optional[MLRunner] = None
+        # This field stores the TrainingRunner object that has been created in the most recent call to the run() method.
+        self.ml_runner: Optional[TrainingRunner] = None
```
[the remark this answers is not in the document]
ant0nsc: done

hi-ml/src/health_ml/training_runner.py
ant0nsc: yes, we want to keep the legacy behaviour: When training a model, we will always also call inference on the test set.

hi-ml/testhiml/testhiml/test_run_ml.py
ant0nsc: I think that would be a good next step to refactor. We should introduce more modes and delete the "run_inference_only" flag

hi-ml/testhiml/testhiml/test_run_ml.py
ant0nsc: I made this change mostly for test speed. I would assume that all the handling of the different checkpoint types is done correctly by the `CheckpointParser` class, so I don't need to check it here again.

hi-ml/src/health_ml/eval_runner.py
kenza-bouzid: correct, but this is only to tell the container to switch some plot options, the extra val epoch is run by calling run_validation

mebristo: I see!

ant0nsc: Good point, @mebristo . I've cleaned that up, now there is no such hook run when doing eval on a new dataset.

PR #860, opened 2023-03-30
ENH: Adds BioViL-T Linear Classifier (Chexpert Classes)

ozan-oktay:
- Adds the capability to instantiate a linear classifier on top frozen BioViL-T image embeddings.
- Note that the model is not trained in an end-to-end fashion.


fepegar:
Shall we review now or wait until it's not a draft anymore?

PR #873, opened 2023-04-17
BUG: Fix extra validation epoch hook call order

kenza-bouzid:
Fix the order of hook call. on_run_extra_val_epoch needs to be run before super().init_inference to account for any changes in datamodule params before re-instantiation

### DIFF
--- a/hi-ml/testhiml/testhiml/test_training_runner.py
+++ b/hi-ml/testhiml/testhiml/test_training_runner.py
@@ -354,7 +354,6 @@ def test_init_inference(
                 assert mock_create_trainer.call_args[1]["num_nodes"] == 1
                 assert mock_create_trainer.call_args[1]["mlflow_run_for_logging"] == expected_mlflow_run_id
                 mock_get_data_module.assert_called_once()
-                mock_get_data_module.asse
                 assert training_runner_hello_world_with_checkpoint.data_module == "dummy_data_module"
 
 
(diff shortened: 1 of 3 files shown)


harshita-s (approved):
thank you, looks good. Just wondering if we could add a test to capture this bug?

hi-ml/testhiml/testhiml/test_training_runner.py
```diff
@@ -324,6 +325,12 @@ def test_init_inference(
                 assert not training_runner_hello_world_with_checkpoint.container.model._on_extra_val_epoch
 
                 training_runner_hello_world_with_checkpoint.init_inference()
+                if run_extra_val_epoch:
+                    assert (
+                        caplog.messages[-3]
+                        == "Preparing to run an extra validation epoch to evaluate the model on the validation set."
+                    )
+                assert caplog.messages[-2] == "Preparing runner for inference."
```
ant0nsc, line 333: If the call order was swapped back to what it was before your change, would this test fail? Maybe better check if the inference bag sizes are changed as expected?

ant0nsc (approved):
Looking good! Just one comment around test "strictness"

hi-ml/testhiml/testhiml/test_training_runner.py
kenza-bouzid: yes, because of the order in the messages, I know it's not ideal because the logs order doesn't guarantee the right functions call order 
note that the change in bag size is specific to MIL container, also get_data_module is mocked here, do you know if there's a way to check if a mock has been called before another?

PR #874, opened 2023-04-18
ENH: Add account key argument to generate checkpoint URL script 

kenza-bouzid:
Some datastores don't have access to account key, so we need a way of specifying it via commandline

### DIFF
--- a/hi-ml-cpath/src/health_cpath/scripts/generate_checkpoint_url.py
+++ b/hi-ml-cpath/src/health_cpath/scripts/generate_checkpoint_url.py
@@ -13,6 +13,7 @@ def get_checkpoint_url_from_aml_run(
     expiry_days: int = 1,
     workspace_config_path: Optional[Path] = None,
     aml_workspace: Optional[Workspace] = None,
+    account_key: Optional[str] = None,
     sas_token: Optional[str] = None,
 ) -> str:
     """Generate a SAS URL for the checkpoint file in the given run.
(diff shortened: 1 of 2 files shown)


harshita-s (approved):

hi-ml-cpath/src/health_cpath/scripts/generate_checkpoint_url.py
```diff
@@ -19,24 +20,23 @@ def get_checkpoint_url_from_aml_run(
...
     :param expiry_days: The number of days the SAS URL is valid for, defaults to 30.
     :param workspace_config_path: The path to the workspace config file, defaults to None.
     :param aml_workspace: The Azure ML workspace to use, defaults to the default workspace.
     :param sas_token: The SAS token to use, defaults to None.
     :return: The SAS URL for the checkpoint.
     """
     workspace = get_workspace(aml_workspace=aml_workspace, workspace_config_path=workspace_config_path)
-    datastore = workspace.get_default_datastore()
-    account_name = datastore.account_name
+    account_name = workspace.get_details()['storageAccount'].split('/')[-1]
```
ant0nsc, line 31: It would be helpful here to print out the name of the storage account. Chances are high that the account key does not match. something like "Workspace {workspace.name} stores its run results in storage account {account_name}."

ant0nsc (approved):
Would be good to have more progress and diagnostic information - otherwise looking good!

hi-ml-cpath/src/health_cpath/scripts/generate_checkpoint_url.py
kenza-bouzid: done

PR #875, opened 2023-04-18
ENH: Clean up design of slides dataset

ant0nsc:
Metadata columns are presently set via class variables, rather than in the constructor. Hence, we can't have several datasets that use different metadata columns, without them interfering with each other. This caused odd test failures because test results now depend on the order in which tests are executed.

### DIFF
--- a/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
+++ b/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
@@ -157,11 +157,6 @@ class SlidesDataset(Dataset):
     """Base class for datasets of WSIs, iterating dictionaries of image paths and metadata.
 
     The output dictionaries are indexed by `..utils.naming.SlideKey`.
-
-    :param SLIDE_ID_COLUMN: CSV column name for slide ID.
-    :param IMAGE_COLUMN: CSV column name for relative path to image file.
-    :param SPLIT_COLUMN: CSV column name for train/test split (optional).
-    :param DEFAULT_CSV_FILENAME: Default name of the dataset CSV at the dataset rood directory.
     """
 
     def __init__(
--- a/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
+++ b/hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
@@ -161,13 +161,6 @@ class SlidesDataset(Dataset):
     :param DEFAULT_CSV_FILENAME: Default name of the dataset CSV at the dataset rood directory.
     """
 
-    SLIDE_ID_COLUMN: str = 'slide_id'
-    IMAGE_COLUMN: str = 'image'
-    MASK_COLUMN: Optional[str] = None
-    SPLIT_COLUMN: Optional[str] = None
-
-    METADATA_COLUMNS: Tuple[str, ...] = ()
-
     DEFAULT_CSV_FILENAME: str = "dataset.csv"
 
     def __init__(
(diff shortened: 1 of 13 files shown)


hi-ml-cpath/src/health_cpath/datasets/tcga_prad_dataset.py
```diff
@@ -11,6 +11,9 @@
 from health_cpath.datasets.base_dataset import DEFAULT_LABEL_COLUMN, SlidesDataset
 
 
+TCGA_PRAD_DATASET_FILE = "dataset.csv"
```
kenza-bouzid, line 14: is this necessary? we can use DEFAULT_CSV_FILENAME from base_dataset? let's try to limit duplicated constants

hi-ml-cpath/src/health_cpath/utils/tiff_conversion_config.py
```diff
@@ -95,12 +95,12 @@ def create_dataset_csv_for_converted_data(self, output_folder: Path) -> None:
...
         new_dataset_df = deepcopy(self.slides_dataset.dataset_df)
-        new_dataset_df[self.slides_dataset.IMAGE_COLUMN] = (
-            new_dataset_df[self.slides_dataset.IMAGE_COLUMN]
+        new_dataset_df[self.slides_dataset.image_column] = (
+            new_dataset_df[self.slides_dataset.image_column]
             .str.replace(AMPERSAND, self.replace_ampersand_by)
             .map(lambda x: str(Path(x).with_suffix(TIFF_EXTENSION)))
         )
-        new_dataset_path = output_folder / (self.converted_dataset_csv or self.slides_dataset.DEFAULT_CSV_FILENAME)
+        new_dataset_path = output_folder / (self.converted_dataset_csv or self.slides_dataset.dataset_csv.name)
```
kenza-bouzid, line 103: why not self.slides_dataset.default_csv_filename is it kept as an attribute?

kenza-bouzid (approved):
looks great, but I think we can do better with constant variables design? we have a lot of duplications in the code base...

hi-ml-cpath/src/health_cpath/datasets/tcga_prad_dataset.py
```diff
@@ -39,7 +38,15 @@ def __init__(
...
         """
-        super().__init__(root, dataset_csv, dataset_df, validate_columns=False, label_column=label_column)
+        super().__init__(
+            root,
+            dataset_csv,
+            dataset_df,
+            validate_columns=False,
+            label_column=label_column,
+            default_csv_filename=TCGA_PRAD_DATASET_FILE,
+            image_column="image_path",
```
harshita-s, line 48: "image_path" could be a constant?

harshita-s (approved):

hi-ml-cpath/src/health_cpath/datasets/base_dataset.py
```diff
@@ -180,44 +174,60 @@ def __init__(
         label_column: str = DEFAULT_LABEL_COLUMN,
```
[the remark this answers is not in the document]
ant0nsc: In this case, I would rather not. The contents of those literals matches, yes, but its use is different.
DEFAULT_LABEL_COLUMN is for use on input files, whereas SlideKey is part of our internal dictionary. The two don't need
to be in sync.

hi-ml-cpath/src/health_cpath/datasets/panda_dataset.py
```diff
@@ -7,6 +7,12 @@
 from typing import Any, Dict, Union, Optional
 from health_cpath.datasets.base_dataset import SlidesDataset
 
+PANDA_SLIDE_ID_COLUMN = 'image_id'
```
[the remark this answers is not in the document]
ant0nsc: good point!

hi-ml-cpath/src/health_cpath/datasets/tcga_prad_dataset.py
ant0nsc: Similar comment here: yes, the constant happens to have the same value, but it's use may be different. I'd use the same
constant for everything that MUST be changed together. For this constant, it is used in the script that writes the
dataset and here where we read it.

hi-ml-cpath/src/health_cpath/utils/tiff_conversion_config.py
ant0nsc: Yup.

PR #877, opened 2023-04-28
ENH: Add a link to the model runner at repo root

ant0nsc:
To simplify finding the runner I'm somewhere else in the directory tree


ant0nsc:
Breaks because of AML v2 issues with symlinks

PR #880, opened 2023-05-03
BUG: Hyperdrive jobs not possible with v2 SDK

### LINKED ISSUE #870: Hyperdrive jobs can't be submitted with SDK v2

ant0nsc:


fepegar:
In case it helps, this has been driving me a bit crazy today as well. I've found from older runs of mine that the shared memory is correctly taken into account when downgrading to `azure-ai-ml==1.5.0`. I was trying `1.15.0` so I think that feature was broken somewhere in between.

fepegar:
> In case it helps, this has been driving me a bit crazy today as well. I've found from older runs of mine that the shared memory is correctly taken into account when downgrading to `azure-ai-ml==1.5.0`. I was trying `1.15.0` so I think that feature was broken somewhere in between.

I've run some tests. This is broken for `azure-ai-ml>=1.13.0`.

PR #889, opened 2023-05-22
BUG: DeepMIL model uses training data even when running in eval mode

ant0nsc:
Instantiating the DeepMIL classifier in `create_model` always relies on the training dataset and module, which is not available when doing evaluation. This PR makes that more flexible and allows to hardcode the number of classes to use for evaluation.

### DIFF
--- a/hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
+++ b/hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
@@ -297,7 +297,6 @@ class BaseMIL(LightningContainer, LoadingParams, EncoderParams, PoolingParams, C
 
     def create_model(self) -> DeepMILModule:
         num_classes = self.get_num_classes()
-
         outputs_handler = self.get_outputs_handler()
         deepmil_module = DeepMILModule(
             label_column=self.get_label_column(),
--- a/hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
+++ b/hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
@@ -260,7 +260,7 @@ class BaseMIL(LightningContainer, LoadingParams, EncoderParams, PoolingParams, C
         :return: A data module.
         """
         if self.data_module is None:
-            if self.runner_mode == RunnerMode.TRAINING:
+            if self.runner_mode == RunnerMode.TRAIN:
                 self.data_module = self.get_data_module()
             elif self.runner_mode == RunnerMode.EVAL_FULL:
                 self.data_module = self.get_eval_data_module()
--- a/hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
+++ b/hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
@@ -13,6 +13,7 @@ from pathlib import Path
 from monai.transforms import Compose
 from pytorch_lightning.callbacks import Callback
 from pytorch_lightning.callbacks.model_checkpoint import ModelCheckpoint
+import torch
 
 from health_azure.utils import create_from_matching_params
 from health_cpath.preprocessing.loading import LoadingParams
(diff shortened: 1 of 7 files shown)


hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
```diff
@@ -241,14 +253,33 @@ def get_transforms_dict(self, image_key: str) -> Optional[Dict[ModelKey, Union[C
...
+        """Gets the number of classes that the model handles. In training mode, this is the number of classes in the
+        training dataset. In evaluation mode, this is read directly from the `eval_num_classes` attribute.
+
+        This method has a side effect in training mode: It creates the data module if it is not yet set in the
+        `data_module` attribute.
+        """
+        if self.runner_mode == RunnerMode.EVAL_FULL:
+            # We are in evaluation mode, so we don't necessarilty have access to the training data module. The eval
+            # data module may not contain all classes, so we need to hardcode those values.
+            return self.eval_num_classes
```
kenza-bouzid, line 266: see comment above 

```suggestion
def get_num_classes(self) -> int:
        if self.num_classes is not None: 
             return self.num_classes 
        if (not hasattr(self, "data_module")) or self.data_module is None:
                if self.runner_mode == RunnerMode.EVAL_FULL:
                    self.data_module = self.get_eval_data_module()
                else:
                    self.data_module = self.get_data_module()
            return self.data_module.train_dataset.n_classes
```

hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
```diff
@@ -241,14 +253,33 @@ def get_transforms_dict(self, image_key: str) -> Optional[Dict[ModelKey, Union[C
...
+        else:
+            if (not hasattr(self, "data_module")) or self.data_module is None:
+                self.data_module = self.get_data_module()
+            return self.data_module.train_dataset.n_classes
+
     def create_model(self) -> DeepMILModule:
-        self.data_module = self.get_data_module()
+        num_classes = self.get_num_classes()
+        # Class weights are only used to create a loss function, which is irrelevant in evaluation mode.
+        class_weights = None if self.runner_mode == RunnerMode.EVAL_FULL else self.data_module.class_weights
```
kenza-bouzid, line 275: wondering if we would still need class_weights those in eval mode to compute the test loss? 
Otherwise YESS, it seems like a good solution

kenza-bouzid (approved):
LGTM :) thanks for fixing this, a small suggestion for num_classes

hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
kenza-bouzid: eval test loss will be different than training mode, but it's probably ok

ant0nsc: as discussed, we don't want to use loss for eval, will add a comment

hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
ant0nsc: Thanks for that suggestion. I've first added it in, but then reverted. I wanted to keep the change limited to the problem we have now. For training, reading out all properties from the dataset is still the most flexible approach.

hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
```diff
@@ -101,7 +109,11 @@ class BaseMIL(LightningContainer, LoadingParams, EncoderParams, PoolingParams, C
     )
 
     def __init__(self, **kwargs: Any) -> None:
+        """
+        :param eval_num_classes: Number of classes to use when running in eval mode. This is provided because
```
dccastro, line 113: What exactly is the problem with having fewer classes in the eval data? If anything really breaks, could it be better to fix that instead?

Also, is this even a issue for binary tasks? Recall that our pipeline treats `num_classes=1` as binary and `=2` as multi-class with 2 categories—so AFAIK we can't even say there's a single class present...?

dccastro (approved):
I can see how this fix might patch over issues with having fewer classes present at evaluation, though it sounds like there may be root causes that could be addressed instead. However, I suspect I'm missing some of the context for this change, so please feel free to ignore my comments if not appropriate.

hi-ml-cpath/src/health_cpath/configs/classification/BaseMIL.py
ant0nsc: Having fewer classes in eval is not the problem we are solving here. We don't want to rely on the training data to
determine the number of classes, because it means making the training dataset available.

PR #904, opened 2023-07-25
BUG: Allow relative path for entry script for SDK v2 submission

### LINKED ISSUE #890: Job submission of scripts in subfolders of cwd fails, saying the script should be under project root
Submitting `~foo/bar.py` from `~`, when manually setting snapshoot root to ~, fails with an error

ant0nsc:
When submitting a script with a path relative to current working directory, SDK v2 raises an error, even though the script exists.

Closes #890


hi-ml-azure/src/health_azure/himl.py
```diff
@@ -340,8 +343,8 @@ def create_crossval_hyperparam_args_v2(
 
 def create_script_run(
     script_params: List[str],
-    snapshot_root_directory: Optional[Path] = None,
-    entry_script: Optional[PathOrString] = None,
+    snapshot_root_directory: Optional[Path],
```
kenza-bouzid, line 346: I wonder if this might have some side effects? 
especially for scripts that are not tested, I had a quick look at present scripts that are submitted to AML (tiles creation...), looks ok, but we might need to update docs

kenza-bouzid (approved):
LGTM!

hi-ml-azure/src/health_azure/himl.py
ant0nsc: It is not used across hi-ml, and anyway it should not be used because it's an internal method.

Shruthi42 (approved):

PR #910, opened 2023-11-01
ENH: Upgrade Python and other packages

ant0nsc:
- Upgrade to Python 3.9 as a minimum
- Upgrade pillow because of security alerts. This required some API changes and update of test result images
- The behaviour of the `LogCaptureFixture` changed, requiring changes from `logging.info` to `logger.info`
- Downgraded `black` from 23 to 22 because of version constraints
- Changes in calls to `WorkflowParams` because of `mypy` issues that were not detected until now
- `hi-ml-azure` now also has its own minimal Conda dev environment


docs/source/examples/1/environment.yml
fepegar, line 1: Would it help to have a single YAML for all the examples?

fepegar (approved):

dccastro (approved):
Thanks for the monumental effort in handling all the version conflicts and breaking changes! I assume it's all working now based on the tests, and the code changes LGTM.

docs/source/examples/1/environment.yml
ant0nsc: Probably yes - Jonathan's thinking here was to have each stage in a self-contained folder. 
However, doing the query/replace here is the least of the problems when doing version upgrades. So I would not bother with it here.

PR #936, opened 2024-05-13
ENH: Add support to pass a prefix for hyperparam args

fepegar:
When `hyperparam_args` is passed, the command is modified to add the argument in the settings. For example, if the looks like `{learning_rate: [1, 2, 3]}`, something like `--learning_rate=${{inputs.learning_rate}}` will be added to the command. But sometimes we wouldn't want the dashes, e.g., when using Hydra (`learning_rate=${{inputs.learning_rate}}` or maybe `+learning_rate=${{inputs.learning_rate}}`). This PR adds support to specify the prefix for the argument (default: `"--"`). For Hydra, we might want, e.g., `""` or `"+"`.


vale-salvatelli (approved):
Looks good

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -419,6 +419,7 @@ def submit_run_v2(
     pytorch_processes_per_node: Optional[int] = None,
     use_mpi_run_for_single_node_jobs: bool = True,
     display_name: Optional[str] = None,
+    argument_prefix: str = "--",
```
ant0nsc, line 422:
```suggestion
    hyperdrive_argument_prefix: str = "--",
```

ant0nsc (approved):

hi-ml-azure/src/health_azure/himl.py
fepegar: Thanks! Fixed in [608d7ae](https://github.com/microsoft/hi-ml/pull/936/commits/608d7ae4ea7f018be79b16c99c2208849c27242a).

PR #961, opened 2025-02-14
ENH: Allow environments from docker build contexts

samb-t:
Currently environments can only be constructed from conda files. This PR allows adds the optional argument `build_context` to `submit_to_azure_if_needed` which can either be a `DockerBuildContext` for SDK v1 or a `BuildContext` for SDK v2. 

Build contexts are dockerfiles or folders containing a dockerfile that set up the docker environment.

For example to set up UV, the following dockerfile can be used
```dockerfile
FROM mcr.microsoft.com/azureml/openmpi4.1.0-cuda11.8-cudnn8-ubuntu22.04
WORKDIR /workspace
RUN curl -LsSf https://astral.sh/uv/install.sh | sh
ENV PATH="/root/.local/bin:$PATH"
CMD ["/bin/bash"]
```

The build context folder could contain a `pyproject.toml` file and `uv sync` added to the dockerfile. Alternatively, `uv sync` can be run at the start of a job, or just `uv run` can be used to execute the script and install all depencies.

### DIFF
--- a/hi-ml-azure/src/health_azure/himl.py
+++ b/hi-ml-azure/src/health_azure/himl.py
@@ -28,6 +28,7 @@ from azure.ai.ml.entities._job.distribution import DistributionConfiguration, Mp
 from azure.ai.ml.sweep import Choice
 from azureml._base_sdk_common import user_agent
 from azureml.core import ComputeTarget, Environment, Experiment, Run, RunConfiguration, ScriptRunConfig, Workspace
+from azureml.core.environment import DockerBuildContext
 from azureml.core.runconfig import DockerConfiguration, MpiConfiguration
 from azureml.data import OutputFileDatasetConfig
 from azureml.data.dataset_consumption_config import DatasetConsumptionConfig
@@ -149,6 +150,7 @@ def create_run_configuration(
     workspace: Workspace,
     compute_cluster_name: str,
     conda_environment_file: Optional[Path] = None,
+    docker_build_context: Optional[DockerBuildContext] = None,
     aml_environment_name: str = "",
     environment_variables: Optional[Dict[str, str]] = None,
     pip_extra_index_url: str = "",
@@ -175,6 +177,7 @@ def create_run_configuration(
         CPU or GPU machines.
     :param conda_environment_file: The conda configuration file that describes which packages are necessary for your
         script to run.
+    :param docker_build_context: The Docker build context that should be used to create a new Docker image.
     :param environment_variables: The environment variables that should be set when running in AzureML.
     :param docker_base_image: The Docker base image that should be used when creating a new Docker image.
     :param docker_shm_size: The Docker shared memory size that should be used when creating a new Docker image.
@@ -194,8 +197,20 @@ def create_run_configuration(
     """
     run_config = RunConfiguration()
 
+    # If both docker_build_context and conda_environment_file are provided, throw an error
+    if docker_build_context and conda_environment_file:
+        raise ValueError("Both docker_build_context and conda_environment_file cannot be provided at the same time.")
+
     if aml_environment_name:
         run_config.environment = Environment.get(workspace, aml_environment_name)
+    elif docker_build_context:
+        environment_name = load_and_hash_directory(Path(docker_build_context.location))
+        new_environment = Environment.from_docker_build_context(
+            name=environment_name,
+            docker_build_context=docker_build_context,
+        )
+        registered_env = register_environment(workspace, new_environment)
+        run_config.environment = registered_env
     elif conda_environment_file:
         # Create an AzureML environment, then check if it exists already. If it exists, use the registered
         # environment, otherwise register the new environment.
@@ -212,7 +227,10 @@ def create_run_configuration(
         registered_env = register_environment(workspace, new_environment)
         run_config.environment = registered_env
     else:
-        raise ValueError("One of the two arguments 'aml_environment_name' or 'conda_environment_file' must be given.")
+        raise ValueError(
+            "One of the three arguments 'aml_environment_name', 'docker_build_context' or "
+            "'conda_environment_file' must be given."
+        )
 
     # By default, include several environment variables that work around known issues in the software stack
     run_config.environment_variables = {**DEFAULT_ENVIRONMENT_VARIABLES, **(environment_variables or {})}
@@ -471,7 +489,7 @@ def submit_run_v2(
         cmd = " ".join(["python", str(entry_script_relative), script_param_str])
     else:
         experiment_name = effective_experiment_name(experiment_name, entry_command)
-        cmd = " ".join([str(entry_command), entry_script_relative, script_param_str])
+        cmd = " ".join([str(entry_command), str(entry_script_relative), script_param_str])
 
     print(f"The following command will be run in AzureML: {cmd}")
 
@@ -735,7 +753,7 @@ def submit_to_azure_if_needed(  # type: ignore
     snapshot_root_directory: Optional[PathOrString] = None,
     script_params: Optional[List[str]] = None,
     conda_environment_file: Optional[PathOrString] = None,
-    build_context: Optional[BuildContext] = None,
+    build_context: Optional[Union[DockerBuildContext, BuildContext]] = None,
     aml_environment_name: str = "",
     experiment_name: Optional[str] = None,
     environment_variables: Optional[Dict[str, str]] = None,
@@ -924,6 +942,10 @@ def submit_to_azure_if_needed(  # type: ignore
         print(f"No snapshot root directory given. Uploading all files in the current directory {Path.cwd()}")
         snapshot_root_directory = Path.cwd()
 
+
+    if conda_environment_file is not None and build_context is not None:
+        raise ValueError("Only one of `conda_environment_file` or `build_context` can be provided.")
+
     if conda_environment_file is None and build_context is None:
         conda_environment_file = find_file_in_parent_to_pythonpath(CONDA_ENVIRONMENT_FILE)
         if conda_environment_file is None:
@@ -943,11 +965,13 @@ def submit_to_azure_if_needed(  # type: ignore
         if strictly_aml_v1:
             assert isinstance(conda_environment_file, Path)
             assert aml_workspace is not None, "An AzureML workspace should have been created already."
+            assert isinstance(build_context, Union[DockerBuildContext, None])
             run_config = create_run_configuration(
                 workspace=aml_workspace,
                 compute_cluster_name=compute_cluster_name,
                 aml_environment_name=aml_environment_name,
                 conda_environment_file=conda_environment_file,
+                docker_build_context=build_context,
                 environment_variables=environment_variables,
                 pip_extra_index_url=pip_extra_index_url,
                 private_pip_wheel_path=_str_to_path(private_pip_wheel_path),
--- a/hi-ml-azure/src/health_azure/himl.py
+++ b/hi-ml-azure/src/health_azure/himl.py
@@ -21,7 +21,7 @@ from typing import Any, Callable, Dict, Generator, List, Optional, Tuple, Union
 
 from azure.ai.ml import Input, MLClient, Output, command
 from azure.ai.ml.constants import InputOutputModes
-from azure.ai.ml.entities import Command, Data
+from azure.ai.ml.entities import BuildContext, Command, Data
 from azure.ai.ml.entities import Environment as EnvironmentV2
 from azure.ai.ml.entities import Job, Sweep, UserIdentityConfiguration
 from azure.ai.ml.entities._job.distribution import DistributionConfiguration, MpiDistribution, PyTorchDistribution
@@ -58,6 +58,7 @@ from health_azure.utils import (
     get_workspace,
     is_run_and_child_runs_completed,
     is_running_in_azure_ml,
+    load_and_hash_directory,
     register_environment,
     register_environment_v2,
     run_duration_string_to_seconds,
@@ -464,13 +465,13 @@ def submit_run_v2(
     root_dir = sanitize_snapshoot_directory(snapshot_root_directory)
     script_params = script_params or []
     script_param_str = create_v2_job_command_line_args_from_params(script_params)
+    entry_script_relative = sanitize_entry_script(entry_script, root_dir)
     if entry_command is None:
-        entry_script_relative = sanitize_entry_script(entry_script, root_dir)
         experiment_name = effective_experiment_name(experiment_name, entry_script_relative)
         cmd = " ".join(["python", str(entry_script_relative), script_param_str])
     else:
         experiment_name = effective_experiment_name(experiment_name, entry_command)
-        cmd = " ".join([str(entry_command), script_param_str])
+        cmd = " ".join([str(entry_command), entry_script_relative, script_param_str])
 
     print(f"The following command will be run in AzureML: {cmd}")
 
@@ -734,6 +735,7 @@ def submit_to_azure_if_needed(  # type: ignore
     snapshot_root_directory: Optional[PathOrString] = None,
     script_params: Optional[List[str]] = None,
     conda_environment_file: Optional[PathOrString] = None,
+    build_context: Optional[BuildContext] = None,
     aml_environment_name: str = "",
     experiment_name: Optional[str] = None,
     environment_variables: Optional[Dict[str, str]] = None,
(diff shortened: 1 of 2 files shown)


hi-ml-azure/src/health_azure/himl.py
```diff
@@ -193,8 +197,24 @@ def create_run_configuration(
     """
     run_config = RunConfiguration()
 
+    # If both docker_build_context and conda_environment_file are provided, throw an error
+    if docker_build_context and conda_environment_file:
```
fepegar, line 201: Do we also need to check for `aml_environment_name`?

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -193,8 +197,24 @@ def create_run_configuration(
...
+        raise ValueError("Both docker_build_context and conda_environment_file cannot be provided at the same time.")
+
     if aml_environment_name:
         run_config.environment = Environment.get(workspace, aml_environment_name)
+    elif docker_build_context:
+        # Check that the dockerfile exists
+        dockerfile_path = Path(docker_build_context.location) / docker_build_context.dockerfile_path
+        if not dockerfile_path.exists():
+            raise ValueError("Dockerfile not found in the provided docker_build_context.")
```
fepegar, line 210:
```suggestion
            raise ValueError(f"Dockerfile not found in the provided docker_build_context '{dockerfile_path }'.")
```

FYI I think Ruff recommends instantiating the error message first to improve readability in error tracebacks. So

```suggestion
            msg = f"Dockerfile not found in the provided docker_build_context '{dockerfile_path }'."
            raise ValueError(msg)
```

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -193,8 +197,24 @@ def create_run_configuration(
...
+        environment_name = load_and_hash_directory(Path(docker_build_context.location))
```
fepegar, line 211: Maybe do `docker_build_context_dir = Path(docker_build_context.location)` and then use that through the method?

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -193,8 +197,24 @@ def create_run_configuration(
...
+        new_environment = Environment.from_docker_build_context(
+            name=environment_name,
+            docker_build_context=docker_build_context,
+        )
+        registered_env = register_environment(workspace, new_environment)
+        run_config.environment = registered_env
```
fepegar, line 217: Very minor: I would maybe replace these three with e.g. `environment = registered_env` and then do `run_config.environment = environment` after the else.

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -346,6 +369,7 @@ def create_script_run(
     snapshot_root_directory: Optional[Path] = None,
     entry_script: Optional[PathOrString] = None,
     entry_command: Optional[PathOrString] = None,
+    python_executable: str = "python",
```
fepegar, line 372: Can you document this?

fepegar: Maybe with examples of why one would like to use something different.

fepegar: I see now you've included precisely those examples in the docs of `submit_run_v2`. Nice! That's what I meant.

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -922,14 +949,21 @@ def submit_to_azure_if_needed(  # type: ignore
         print(f"No snapshot root directory given. Uploading all files in the current directory {Path.cwd()}")
         snapshot_root_directory = Path.cwd()
 
-    if conda_environment_file is None:
+    if conda_environment_file is not None and build_context is not None:
```
fepegar, line 952: As above, wondering if we should also check for `aml_environment_name`.

fepegar: Applies for the code below as well.

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -988,11 +1026,21 @@ def submit_to_azure_if_needed(  # type: ignore
...
-            )
+            if conda_environment_file is not None:
+                assert isinstance(conda_environment_file, Path)
+                environment = create_python_environment_v2(
+                    conda_environment_file=conda_environment_file, docker_base_image=docker_base_image
+                )
+            elif build_context is not None:
+                assert isinstance(build_context, BuildContext)
+                # Load all files in the build context directory into a single string
+                # and run generate_unique_environment_name
```
fepegar, line 1037: `generate_unique_environment_name` would also work :)
(see my comment in `load_and_hash_directory`)

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -638,6 +638,25 @@ def is_conda_file_with_pip_include(conda_file: Path) -> Tuple[bool, Dict]:
     return False, conda_yaml
 
 
+def load_and_hash_directory(folder_path: Path) -> str:
```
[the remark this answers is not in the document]
samb-t: Thanks! Added - https://github.com/microsoft/hi-ml/blob/b1aa5a8c4442daa3d167fff31e5691875cb8e0f9/hi-ml-azure/src/health_azure/utils.py#L646-L653

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -638,6 +638,25 @@ def is_conda_file_with_pip_include(conda_file: Path) -> Tuple[bool, Dict]:
...
+    # Initialize an empty string to hold the concatenated contents
+    concatenated_content = b''
+
+    # Use pathlib to get all files in the folder
```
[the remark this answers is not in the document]
samb-t: Removed

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -638,6 +638,25 @@ def is_conda_file_with_pip_include(conda_file: Path) -> Tuple[bool, Dict]:
...
+    folder = Path(folder_path)
+    for filepath in folder.glob('*'):
```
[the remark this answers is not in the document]
samb-t: +1 for rglob, definitely makes more sense: https://github.com/microsoft/hi-ml/blob/abbcc4ea4b39b7f276469701906ae9551633ba74/hi-ml-azure/src/health_azure/utils.py#L656-L659

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -638,6 +638,25 @@ def is_conda_file_with_pip_include(conda_file: Path) -> Tuple[bool, Dict]:
...
+        # Check if it's a file (not a directory)
+        if filepath.is_file():
+            with filepath.open('rb') as file:
+                concatenated_content += file.read()
+
+    hash_object = hashlib.sha1()
+    hash_object.update(concatenated_content)
+    overall_hash = hash_object.hexdigest()[:32]
```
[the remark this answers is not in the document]
samb-t: I set it up to generated names in the same structure as this function for conda yamls: https://github.com/microsoft/hi-ml/blob/e50a3c1174665a25b5ec6296039b147e882a1cd9/hi-ml-azure/src/health_azure/utils.py#L641-L654

I've updated both functions to take those values as arguments and documented them https://github.com/microsoft/hi-ml/blob/abbcc4ea4b39b7f276469701906ae9551633ba74/hi-ml-azure/src/health_azure/utils.py#L641-L653

https://github.com/microsoft/hi-ml/blob/abbcc4ea4b39b7f276469701906ae9551633ba74/hi-ml-azure/src/health_azure/utils.py#L668-L682

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -638,6 +638,25 @@ def is_conda_file_with_pip_include(conda_file: Path) -> Tuple[bool, Dict]:
...
+    unique_env_name = f"HealthML-{overall_hash}"
+    return unique_env_name
```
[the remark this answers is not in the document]
samb-t: Renamed to https://github.com/microsoft/hi-ml/blob/abbcc4ea4b39b7f276469701906ae9551633ba74/hi-ml-azure/src/health_azure/utils.py#L641

And changed the conda description one to https://github.com/microsoft/hi-ml/blob/abbcc4ea4b39b7f276469701906ae9551633ba74/hi-ml-azure/src/health_azure/utils.py#L668

hi-ml-azure/testazure/testazure/test_himl.py
```diff
@@ -250,7 +252,10 @@ def test_create_run_configuration_fails(
     mock_workspace.compute_targets = {existing_compute_target: 123}
     with pytest.raises(ValueError) as e:
         himl.create_run_configuration(compute_cluster_name="b", workspace=mock_workspace)
-    assert "One of the two arguments 'aml_environment_name' or 'conda_environment_file' must be given." == str(e.value)
+    assert (
+        "One of the three arguments 'aml_environment_name', 'docker_build_context' or "
+        "'conda_environment_file' must be given."
+    ) == str(e.value)
```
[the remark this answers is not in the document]
samb-t: Done, thanks! https://github.com/microsoft/hi-ml/blob/abbcc4ea4b39b7f276469701906ae9551633ba74/hi-ml-azure/testazure/testazure/test_himl.py#L253-L260

hi-ml-azure/src/health_azure/utils.py
```diff
@@ -666,9 +666,7 @@ def generate_unique_environment_name_from_directory(
 
 
 def generate_unique_environment_name_from_description(
-    environment_description_string: str,
-    environment_name_prefix: str = "HealthML-",
-    num_hash_digits: int = 32
+    environment_description_string: str, environment_name_prefix: str = "HealthML-", num_hash_digits: int = 32
```
fepegar, line 669: I think the formatter did this because there's no comma after `32`.

hi-ml-azure/src/health_azure/himl.py
samb-t: Good question, I went for no as the current behaviour was that if `aml_environment_name` was passed it uses that. But I would be absolutely down for throwing an error in any case when multiple which it doesn't make sense to pass at the same time, are passed at the same time.

hi-ml-azure/src/health_azure/himl.py
samb-t: Think I get what you mean, like this? https://github.com/microsoft/hi-ml/blob/abbcc4ea4b39b7f276469701906ae9551633ba74/hi-ml-azure/src/health_azure/himl.py#L204-L241

hi-ml-azure/src/health_azure/himl.py
samb-t: Ah thanks I missed this one. Copied here too https://github.com/microsoft/hi-ml/blob/abbcc4ea4b39b7f276469701906ae9551633ba74/hi-ml-azure/src/health_azure/himl.py#L390-L391

hi-ml-azure/src/health_azure/himl.py
samb-t: As above, went for `generate_unique_environment_name_from_directory`.

hi-ml-azure/src/health_azure/utils.py
samb-t: Thanks! Moved back to three lines https://github.com/microsoft/hi-ml/blob/905736838f2652eafc3ac9699ad14142aae86252/hi-ml-azure/src/health_azure/utils.py#L668-L672

hi-ml-azure/src/health_azure/himl.py
ant0nsc: agree that conflicting arguments should throw errors

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -193,8 +197,26 @@ def create_run_configuration(
...
+        raise ValueError("Both docker_build_context and conda_environment_file cannot be provided at the same time.")
+
     if aml_environment_name:
-        run_config.environment = Environment.get(workspace, aml_environment_name)
+        environment = Environment.get(workspace, aml_environment_name)
+    elif docker_build_context:
+        # Check that the dockerfile exists
+        docker_build_context_dir = Path(docker_build_context.location)
+        dockerfile_path = docker_build_context_dir / docker_build_context.dockerfile_path
+        if not dockerfile_path.exists():
```
ant0nsc, line 210:
```suggestion
        if not dockerfile_path.is_file():
```

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -193,8 +197,26 @@ def create_run_configuration(
...
+            msg = f"Dockerfile not found in the provided docker_build_context '{dockerfile_path }'."
+            raise ValueError(msg)
+        environment_name = generate_unique_environment_name_from_directory(docker_build_context_dir)
+        new_environment = Environment.from_docker_build_context(
+            name=environment_name,
+            docker_build_context=docker_build_context,
+        )
+        registered_env = register_environment(workspace, new_environment)
+        environment = registered_env
```
ant0nsc, line 219: move this whole section into a helper function? The whole function is big already.

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -359,6 +387,8 @@ def create_script_run(
         executed. If entry_command is provided, this argument is ignored.
     :param entry_command: The command that should be run in AzureML. Command arguments will be taken from
         the 'script_params' argument. If provided, this will override the entry_script argument.
+    :param python_executable: The command that should be used to launch the script. For example, "python",
+        "torchrun" or "uv run".
```
ant0nsc, line 391: I'm not sure we really need this extra argument. Looking the change in lines 400-401, it seems to me that you can achieve the desired effect already with `entry_command`?

ant0nsc (approved):
Thanks a lot for making those changed! Support for Docker is great to have.
Left a few questions, mostly around: Do we _really_ need to add another argument `python_exec`?

samb-t:
For SDK v1, if a docker build context is passed, I have now set it to hash the URI of the build context in blob storage. This is potentially problematic if the underlying files change - the environment won't update automatically. Better solutions very much welcome

hi-ml-azure/src/health_azure/himl.py
samb-t: Removed this as for SDK v1, `location` corresponds to the location in blob storage as opposed to the local file path (as in SDK v2)

hi-ml-azure/src/health_azure/himl.py
samb-t: Done! Added a function to get and `EnvironmentV2`: https://github.com/microsoft/hi-ml/blob/9c12e58e37e9e25ff13586f53425e98f8311bcb2/hi-ml-azure/src/health_azure/himl.py#L157-L198

hi-ml-azure/src/health_azure/himl.py
samb-t: We spoke about this in person and decided it's fine to keep it this way to reduce making changes to existing code, as the current behaviour when `entry_command` is passed is to not inject the script name.

hi-ml-azure/src/health_azure/utils.py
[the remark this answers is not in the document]
samb-t: I've changed the function name back to `generate_unique_environment_name`. It now takes `hashable` which if it's a string - hashes the string. If it's a Path it will hash the contents of the file or directory. https://github.com/microsoft/hi-ml/blob/9c12e58e37e9e25ff13586f53425e98f8311bcb2/hi-ml-azure/src/health_azure/utils.py#L641-L675

hi-ml-azure/src/health_azure/himl.py
samb-t: Sounds good. Added the following checks:
https://github.com/microsoft/hi-ml/blob/9c12e58e37e9e25ff13586f53425e98f8311bcb2/hi-ml-azure/src/health_azure/himl.py#L252-L262

https://github.com/microsoft/hi-ml/blob/9c12e58e37e9e25ff13586f53425e98f8311bcb2/hi-ml-azure/src/health_azure/himl.py#L170-L180

hi-ml-azure/src/health_azure/utils.py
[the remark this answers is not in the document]
fepegar: Very nice. Thank you!

FYI, I think a docstring convention is

```python
def f():
    """Do something.
    
    Additional explanations here.
    """
```

or

```python
def f():
    """Do something."""
```

Ruff etc. might complain otherwise, but I'm not sure.

PR #964, opened 2025-04-01
ENH: Add support to pass parameters used for Singularity

fepegar:


fepegar:
image

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -868,6 +878,8 @@ def submit_to_azure_if_needed(  # type: ignore
         the 'script_params' argument. If provided, this will override the entry_script argument.
     :param compute_cluster_name: The name of the AzureML cluster that should run the job. This can be a cluster with
         CPU or GPU machines.
+    :param instance_type: The type of virtual machine to be used by the compute target.
```
ant0nsc, line 881:
```suggestion
    :param instance_type: The type of virtual machine to be used by the compute target. This only affects job submission via SDK v2.
```

hi-ml-azure/src/health_azure/himl.py
```diff
@@ -868,6 +878,8 @@ def submit_to_azure_if_needed(  # type: ignore
...
+    :param resources_properties: Additional properties to set on the job resources.
```
ant0nsc, line 882:
```suggestion
    :param resources_properties: A dictionary with additional properties to set on the job resources. This only affects jobs submission via SDK v2.
```

ant0nsc (approved):
Thank you :-)

ant0nsc:
@fepegar Ubuntu fixes in #965

samb-t (approved):
Approved. If we drop support for SDK v1 at some point, I would be in favour of simplifying this and just directly passing a [`JobResourceConfiguration`](https://learn.microsoft.com/en-us/python/api/azure-ai-ml/azure.ai.ml.entities.jobresourceconfiguration) object rather than flattening everything out. E.g. I can see a world where we want to pass a locations list as well.

fepegar:
Thank you both for reviewing.

